PR #3, opened 2015-08-12
Mkdir before trying to write to the file

illicitonion:


leonerd:
Your commit descriptions suggest adding a mkdir(), which the first patch adds, but the second patch then removes it again. Is this intentional?

illicitonion:
Yes, write_yaml_file pretty much exists to do the mkdir, so I figured using it, rather than repeating its code, was the right thing to do.

leonerd:
Ahright. In that case, LGTM

Kegsay:
Test comment

PR #12, opened 2015-09-09
Use an existing room_id when testing if we can set a room alias

erikjohnston:


tests/50federation/11query-directory.pl
```diff
@@ -44,13 +44,12 @@
 
 test "Inbound federation can query room alias directory",
    requires => [qw( outbound_client do_request_json first_home_server
-                    can_create_room_alias )],
+                    room_id can_create_room_alias)],
```
leonerd: I try to put the "real value things" on one line, and the "can_..." ability markers on a separate line. So room_id probably wants to be on line 46

leonerd:
LGTM

PR #14, opened 2015-09-10
SPEC-216: Allow users to view the history of rooms that they have left.

NegativeMjark:
Check the changes made by matrix-org/matrix-doc#40


tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+            provide departed_room_id => $room_id;
+
+            $change_room_powerlevels->($user_a, $room_id, sub {
+                my ( $levels ) = @_;
+                $levels->{users}{ $user_b->user_id } = 50;
+            })
+        })->then( sub {
+            $do_request_json_for->( $user_b,
+                method => "PUT",
+                uri => "/rooms/$room_id/state/m.room.name",
```
leonerd: Is this up to date with latest 'master'? Specifically, commit a356cf91 would mean that all these URIs now need /api/v1/ prepending to it

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+                content => { "name" => "N1. B's room name before A left", },
+            )->on_done(sub { pass "User B set the room name the first time" })
```
leonerd: I like this use of `->on_done` for `pass` messages. I'll have a go at updating other tests to match it.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+    requires => [qw(do_request_json departed_room_id)],
+    check => sub {
+        my ($do_request_json, $departed_room_id) = @_;
+
+        $do_request_json->(
+            method => "GET",
+            uri => "/initialSync",
+            params => { limit => 2 },
+        )->then( sub {
+            my ( $body ) = @_;
```
leonerd: As there's a lot of `die` calls later on, it might be nice to future developers to put a call to
    log_if_fail "initialSync data", $body;

here so that if it dies, they can see what the `$body` data was.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+    requires => [qw(do_request_json departed_room_id)],
+    check => sub {
+        my ($do_request_json, $departed_room_id) = @_;
+
+        $do_request_json->(
+            method => "GET",
+            uri => "/rooms/$departed_room_id/initialSync",
+            params => { limit => 2 },
+        )->then( sub {
+            my ( $room ) = @_;
```
leonerd: Ditto `log_if_fail`

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+            params => { limit => 2 },
+        )->then( sub {
+            my ( $body ) = @_;
+
+            require_json_keys( $body, qw( rooms ) );
+
+            my ( $room ) = grep { $_->{room_id} eq $departed_room_id }
+                @{$body->{rooms}};
+
+            die "Departed room not in /initialSync"
```
leonerd: A somewhat stylistic comment, but I generally prefer these sorts of test assertions to read

  condition I want to be true or
    die "The condition didn't hold"

for example  https://github.com/matrix-org/sytest/blob/master/tests/50federation/11query-directory.pl#L32

It emphasises the desired state and de-emphasises the error messages if that state does not hold, rather than vice-versa.

leonerd:
I don't know how picky to be on the minor whitespace inconsistencies such as `qw(key names)` instead of `qw( key names )` and so on... Feel free to adjust them if you're there but I won't let that alone block it.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
+multi_test "Setup a room, and havve the first user leave (SPEC-216)",
```
illicitonion: Typo havve

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+
+    requires => [qw(
+        make_test_room change_room_powerlevels
+            do_request_json_for user more_users
+        can_create_room
+    )],
+
+    provides => [qw( departed_room_id )],
+
+    # User A creates a room.
```
illicitonion: Woah. This is a whole lot of comments.

If the code isn't clear enough on its own, could we pull out well-named helper subs, so that the code reads more like these comments, rather than write everything both in non-compilable and compilable form?

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,272 @@
...
+
+            $change_room_powerlevels->($user_a, $room_id, sub {
+                my ( $levels ) = @_;
+                $levels->{users}{ $user_b->user_id } = 50;
+            })
+        })->then( sub {
+            $do_request_json_for->( $user_b,
+                method => "PUT",
+                uri => "/api/v1/rooms/$room_id/state/m.room.name",
+                content => { "name" => "N1. B's room name before A left", },
```
illicitonion: This implies that we have specified that the default power level required to set a room's name is 50.

If this isn't explicitly specified in the spec, this test should probably explicitly set that power level requirement.

tests/30rooms/11leaving.pl
NegativeMjark: Fixed

tests/30rooms/11leaving.pl
NegativeMjark: I've set a comment explaining the choice of 50 for the power level.

NegativeMjark:
This is now in #17

PR #16, opened 2015-09-22
Disable stats reporting on sytest

illicitonion:
Pairs with https://github.com/matrix-org/synapse/pull/284

### DIFF
--- a/tests/05synapse.pl
+++ b/tests/05synapse.pl
@@ -76,6 +76,7 @@ prepare "Starting synapse",
                recaptcha_private_key    => "sytest_recaptcha_private_key",
 
                use_insecure_ssl_client_just_for_testing_do_not_use => 1,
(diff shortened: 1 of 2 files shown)


tests/05synapse.pl
```diff
@@ -76,6 +76,7 @@ sub gen_token
                recaptcha_private_key    => "sytest_recaptcha_private_key",
 
                use_insecure_ssl_client_just_for_testing_do_not_use => 1,
+               report_stats => "False",
```
leonerd: Are you sure this is meant to have a capital 'F' in 'False'?

leonerd:
LGTM, aside from "False" issue

tests/05synapse.pl
illicitonion: The existing ones are all False. Leaving as-is.

PR #17, opened 2015-09-28
SPEC-216: Allow users to view the history of rooms that they have left.

NegativeMjark:
Fix #14 to point at the right branch


tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
+multi_test "Setup a room, and have the first user leave (SPEC-216)",
```
leonerd: Since this doesn't actually check anything and merely prepares things, it probably wants to be `prepare` instead of `multi_test`.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+
+    requires => [qw(
+        make_test_room change_room_powerlevels do_request_json_for user
+        more_users can_create_room
```
leonerd: Put `can` requirements on their own line so it's easier to add new (valued) dependencies later.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+    )],
+
+    provides => [qw( departed_room_id )],
```
leonerd: As this room ID is only used within this file, it saves leaking it into other files (and increasing the interdependency between them) by not putting it in the env., and just storing it as a local variable within the file that the later `test` blocks can read. For prior art, see the use of

  https://github.com/matrix-org/sytest/blob/master/tests/30rooms/08levels.pl#L3

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+    # User B will send a message with text "M3. B's message after A left"
+    #
+    do => sub {
+        my (
+            $make_test_room, $change_room_powerlevels, $do_request_json_for,
+            $user_a, $more_users
+        ) = @_;
+        my $user_b = $more_users->[1];
+
+        my $room_id;
```
leonerd: In fact if you just pull this out to the toplevel in the file, the other `test` blocks will be able to see it.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+
+        $make_test_room->([$user_a, $user_b])->then( sub {
```
leonerd: Spacing (see `STYLE-GUIDE.rst`)

    $make_test-room->( [ $user_a, $user_b ] )->then( sub {

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+                # Set user B's power level so that they can set the room
+                # name. By default the level to set a room name is 50.
+                $levels->{users}{ $user_b->user_id } = 50;
+            })
+        })->then( sub {
+            $do_request_json_for->( $user_b,
+                method => "PUT",
+                uri => "/api/v1/rooms/$room_id/state/m.room.name",
+                content => { "name" => "N1. B's room name before A left", },
+            )->on_done(sub { pass "User B set the room name the first time" })
```
leonerd: Since we're really not expecting that all of these preparation steps will fail (because we're only doing things that earlier test files have asserted work), there's no real benefit in having all the `pass` messages here. They might as well be removed to reduce the verbosity of output

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+                uri => "/api/v1/rooms/$room_id/state/madeup.test.state",
+                content => { "body" => "S2. B's state after A left", },
+            )->on_done(sub { pass "User B set the state the second time" })
+        })
+    };
+
+
+test "A departed room is still included in /initialSync (SPEC-216)",
+    requires => [qw( do_request_json departed_room_id )],
+    check => sub {
```
leonerd: Add blank line above

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+        $do_request_json->(
+            method => "GET",
+            uri => "/api/v1/initialSync",
+            params => { limit => 2 },
+        )->then( sub {
+            my ( $body ) = @_;
+
+            require_json_keys( $body, qw( rooms ) );
+
+            my ( $room ) = grep { $_->{room_id} eq $departed_room_id }
```
leonerd: More idiomatic to use `List::Util::first` here

    my $room = first { COND } LIST;

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+                @{$body->{rooms}};
+
+            die "Departed room not in /initialSync"
```
leonerd: As per STYLE-GUIDE, these should be written as condition-first assertions, so the required state comes upfront

    $room or
       die "Departed room not in /initialSync";

and in fact in this particular case, the trailing `or die` can be appended to the `first` call line.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,299 @@
...
+
+            Future->done(1);
+        })
+    };
+
+test "Getting messages going forward is limited for a departed room (SPEC-216)",
+    requires => [qw( do_request_json departed_room_id )],
+    check => sub {
+        my ($do_request_json, $departed_room_id) = @_;
+
```
leonerd: Extraneous blank line

tests/30rooms/11leaving.pl
NegativeMjark: I'd like to keep the progress logging that multi_test provides and that currently doesn't seem possible with prepare.

NegativeMjark:
I think I've responded to the feedback from @leonerd up to this point.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,292 @@
...
+my $room_id;
+
+multi_test "Setup a room, and have the first user leave (SPEC-216)",
+    requires => [qw(
+        make_test_room change_room_powerlevels do_request_json_for user
+        more_users
+        can_create_room
+    )],
+
+    # User A creates a room.
```
illicitonion: My old comment about the comments from #14 still stands

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,292 @@
...
+        my (
+            $make_test_room, $change_room_powerlevels, $do_request_json_for,
+            $user_a, $more_users
+        ) = @_;
+        my $user_b = $more_users->[1];
+
+        $make_test_room->( [$user_a, $user_b] )->then( sub {
+            ( $room_id ) = @_;
+
+            $change_room_powerlevels->($user_a, $room_id, sub {
```
illicitonion: Nit: Space after (

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,292 @@
...
+                my ( $levels ) = @_;
+                # Set user B's power level so that they can set the room
+                # name. By default the level to set a room name is 50.
+                $levels->{users}{ $user_b->user_id } = 50;
```
illicitonion: I would still rather explicitly set the room's required power level.

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,292 @@
...
+                method => "PUT",
+                uri => "/api/v1/rooms/$room_id/state/madeup.test.state",
+                content => { "body" => "S2. B's state after A left", },
+            )->on_done( sub { pass "User B set the state the second time" } )
+        })
+    };
+
+
+test "A departed room is still included in /initialSync (SPEC-216)",
+    requires => [qw( do_request_json )],
```
illicitonion: There seems to be no dependency on the above test here? Maybe you could require the room_id?

(And for below tests)

tests/30rooms/11leaving.pl
```diff
@@ -0,0 +1,292 @@
...
+    check => sub {
+        my ( $do_request_json ) = @_;
+
+        $do_request_json->(
+            method => "GET",
+            uri => "/api/v1/rooms/$room_id/members",
+        )->then( sub {
+            my ( $body ) = @_;
+
+            require_json_keys( $body, qw( chunk ) );
```
illicitonion: No assertions on its contents?

(Specifically, that the left user is still in the member list)

tests/30rooms/11leaving.pl
NegativeMjark: @leonerd didn't want to pass the room id as a dependency since it wasn't used outside this file. https://github.com/matrix-org/sytest/pull/17#discussion-diff-40563341
I suggest arguing with him.

tests/30rooms/11leaving.pl
NegativeMjark: I've added some checks e990204

tests/30rooms/11leaving.pl
NegativeMjark: Done e93a48e.

NegativeMjark:
I think I've responded to the comments from @illicitonion

NegativeMjark:
After discussing with @illicitonion and @leonerd we concluded that both options for passing the room_id were wrong and that I can pick whichever one I want to. Hopefully we will find a better option but I can probably commit it as is for now since it matches existing tests that will need fixing anyway when we do.

PR #21, opened 2015-10-08
Add a flag to initial sync for if we want rooms that the user has left

NegativeMjark:
Documented in matrix-org/matrix-doc#88


leonerd:
LGTM

PR #22, opened 2015-10-13
Don't output control characters if not to TTY

illicitonion:
Side effects:
 * Stop mixing random output to STDERR
 * Stop outputting "Waiting..." for tests which are longer than 2
   seconds

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -321,12 +321,6 @@ sub _run_test
       Future->wait_any(
          $f_test,
 
-         $loop->delay_future( after => 2 )
-            ->then( sub {
-               $output->start_waiting;
-               $loop->new_future->on_cancel( sub { $output->stop_waiting });
-            }),
-
          $loop->delay_future( after => $params{timeout} // 10 )
             ->then_fail( "Timed out waiting for test" )
       )->get;
--- a/tests/01http-server.pl
+++ b/tests/01http-server.pl
@@ -127,12 +127,12 @@ package SyTest::HTTPServer {
       my $path = uri_unescape $request->path;
 
       if( $CLIENT_LOG ) {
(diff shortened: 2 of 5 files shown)


run-tests.pl
```diff
@@ -321,12 +321,6 @@ sub _run_test
       Future->wait_any(
          $f_test,
 
-         $loop->delay_future( after => 2 )
```
leonerd: Any particular reason to get rid of the "slow test" notification, rather than just make the outputting of it conditional on being a TTY, like the colouring?

illicitonion: I didn't see enough value to be worth keeping.

What's it guarding against? What's it meant to be telling me? If a test has taken two seconds, I *know* I'm waiting on it.

leonerd: Hmm.. I guess. I think I added that before adding the overall 10 second deadline, so it's probably less interesting now.

Fine, let it die.

tests/01http-server.pl
```diff
@@ -127,7 +127,9 @@ package SyTest::HTTPServer {
       my $path = uri_unescape $request->path;
 
       if( $CLIENT_LOG ) {
-         print STDERR "\e[1;32mReceived Request\e[m for $method $path:\n";
+         my $green = -t STDERR ? "\e[1;32m" : "";
```
leonerd: Do you mean `STDERR` here? Surely `STDOUT`

leonerd:
LGTM, aside from STDOUT/STDERR conflation in `tests/01http-server.pl`

tests/01http-server.pl
illicitonion: One of these days, I will stop assuming that the code you write had any intention behind it at all.

Done.

PR #25, opened 2015-10-19
Add tests for rejecting invites

illicitonion:


tests/30rooms/06invite.pl
```diff
@@ -106,3 +113,36 @@
          Future->done(1);
       });
    };
+
+my $other_local_user_preparer = local_user_preparer();
+
+test "Invited user can reject invite",
+   requires => [ do { my $creator = local_user_preparer(); $creator, local_user_preparer(), inviteonly_room_preparer( creator => $creator ) } ],
```
leonerd: Some more linefeeds would help here. Maybe

    requires => [ local_user_preparer(),
       do {
          my $creator = local_user_preparer();
          $creator, inviteonly_room_preparer( creator => $creator );
       } ],

by pulling the (unrelated) invitee up front. You'll have to swap the order of `$creator` and `$invitee` to the invoked function below then.

tests/30rooms/06invite.pl
```diff
@@ -2,39 +2,46 @@
...
+      do => sub {
+         my ( $creator ) = @_;
+
+         matrix_create_room( $creator,
+            # visibility: "private" actually means join_rule: "invite"
+            # See SPEC-74
+            visibility => "private",
          )->then( sub {
-            my ( $body ) = @_;
+            my ( $room_id ) = @_;
```
leonerd: I can't quite recall why all this checking logic is sat here. It probably doesn't want to live in the preparer itself; it should be done just once, in a test somewhere else. But leave it here for now since it was there before; I'll move it out sometime.

leonerd:
LGTM, asides the linebreak formatting.

Given the test failure, I suspect it needs a merge on synapse before this itself, yes?

tests/30rooms/06invite.pl
illicitonion: Done

illicitonion:
Yep, blocked on https://github.com/matrix-org/synapse/pull/317

illicitonion:
Manually merged

PR #26, opened 2015-10-19
Add tests for getting archived rooms in v2 sync

NegativeMjark:
Tests changes made in matrix-org/synapse#316


tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+     matrix_register_user_with_filter( $http, {} )->then( sub {
+         ( $user, $filter_id ) = @_;
+
+         matrix_create_room( $user );
+      })->then( sub {
+         ( $room_id ) = @_;
+
+         matrix_sync( $user, filter => $filter_id );
+      })->then( sub {
+         my ($body ) = @_;
```
leonerd: my ( $body )

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+
+         $next = $body->{next_batch};
```
leonerd: You can avoid the always-messy antipattern of capturing outer lexicals for later mutation here, by doing

    matrix_leave_room( $user, $room_id )
       ->then_done( $body->{next_batch} );

Causing the subsequent `then` to be able to receive

    my ( $next ) = @_;
    matrix_sync( $user, filter => $filter_id, since => $next );

That might seem a little neater, though it's swings and roundabouts either way.

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+         ( $user, $filter_id ) = @_;
+
+         Future->needs_all(
+            matrix_create_room( $user )->on_done( sub { ( $room_id_1 ) = @_; } ),
+            matrix_create_room( $user )->on_done( sub { ( $room_id_2 ) = @_; } ),
+         );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id );
+      })->then( sub {
+         my ($body ) = @_;
```
leonerd: my ( $body )

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+
+         $next = $body->{next_batch};
```
leonerd: Similar comments about captured outer lexicals and `then_done` chaining

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+
+         matrix_leave_room( $user, $room_id_1 );
+      })->then( sub {
+         Future->needs_all( map {
+            matrix_send_room_message( $user, $room_id_2,
```
leonerd: It's unclear from the subsequent stages what these filler messages are doing. Also, you might want to pick a different word than "filler", due to its visual similarity and ease of accidental misreading/mistyping with "filter" :)

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+         matrix_register_user_with_filter( $http, $filter ),
+         matrix_register_user_with_filter( $http, $filter ),
+      )->then( sub {
+         ( $user_a, $filter_id_a, $user_b, $filter_id_b ) = @_;
+
+         matrix_create_room( $user_a );
+      })->then( sub {
+         ( $room_id ) = @_;
+
+         matrix_invite_user_to_room( $user_a, $user_b, $room_id );
```
leonerd: Is this invite necessary? By default `matrix_create_room` creates public rooms that any may join

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+      })->then( sub {
+         matrix_join_room( $user_b, $room_id );
```
leonerd: For that matter; is this separate create-then-join required, as compared to `matrix_create_and_join_room` ?

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+      })->then( sub {
+         matrix_put_room_state( $user_a, $room_id,
+            type      => "a.madeup.test.state",
+            content   => { "my_key" => "after" },
+            state_key => "",
+         );
+      })->then( sub {
+         matrix_sync( $user_b, filter => $filter_id_b );
+      })->then( sub {
+         my ( $body ) = @_;
```
leonerd: Since there are a number of `die`s in this block, it might be useful to aid debugging on failure, to put

    log_if_fail "User B sync", $body;

tests/31sync/09archived.pl
```diff
@@ -0,0 +1,193 @@
...
+
+         my $room = $body->{rooms}{archived}{$room_id};
+         require_json_keys( $room, qw( event_map timeline state ));
+         @{ $room->{state}{events} } == 1
+            or die "Expected a single state event";
+         @{ $room->{timeline}{events} } == 1
+            or die "Expected a single timeline event";
+         $room->{event_map}{$room->{state}{events}[0]}{content}{my_key}
```
leonerd: Complex inner key expressions want extra whitespace to make them stand out:

    $room->{event_map}{ $room->{state}{events}[0] }{content}{my_key}

Or consider pulling that out

    my $state_event_id = $room->{state}{events}[0];

tests/31sync/09archived.pl
NegativeMjark: It's necessary if I don't want to accidentally setup a v1 events stream for the user.

tests/31sync/09archived.pl
NegativeMjark: Eh. I wasn't sure if that was a documented, intentional behaviour of matrix_create_room.

tests/31sync/09archived.pl
NegativeMjark: Fixed 7a8849f

tests/31sync/09archived.pl
NegativeMjark: Hmm, maybe. I don't know if it is that valuable given how easy it is to turn on client request/response logging in sytest

tests/31sync/09archived.pl
NegativeMjark: Done 7a8849f

tests/31sync/09archived.pl
NegativeMjark: I'm not sure how well that works when there are many steps between the first ``matrix_sync`` and the second. I think declaring a list of outer variables at the start is fine. I don't think I mutate the variable anywhere other than from undef -> final value, so it shouldn't be too much of a problem.

tests/31sync/09archived.pl
NegativeMjark: Fixed 7a8849f

tests/31sync/09archived.pl
leonerd: That's the intended default, yes. If it's not documented yet we could add it. But feel free to rely on it anyway.

tests/31sync/09archived.pl
leonerd: Ah in that case, as with a number of other tests, we could just add a flag for creating users without the v1 events.
Though I'm given to ask: Why does setting up a v1 event stream make any difference to v2? Does the server care about that?

tests/31sync/09archived.pl
leonerd: It's a lot less noisy to print just verbose information out of a single failing test when it fails, than see all of the server/client traffic

tests/31sync/09archived.pl
NegativeMjark: Yes, it affects the presence state of the user.

tests/31sync/09archived.pl
NegativeMjark: I'd rather avoid stuffing too much logging on failure. I found it very difficult to debug some of the tests due to overenthusiastic use of log_if_fail, since it made it more annoying to read the client logs. I'd suggest leaving it as is, and if someone finds it too difficult to debug without that logging then they can always add it themselves.

tests/31sync/09archived.pl
NegativeMjark: Done dff6d8e

leonerd:
LGTM

PR #28, opened 2015-10-20
Add a matrix_typing sub for marking a user as typing

NegativeMjark:
Since I'm just about to start checking for typing notifs in v2 sync.


leonerd:
LGTM

PR #30, opened 2015-10-23
Filter the v2 sync result to include only m.presence events

NegativeMjark:
So as to test the v2 presence filtering.


leonerd:
Does it perhaps also want a test (somewhere? Maybe not this file) that tests what it used to test before adding the filter? Namely, that if you don't ask that you want presence, you still get it among the other things anyway

leonerd:
LGTM

PR #31, opened 2015-10-26
Add some tests for event size limits

illicitonion:


tests/80torture/03events.pl
```diff
@@ -54,3 +54,32 @@
          params => { from => $user->eventstream_token, timeout => "hello" },
       )->main::expect_http_4xx;
    };
+
+test "Event size limits",
+   requires => do {
+      my $user = local_user_preparer();
+      [ $user, room_preparer( requires_users => [ $user ] ) ];
```
leonerd: FYI, there's `local_user_and_room_preparers` in `10apidoc/33room-members.pl` which does exactly this. You could use that instead.

leonerd:
LGTM, though could use existing function.

tests/80torture/03events.pl
illicitonion: Done

illicitonion:
Squash-merged as 526328b

PR #32, opened 2015-10-26
Integration tests for full_state incremental sync

richvdh:
Implement tests for the full_state incremental sync, as specced at
https://github.com/matrix-org/matrix-doc/pull/133 and implemented at
https://github.com/matrix-org/synapse/pull/332

### DIFF
--- a/tests/31sync/03joined.pl
+++ b/tests/31sync/03joined.pl
@@ -34,9 +34,34 @@ test "Can sync a joined room",
          my $room = $body->{rooms}{joined}{$room_id};
          (!defined $room) or die "Unchanged rooms shouldn't be in the sync response";
 
-         # do another sync, with full_state
+         Future->done(1)
+      })
+   };
+
+
+test "Full state sync includes joined rooms",
+   requires => [qw( first_api_client can_sync )],
+
+   check => sub {
+      my ( $http ) = @_;
+
+      my ( $user, $filter_id, $room_id );
+
+      my $filter = { room => { timeline => { limit => 10 } } };
+
+      matrix_register_user_with_filter( $http, $filter )->then( sub {
+         ( $user, $filter ) = @_;
+
+         matrix_create_room( $user )
+      })->then( sub {
+         ( $room_id ) = @_;
+
+         matrix_sync( $user, filter => $filter_id )
+      })->then( sub {
+         my ( $body ) = @_;
+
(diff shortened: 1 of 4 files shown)


tests/31sync/04timeline.pl
```diff
@@ -247,6 +247,60 @@
...
+         @{ $room->{timeline}{events} } == 1
+             or die "Expected only one timeline event";
+         my $event_id = $room->{timeline}{events}[0];
+         $room->{event_map}{$event_id}{type} eq "another.filler.type"
+            or die "Unexpected timeline event type";
+
+         Future->done(1);
+      })
+   };
+
```
leonerd: I'm sure we don't need three blank lines here :)

tests/31sync/06state.pl
```diff
@@ -191,6 +191,89 @@
...
+         $event->{content}{my_key} == 2
+            or die "Unexpected event content";
+
+         @{ $room->{timeline}{events} } == 1
+             or die "Expected only one timeline event";
+
+         Future->done(1);
+      })
+   };
+
```
leonerd: And we definitely don't need four blank lines here.

tests/31sync/03joined.pl
```diff
@@ -34,6 +34,21 @@
          my $room = $body->{rooms}{joined}{$room_id};
          (!defined $room) or die "Unchanged rooms shouldn't be in the sync response";
 
+         # do another sync, with full_state
```
leonerd: Is this test intended to check interaction with the previous sync above? If not, then perhaps it would be best in its own test, for independence.

tests/31sync/06state.pl
richvdh: five, actually.

I'm not sure I share your fear of whitespace, but which ones do you think need removing?

tests/31sync/03joined.pl
richvdh: yup. done.

tests/31sync/06state.pl
leonerd: I mean the four blank lines between this test and the next; lines 273 to 276. Usually one is the style, though some people are using two blank lines to separate toplevel functions - I don't mind that. Four just seems a bit excessive.

richvdh: Oh! yes it does. two seems like a good amount - it distinguishes the gap between tests from blocks within tests.

leonerd:
LGTM

PR #37, opened 2015-10-29
Check that the timeline ``limited`` flag is set correctly when syncing a newly joined room.

NegativeMjark:


tests/31sync/03joined.pl
```diff
@@ -124,3 +124,65 @@
...
+      my ( $user_a, $filter_id_a, $user_b, $filter_id_b, $room_id, $next_b );
+
+      my $filter = {
+         room => {
+            timeline => { types => [ "m.room.message" ], limit => 10 },
+            state    => { types => [] },
+         }
+      };
+
+      Future->needs_all(
```
illicitonion: How much would you hate me if I asked you to phrase this as:

Future->need_all(
  matrix_register_user_with_filter( $http, $filter )->then( sub {
    ( $user_a, $filter_id_a ) = @_;
  }),
  matrix_register_user_with_filter( $http, $filter )->then( sub {
    ( $user_b, $filter_id_b ) = @_;
  }),
)->then(...

I don't really like that the output of the two calls are magically concatenated into a list for assorted reasons

tests/31sync/03joined.pl
```diff
@@ -124,3 +124,65 @@
...
+         matrix_join_room( $user_b, $room_id );
+      })->then( sub {
+         matrix_sync( $user_b, filter => $filter_id_b, since => $next_b );
+      })->then( sub {
+         my ( $body ) = @_;
+         my $room = $body->{rooms}{joined}{$room_id};
+         my $timeline = $room->{timeline};
+
+         if( @{ $timeline->{events} } == 6 ) {
+            # We could assert that the timeline wasn't limited in this case
```
illicitonion: Let's in fact assert that $timeline->{limited} == JSON::false, and maybe also assert that all six events are m.room.message events (i.e. verify that we're actually filtering out the state event from the join)

NegativeMjark: Done 02ca0ea

tests/31sync/03joined.pl
NegativeMjark: Done fad07fc

tests/31sync/03joined.pl
```diff
@@ -124,3 +124,67 @@
...
+         } 0 .. 3 );
+      })->then( sub {
+         matrix_join_room( $user_b, $room_id );
+      })->then( sub {
+         matrix_sync( $user_b, filter => $filter_id_b, since => $next_b );
+      })->then( sub {
+         my ( $body ) = @_;
+         my $room = $body->{rooms}{joined}{$room_id};
+         my $timeline = $room->{timeline};
+
```
illicitonion: log_if_fail "Timeline", $timeline;

illicitonion:
LGTM with one comment

tests/31sync/03joined.pl
```diff
@@ -124,3 +124,67 @@
...
+            $room->{event_map}{$_}{type} eq "m.room.message"
+               or die "Only expected 'm.room.message' events";
+         } @{ $timeline->{events} };
+
+         if( @{ $timeline->{events} } == 6 ) {
+            $timeline->{limited} == JSON::false
+               or die "Timeline doesn't have all the events so should be limited";
+         }
+         else {
+            $timeline->{limited} == JSON::true
```
leonerd: Is there a particular reason to actually compare this to JSON::true (and false above); or just testing for truthiness? If not, then I'd suggest just use the truth of the value, rather than this full comparison.

NegativeMjark: Yes. I want the JSON to contain ``false`` or ``true``. I'd say that having ``1`` or ``"yep"`` in the JSON is a bug.

tests/31sync/03joined.pl
NegativeMjark: Done 75149b4

PR #38, opened 2015-11-02
Test restriction of unfederated rooms

leonerd:
Test that remote users can't join a room with `m.federate` set to `false`


leonerd:
Assigning to @erikjohnston as the author of the `m.federate` spec and synapse implementation.

erikjohnston:
LGTM

PR #39, opened 2015-11-02
Check that read receipts appear in v2 sync

NegativeMjark:
Tests changes made in matrix-org/synapse#341

### DIFF
--- a/tests/31sync/12receipts.pl
+++ b/tests/31sync/12receipts.pl
@@ -1,0 +1,133 @@
+push our @EXPORT, qw( matrix_advance_receipt );
+
+=head2 matrix_advance_receipt
+
+   matrix_advance_receipt( $user, $room_id, $receipt_type, $event_id )->get;
+
+Update the postion of the up-to-here receipt marker for a room.
+
+=cut
+
+#TODO This should go with the acuatal receipt tests when they exists.
+sub matrix_advance_receipt
+{
+   my ( $user, $room_id, $receipt_type, $event_id ) = @_;
+
+   do_request_json_for( $user,
+      method  => "POST",
+      uri     => "/v2_alpha/rooms/$room_id/receipt/$receipt_type/$event_id",
+      content => {}
+   );
+}
+
+
+test "Read receipts appear in initial v2 /sync",
+   requires => [qw( first_api_client can_sync )],
+
+   check => sub {
+      my ( $http ) = @_;
+
+      my ( $user, $filter_id, $room_id, $event_id );
+
+      my $filter = {
+         room => {
+            state => { types => [] },
+            timeline => { types => [] },
+            ephemeral => { types => ["m.receipt"] },
+         },
+         presence => { types => [] },
+      };
+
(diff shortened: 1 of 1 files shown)


tests/31sync/12receipts.pl
```diff
@@ -0,0 +1,133 @@
...
+
+=head2 matrix_advance_receipt
+
+   matrix_advance_receipt( $user, $room_id, $receipt_type, $event_id )->get;
+
+Update the postion of the up-to-here receipt marker for a room.
+
+=cut
+
+#TODO This should go with the acuatal receipt tests when they exists.
```
leonerd: You might want to see PR #44 and use that when it lands.

tests/31sync/12receipts.pl
```diff
@@ -0,0 +1,133 @@
...
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         my $room = $body->{rooms}{joined}{$room_id};
+
+         my $ephemeral = $room->{ephemeral}{events};
+
+         @{ $ephemeral } == 1 or die "Expected a m.receipt event";
```
leonerd: This test seems a little fragile; expecting there to be exactly one receipt in the list, being the one the test wanted. You may want to see PR #44 again; specifically the `find_receipt` function added in `30rooms/21receipts.pl` which looks for a matching receipt. That would be more robust against additional events being sent by a server.

NegativeMjark: Hmm. It's hard to see why a server would send another m.receipt given that I'm filtering down to only m.receipt events and this test is the only thing interacting with the room.

leonerd: (after IRL chat): Yes that's fair enough, and actually it's nice to have a test that the spec is obeyed here, because there should indeed only be one event.

tests/31sync/12receipts.pl
```diff
@@ -0,0 +1,110 @@
...
+   check => sub {
+      my ( $http ) = @_;
+
+      my ( $user, $filter_id, $room_id, $event_id );
+
+      my $filter = {
+         room => {
+            state => { types => [] },
+            timeline => { types => [] },
+            ephemeral => { types => ["m.receipt"] },
```
leonerd: Align these three lines' `=>` arrows.
Put some whitespace inside the `[ ]` brackets - `[ "m.receipt" ]`.

leonerd:
LGTM aside from a couple of whitespace fixes. (in two places)

tests/31sync/12receipts.pl
```diff
@@ -0,0 +1,110 @@
...
+   check => sub {
+      my ( $http ) = @_;
+
+      my ( $user, $filter_id, $room_id, $event_id, $next_batch );
+
+      my $filter = {
+         room => {
+            state => { types => [] },
+            timeline => { types => [] },
+            ephemeral => { types => ["m.receipt"] },
```
leonerd: Ditto w/s here

leonerd:
LGTM

PR #45, opened 2015-11-05
Check that creating invalid filters returns 4xx

NegativeMjark:


tests/80torture/10filters.pl
```diff
@@ -0,0 +1,34 @@
...
+];
+
+
+test "Check creating invalid filters returns 4xx",
+   requires => [qw( first_api_client )],
+
+   check => sub {
+      my ( $http ) = @_;
+
+      matrix_register_user( $http, undef, with_events => 0 )->then( sub {
```
leonerd: You can avoid some boilerplate and an entire nesting level here by doing

    requires => [ local_user_preparer() ],

    check => sub {
       my ( $user ) = @_;
       ...

NegativeMjark: Done b97c0e0

tests/80torture/10filters.pl
```diff
@@ -0,0 +1,35 @@
...
+
+      Future->wait_all( map {
+         my $filter = $_;
+         matrix_create_filter( $user, $_ )
+            ->main::expect_http_4xx
+            ->on_fail( sub { log_if_fail "Filter:", $filter; });
+      } @{ $INVALID_FILTERS } )->then( sub {
+         # Wait for all the requests to finish, then check that all of them
+         # succeeded.
+         Future->needs_all( @_ );
```
leonerd: I'm not following the reason for this `wait_all` to generate a list of futures, only to then `needs_all` on those futures. Why not just `needs_all` in the first place?

NegativeMjark: Cause otherwise the futures get cancelled if one of the filters fail, triggering the on_fail to log filters that haven't actually failed.

leonerd: Ahyes; that's actually quite cute. It allows each of the attempts to be fully independent and run until completion or success, and only when each is ready we make the final pass/fail decision for the test as a whole.

I might have to document that in Future's docs actually...

leonerd:
LGTM

PR #48, opened 2015-11-05
Make test deterministic

illicitonion:
Right now it hopes that the ordering of two event IDs happens to be
consistent.


NegativeMjark:
LGTM

PR #49, opened 2015-11-06
Explicitly deref the hashes before calling keys on them

NegativeMjark:
Otherwise I get an error on perl 5.22.0 of:

"keys on reference is experimental at tests/30rooms/13anonymousaccess.pl line 216."


illicitonion:
LGTM

PR #50, opened 2015-11-09
Add a test for SYN-513.

ara4n:


tests/42tags.pl
```diff
@@ -378,3 +382,49 @@ sub check_private_user_data {
          Future->done( 1 );
       });
    };
+
+test "Deleted tags appear in an incremental v2 /sync",
```
leonerd: If you add a `bug` argument here, then sytest will be OK about the fact the test fails.

    bug => "SYN-513",

You can then run tests presuming this bug to be fixed, by passing

    ./run-tests.pl ... --fix SYN-513

and that tag will be ignored.

tests/42tags.pl
```diff
@@ -378,3 +385,49 @@ sub check_private_user_data {
...
+         matrix_remove_tag( $user, $room_id, "test_tag" );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id, since => $next_batch );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         my $room = $body->{rooms}{joined}{$room_id};
+         require_json_keys( $room, qw( private_user_data ) );
+
+         check_private_user_data( $room->{private_user_data}{events}, 1 );
```
illicitonion: Reading this, it's really unclear what the 1 means.

Can you make it more clear?

leonerd: I'd consider doing

    ..., expect_empty => 1 )

at the callsite, and unpacking that as a named arg in the function.

tests/42tags.pl
leonerd: Actually this appears not to be necessary; the test currently passes against synapse from `develop` (v0.10.1-rc1-208-g466b4ec)

leonerd:
@illicitonion PTAL

NegativeMjark:
LGTM

illicitonion:
LGTM

PR #51, opened 2015-11-09
Move 'bug' parameter handling up to test() from _run_test()

richvdh:
This has two advantages
* The offending test is correctly marked as 'EXPECTED FAIL' instead of
  appearing as a FAIL
* When running with -s, execution does not abort on an expected failure.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -372,10 +372,6 @@ sub _run_test
 {
    my ( $t, %params ) = @_;
 
-   # We expect this test to fail if it's declared to be dependent on a bug that
-   # is not yet fixed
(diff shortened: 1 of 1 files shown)


run-tests.pl
```diff
@@ -492,6 +488,10 @@ sub test
 {
    my ( $name, %params ) = @_;
 
+   # We expect this test to fail if it's declared to be dependent on a bug that
```
leonerd: Since `multi_test` is a separate thing, don't forget to copy this also to there.

richvdh: gpwm

done

richvdh:
@leonerd: happy now?

leonerd:
LGTM

PR #52, opened 2015-11-09
Tests for guest users joining rooms and sending messages

illicitonion:
Tests for https://github.com/matrix-org/synapse/pull/355

### DIFF
--- a/tests/30rooms/13anonymousaccess.pl
+++ b/tests/30rooms/13anonymousaccess.pl
@@ -1,197 +1,174 @@
+use Future::Utils qw( try_repeat_until_success );
+
 test "Anonymous user cannot view non-world-readable rooms",
-   requires => [ qw( first_api_client ), local_user_preparer() ],
+   requires => [ anonymous_user_preparer(), local_user_preparer() ],
 
    do => sub {
-      my ( $api_client, $user ) = @_;
+      my ( $anonymous_user, $user ) = @_;
 
-      my $anonymous_user;
       my $room_id;
 
-      register_anonymous_user( $api_client )->then( sub {
-         ( $anonymous_user ) = @_;
+      matrix_create_and_join_room( [ $user ] )
+      ->then( sub {
+         ( $room_id ) = @_;
 
-         matrix_create_and_join_room( [ $user ] )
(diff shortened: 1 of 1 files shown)


tests/30rooms/13anonymousaccess.pl
```diff
@@ -297,3 +479,19 @@ sub matrix_set_room_history_visibility
...
 }
+
+sub matrix_get_room_membership
+{
+   my ( $checking_user, $room_id, $tested_user ) = @_;
+
+   matrix_get_room_state( $checking_user, $room_id,
+      type => "m.room.member",
+      state_key => $tested_user->user_id,
+   )->then( sub {
```
leonerd: In the two-argument version of `->then` I usually notate this as

    ->then(
       sub { ... },
       sub { ... },
    )

With each `sub` sitting on its own line, so it looks Obviously Different to the normal single-arg `->then` chaining; and thus calls attention to it on skimming. E.g. compare to

  https://github.com/matrix-org/sytest/blob/develop/tests/00expect_http_fail.pl#L10

tests/30rooms/13anonymousaccess.pl
```diff
@@ -234,6 +236,88 @@
...
+   do => sub {
+      my ( $user, $room_id ) = @_;
+
+      matrix_set_room_guest_access( $user, $room_id, "can_join" );
+   },
+
+   check => sub {
+      my ( $user, $room_id, $api_client ) = @_;
+
+      register_anonymous_user( $api_client )->then( sub {
```
leonerd: Beware that over the course of running this test, you'll register two anonymous users; one for the pre-check before the `do` block and another one after it.

You might prefer to make a `anonymous_user_preparer()` to share around, so that the test can start off already with an anonymous user. That'll reduce the amount of async indent in the tests, and also mean you'll keep the same user for both `check` blocks.

leonerd:
LGTM.

Might be good to merge that into develop before my PR #60 because of the new preparer having to be renamed to fixture - make that my problem ;)

NegativeMjark:
LGTM. Shame that switching to the preparer happened in the same PR

PR #53, opened 2015-11-09
Supply display_name from identity server not tests

illicitonion:


Kegsay:
I don't know enough about sytest to know whether what you've done is good. Checks are failing so I'm guessing it's not good? Either way, removing assignment on this one, sorry!

illicitonion:
Checks are failing because the synapse change hasn't landed :)

leonerd:
LGTM I guess. But wait for synapse to catch up before merging :)

PR #59, opened 2015-11-10
Neater solution to shutting down fake ID server; which doesn't upset …

leonerd:
…IO::Async::Loop::Epoll (SYT-36)


illicitonion:
This isn't particularly intuitive, but sure, rubberstamp, LGTM, whatever

leonerd:
I considered giving it a oneline `->stop_listening` method wrapper, but it didn't overly seem worth it. It's almost a documentation-level thing at that point.

PR #60, opened 2015-11-10
Rename "preparer" to "fixture"; give them teardown ability; document them

leonerd:
This fairly large set of changes is four conceptually-small pieces:

  * Rename `preparer` to `fixture`, and rename the `do` block to `setup` - this change represents the vast majority of the line count in this PR.

  * Give fixtures the ability to have a `teardown` block

  * Write some documentation about the general use of fixtures in the main `DEVELOP.rst`

  * Use the `teardown` fixture ability in 12thirdpartyinvites.pl to shut down the fake ID server at the end of each test

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -298,8 +298,8 @@ sub preparer
 {
    my %args = @_;
 
-   my $do = $args{do} or croak "preparer needs a 'do' block";
-   ref( $do ) eq "CODE" or croak "Expected preparer 'do' block to be CODE";
+   my $setup = $args{setup} or croak "preparer needs a 'setup' block";
+   ref( $setup ) eq "CODE" or croak "Expected preparer 'setup' block to be CODE";
 
    my @req_futures;
    my $f_start = Future->new;
@@ -332,7 +332,7 @@ sub preparer
       sub { $f_start->done( @_ ) unless $f_start->is_ready },
 
       Future->needs_all( @req_futures )
-         ->then( $do )
+         ->then( $setup )
    );
 }
 
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -292,21 +292,21 @@ sub log_if_fail
    push @log_if_fail_lines, split m/\n/, pp( $structure ) if @_ > 1;
 }
 
-struct Preparer => [qw( requires start result )], predicate => "is_Preparer";
+struct Fixture => [qw( requires start result )], predicate => "is_Fixture";
 
-sub preparer
+sub fixture
 {
    my %args = @_;
 
-   my $setup = $args{setup} or croak "preparer needs a 'setup' block";
-   ref( $setup ) eq "CODE" or croak "Expected preparer 'setup' block to be CODE";
+   my $setup = $args{setup} or croak "fixture needs a 'setup' block";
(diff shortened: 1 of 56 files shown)


leonerd:
As it performs a disturbing renaming across the codebase, I propose to merge and fix up any pending branches other authors may have outstanding after this is merged.

run-tests.pl
```diff
@@ -322,19 +325,35 @@ sub preparer
...
 
       Future->needs_all( @req_futures )
-         ->then( $do )
+         ->then( $setup ),
+
+      $teardown ? sub {
+         my ( $self ) = @_;
+         my $result_f = $self->result;
+         $self->result = Future->fail(
+            "This Fixture has been torn down and cannot be used again"
```
illicitonion: Does this actually fail the test? Or is the failure ignored?

run-tests.pl
```diff
@@ -470,6 +491,15 @@ sub _run_test
       1;
    };
 
+   Future->needs_all( map {
+      if( is_Fixture( $_ ) and $_->teardown ) {
```
illicitonion: Perl trivia: Why does this not call teardown?

tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -181,29 +175,25 @@ sub can_invite_unbound_3pid
    };
 
 sub invite_should_fail {
-   my ( $inviter, $invitee, $hs_base_url, $bind_sub ) = @_;
+   my ( $inviter, $invitee, $hs_base_url, $id_server, $bind_sub ) = @_;
```
illicitonion: Unrelated: $hs_base_url is unused; we should stop passing it at some point. I can pick this up when this lands.

tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -240,24 +230,30 @@ sub do_3pid_invite {
...
+            family => "inet",
+
+            SSL_cert_file => "$DIR/../../keys/tls-selfsigned.crt",
+            SSL_key_file => "$DIR/../../keys/tls-selfsigned.key",
+         )->then_done( $id_server );
+      },
+
+      teardown => sub {
+         my ( $id_server ) = @_;
+         $loop->remove( $id_server );
```
illicitonion: This should also kill any existing connections, in case a homeserver is keep-alive-ing

illicitonion:
LGTM aside from comments

run-tests.pl
leonerd: It fails looking like:

    Running tests/10apidoc/10profile-displayname.pl...
      Testing if: PUT /profile/:user_id/displayname sets my name... PASS
      Testing if: GET /profile/:user_id/displayname publicly accessible... ABORT tests/10apidoc/10profile-displayname.pl because This Fixture has been torn down and cannot be used again at ./run-tests.pl line 350.

Resulting in eventually:

    77 tests FAILED

run-tests.pl
leonerd: It does. It calls the `teardown` accessor method on the `Fixture` structure. That just returns the value of that structure field; which is itself a `CODE` reference. It doesn't invoke the actual function referred by that.

PR #62, opened 2015-11-10
Tests for /publicRooms

illicitonion:


tests/30rooms/13anonymousaccess.pl
```diff
@@ -385,6 +385,109 @@ sub check_events
...
+            Future->needs_all(
+               matrix_set_room_guest_access( $user, $room_id, "can_join" ),
+               matrix_set_room_history_visibility( $user, $room_id, "world_readable" ),
+            );
+         }),
+      )->then( sub {
+         $http->do_request_json(
+            method => "GET",
+            uri    => "/api/v1/publicRooms",
+         )->then( sub {
```
leonerd: The point of `->then` chaining is that you don't have to indefinitely nest further and further in here. This can be

       $http->do_request(
          ...
       )
    })->then( sub {
       ...

and outdent the remainder of the test.

tests/30rooms/13anonymousaccess.pl
```diff
@@ -385,6 +385,109 @@ sub check_events
...
+                  elsif( $alias =~ m/^\Q#listingtest3:/ ) {
+                     $seen{listingtest3} = $world_readable == JSON::false && $guest_can_join == JSON::true;
+                  }
+                  elsif( $alias =~ m/^\Q#listingtest4:/ ) {
+                     $seen{listingtest4} = $world_readable == JSON::true && $guest_can_join == JSON::true;
+                  }
+               }
+            }
+
+            while ( my ( $key, $value ) = each( %seen ) ) {
```
leonerd: Don't use `each()`. It's terrible and needs to be taken out and shot.[1]

Instead, this wants to be:

    foreach my $key ( keys %seen ) {
       $seen{$key} or die "Wrong for $key";
    }

---

[1]: To store the iterator position, it abuses a slot on the underlying implementation of the actual hash. While it's technically not too bad on a small locally-scoped lexical here, it makes very brittle patterns where iteration gets broken by seemingly unrelated touches made far away. The `foreach` suggested does not suffer this problem.

tests/30rooms/13anonymousaccess.pl
```diff
@@ -385,6 +385,109 @@ sub check_events
...
+            );
+
+            foreach my $room ( @{ $body->{chunk} } ) {
+               my $aliases = $room->{aliases};
+               my $world_readable = $room->{world_readable};
+               my $guest_can_join = $room->{guest_can_join};
+
+               foreach my $alias ( @{$aliases} ) {
+                  if( $alias =~ m/^\Q#listingtest0:/ ) {
+                     $seen{listingtest0} = $world_readable == JSON::false && $guest_can_join == JSON::false;
```
leonerd: So contrary to my earlier statements, `==` comparison on JSON booleans isn't really DWIM'ing here. This is just doing boolean truth comparison, because the JSON values helpfully overload the `==` operator. Since we're really requiring on the API that these really honestly be true JSON booleans, you might want to put

    require_json_boolean( my $world_readable = $room->{world_readable} );

at the top of the loop, and just test on simple boolean truth later on; e.g.

    $seen{listeningtest0} = !$world_readable && !$guest_can_join;

tests/30rooms/13anonymousaccess.pl

tests/30rooms/13anonymousaccess.pl

leonerd:
Code LGTM.
Seems to fail though - does it depend on a branch of synapse?

illicitonion:
Yep, depends on the branch with the same name, waiting for Mjark's review :)

PR #63, opened 2015-11-11
Remove cases of postfix 'if' and 'unless' conditions in tests/

leonerd:
To match STYLE-GUIDE.rst, replace cases of `SMT if/unless COND` with `COND and/or SMT`.


leonerd:
Additionally, the filter condition tests for uses of `await_event_for()` are now explicitly doing `return 0` rather than just returning an empty value that gets treated as false.

leonerd:
Abandoning this by mutual agreement

PR #64, opened 2015-11-11
Tests for guest access to room initialSync

illicitonion:
Depends on https://github.com/matrix-org/synapse/pull/360


NegativeMjark:
LGTM

PR #65, opened 2015-11-11
Check room metadata in received invites (SYT-20)

leonerd:


tests/10apidoc/03events-initial.pl
```diff
@@ -142,7 +142,7 @@ sub await_event_for
          my $found;
          foreach my $event ( @_ ) {
             not $found and $filter->( $event ) and
```
illicitonion: Cute though the terseness of this loop is, perhaps it could be expanded into something more legible?

tests/30rooms/06invite.pl
```diff
@@ -87,6 +87,9 @@ sub inviteonly_room_fixture
          $content->{membership} eq "invite" or
             die "Expected membership to be 'invite'";
 
+         require_json_keys( $event, qw( invite_room_state ));
```
illicitonion: invite_room_state is optional; we accordingly shouldn't require this.

Opened https://github.com/matrix-org/matrix-doc/pull/160 to clarify this.

tests/30rooms/06invite.pl
```diff
@@ -154,3 +157,76 @@ sub invited_user_can_reject_invite
...
+         await_event_for( $invitee, sub {
+            my ( $event ) = @_;
+            return $event->{type} eq "m.room.member" &&
+                   $event->{room_id} eq $room_id;
+         });
+      })->then( sub {
+         my ( $event ) = @_;
+
+         require_json_keys( $event, qw( invite_room_state ));
+         my %state_by_type = map {
```
illicitonion: I think this whole block would be clearer as a for loop rather than this sequence of fetches and maps which actually follows below.

Something along the lines of:

```perl:
my @futures = ();

foreach my $event_type (...) {
   push @futures, matrix_get_room_state( $creator, $room_id, type => $event_type )->then( sub {
      my ( $room_content ) = @_;
      my $invite_content = $event->{invite_room_state}->{$event_type};
      all { $room_content->{$_} eq $invite_content->{$_} } keys %$room_content, keys %$in_invite or
         die "Content does not match for event type $event_type"; # Or better yet, assert_hash_equals( $room_content, $invite_content )
   });
   Future->needs_all( $futures );
```

tests/10apidoc/03events-initial.pl
leonerd: Done 4797156. Turns out I had neatened that style of pattern up in a utility function, then managed to forget to actually use it here.

tests/30rooms/06invite.pl
leonerd: Ah. I can make this one optional then, but then what to do of the later test below? I guess it should just stop early if it didn't find the result.

leonerd: Currently awaiting rewriting atop PR #71 which should make this alteration better.

leonerd: Done as of 0c74968

tests/30rooms/06invite.pl
leonerd: Done in ac438d5

leonerd:
PTAL

illicitonion:
LGTM

PR #67, opened 2015-11-12
More tests for guest access

illicitonion:


tests/30rooms/13anonymousaccess.pl
```diff
@@ -272,6 +290,27 @@
...
+         matrix_set_room_history_visibility( $user, $room_id, "world_readable" ),
+         matrix_set_room_guest_access( $user, $room_id, "can_join" ),
+      )->then( sub {
+         matrix_join_room( $anonymous_user, $room_id );
+      })->then( sub {
+         matrix_leave_room( $anonymous_user, $room_id );
+      })->then( sub {
+         do_request_json_for( $anonymous_user,
+            method => "GET",
+            uri    => "/api/v1/rooms/$room_id/state/m.room.member/".$user->user_id,
```
NegativeMjark: You could probably use ``:user_id``

illicitonion: I'm consciously using a different user's user ID, in case there's some weird special casing in a server which says something like "you can always find out your own membership state".

NegativeMjark: Ah, so you are, nevermind

NegativeMjark:
LGTM, other than it doesn't pass on develop yet.

PR #68, opened 2015-11-12
Cherrypick fixes to 'await_event_for()' from PR #65

leonerd:
I want to fix up some other tests before I finish work on the remaining comments on PR #65, but to fix those up requires first this (unrelated) fix to `await_event_for()`. So I've cherrypicked them here.


illicitonion:
LGTM

PR #69, opened 2015-11-12
Update tests for SPEC-254

richvdh:
Sync now returns state corresponding to the *start* of the timeline. Update the
tests to reflect that.

This code tests the changes introduced in
https://github.com/matrix-org/synapse/pull/373.

### DIFF
--- a/tests/31sync/06state.pl
+++ b/tests/31sync/06state.pl
@@ -2,6 +2,7 @@ use Future::Utils qw( repeat );
 
 # call /sync repeatedly until it returns a result
 # with an event in the given room
+# TODO: it might be good to combine this with await_event_for() at some point.
 sub wait_for_event_in_room {
     my ($user, $room_id, %params) = @_;
 
@@ -18,7 +19,7 @@ sub wait_for_event_in_room {
                           scalar @{ $room->{state}{events}})) {
                 Future->done($body);
             } else {
-                delay(0.1) -> then_done(undef);
+                delay(0.1)->then_done(undef);
             }
         });
     }, while => sub {!$_[0]->failure and !$_[0]->get});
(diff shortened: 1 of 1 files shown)


tests/31sync/06state.pl
```diff
@@ -1,4 +1,30 @@
-test "State is included in the initial sync",
+use Future::Utils qw( repeat );
+
+# call /sync repeatedly until it returns a result
+# with an event in the given room
+sub wait_for_event_in_room {
```
leonerd: Did you want to be using `await_event_for()` here instead? I'm aware that waits for an event to arrive on the event stream, rather than polling for some amount of state, but would that suffice somehow?

richvdh: Well, I could, but I specifically want to use ``/sync`` rather than ``/events``, so I'd have to have an extra call to ``/sync`` once I decided there was something worth syncing for.

That's not to say that this shouldn't be implemented in a more generic way...

tests/31sync/06state.pl
```diff
@@ -1,4 +1,30 @@
...
+        matrix_sync( $user, %{ $sync_params } )->then( sub {
+            my ( $body ) = @_;
+
+            my $room = $body->{rooms}{joined}{$room_id};
+
+            if( $room && (scalar @{ $room->{timeline}{events}} ||
+                          scalar @{ $room->{state}{events}})) {
+                Future->done($body);
+            } else {
+                delay(0.1) -> then_done(undef);
```
leonerd: Remove whitespace around `->`

tests/31sync/06state.pl
leonerd: For expediency in getting this committed now, I'd say leave it how it is. But perhaps put a TODO note in to suggest finding a way to run it using `await_event_for()` instead, so that can be neatened up at some future time.

Mmm.. technical debt.. ;)

leonerd:
LGTM asides two small comments.

PR #71, opened 2015-11-12
Merge a sequence of three `test`s into a single `multi_test`

leonerd:
Avoids inter-test related state due to shared fixtures. Makes it easier to extend the invite test in PR #65 later.

### DIFF
--- a/tests/30rooms/06invite.pl
+++ b/tests/30rooms/06invite.pl
@@ -43,16 +43,6 @@ sub inviteonly_room_fixture
 
 my $inviteonly_room_fixture = inviteonly_room_fixture( creator => $creator_fixture );
 
-test "Uninvited users cannot join the room",
-   requires => [ local_user_fixture(), $inviteonly_room_fixture ],
-
-   check => sub {
-      my ( $uninvited, $room_id ) = @_;
-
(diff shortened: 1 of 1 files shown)


tests/30rooms/06invite.pl
```diff
@@ -41,65 +41,42 @@ sub inviteonly_room_fixture
...
-         ->main::expect_http_403;
-   };
-
-my $invited_user_fixture = local_user_fixture();
-
-test "Can invite users to invite-only rooms",
-   requires => [ $creator_fixture, $invited_user_fixture, $inviteonly_room_fixture,
-                qw( can_invite_room )],
+multi_test "Can invite users to invite-only rooms",
+   requires => [ $creator_fixture, local_user_fixture(),
```
illicitonion: Any reason not to use an independent creator too? (And in the other test below)

leonerd: Ah; fair point. Done at a7ec32c. PTAL.

illicitonion:
LGTM

leonerd:
Jenkins appears unhappy but that's an unrelated test breakage - I'll merge this one in now.

PR #72, opened 2015-11-12
Add a test for searching for an event by its content.body

NegativeMjark:
Tests the search API added by matrix-org/matrix-doc#96

### DIFF
--- a/tests/43search.pl
+++ b/tests/43search.pl
@@ -1,0 +1,49 @@
+test "Can search for an event by body",
+   requires => [local_user_and_room_fixtures()],
+
+   check => sub {
+      my ( $user, $room_id ) = @_;
+
+      my ( $event_id );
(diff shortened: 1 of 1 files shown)


tests/43search.pl
```diff
@@ -0,0 +1,49 @@
+test "Can search for an event by body",
+   requires => [local_user_and_room_fixtures()],
```
leonerd: Add spaces inside `[ ... ]`

leonerd:
Stylistically LGTM

I wonder if this test and/or the spec are going to paint impls. into a corner though, because effectively we're requiring synchronous update of the FTS index. I could imagine implementations that want to farm off updating the FTS to some other system in the background; so this style of "submit then immediately search" test may not produce a result first off.

PR #73, opened 2015-11-13
Use stream tokens properly

illicitonion:
This is getting slightly closer to the territory where I should probably be using await_event_for, but... meh.

### DIFF
--- a/tests/30rooms/13anonymousaccess.pl
+++ b/tests/30rooms/13anonymousaccess.pl
@@ -97,17 +97,11 @@ test "Anonymous user can call /events on world_readable room",
             do_request_json_for( $anonymous_user,
                method => "GET",
                uri    => "/api/v1/events",
-               params => {
-                  limit => "2",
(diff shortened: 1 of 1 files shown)


tests/30rooms/13anonymousaccess.pl
```diff
@@ -120,15 +110,122 @@
...
+                     from    => $stream_token,
+                  },
+               )->then( sub {
+                  my ( $body ) = @_;
+
+                  my $chunk = ignore_presence_for( [ $anonymous_user ], @{ $body->{chunk} } );
+                  log_if_fail "chunk", $chunk;
+                  @{ $chunk } == 1 or die "Want exactly one event";
+                  my $event = $chunk->[0];
+                  log_if_fail "event", $event;
```
leonerd: There's no need logging `$event` here since it'll just be the first (only surely?) element of the `$chunk` you logged a moment ago.

PR #74, opened 2015-11-13
Use stream tokens properly

illicitonion:
This is getting slightly closer to the territory where I should probably be using await_event_for, but... meh.

### DIFF
--- a/tests/30rooms/13anonymousaccess.pl
+++ b/tests/30rooms/13anonymousaccess.pl
@@ -80,12 +80,12 @@ test "Anonymous user cannot call /events on non-world_readable room",
    };
 
 test "Anonymous user can call /events on world_readable room",
-   requires => [ anonymous_user_fixture(), local_user_fixture() ],
+   requires => [ anonymous_user_fixture(), local_user_fixture(), local_user_fixture() ],
 
    do => sub {
-      my ( $anonymous_user, $user ) = @_;
+      my ( $anonymous_user, $user, $user_not_in_room ) = @_;
 
-      my $room_id;
+      my ( $room_id, $sent_event_id );
 
       matrix_create_and_join_room( [ $user ] )
       ->then( sub {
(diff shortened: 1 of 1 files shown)


tests/30rooms/13anonymousaccess.pl
```diff
@@ -120,15 +110,121 @@
...
+                  $event->{content}->{user_id} eq $user->user_id or die "Wrong user";
+
+                  Future->done( $body->{end} );
+               }),
+
+               do_request_json_for( $user_not_in_room,
+                  method  => "PUT",
+                  uri     => "/api/v1/presence/:user_id/status",
+                  content => { presence => "online", status_msg => "Worshiping lemurs' tails" },
+               )->then( sub {
```
leonerd: Is there actually a sequencing requirement between these two presence updates? If not, it'd be neater to put them as separate components of the `needs_all`

tests/30rooms/13anonymousaccess.pl
```diff
@@ -701,3 +797,18 @@ sub matrix_get_room_membership
...
+   my @filtered;
+   foreach my $event (@raw) {
+      my $include = 1;
+      foreach my $user (@$users) {
+         $include = 0 if $event->{type} eq "m.presence" and $event->{content}->{user_id} eq $user->user_id;
+      }
+      push @filtered, $event if $include;
+   }
+   \@filtered
+}
```
leonerd: This is a manually-unrolled way of basically writing a `grep` filter with an `any` test inside. I'd suggest a shorter:

    my ( $ignored_users, @events ) = @_;

    return [ grep {
       my $event = $_;
       not(
          $event->{type} eq "m.presence" and any { $event->{content}{user_id} eq $_->user_id } @$ignored_users
       )
    } @events ];

Or maybe skip the `any` in this case and pull it into a lookup hash instead:

    my %ignored_userids = map { $_->user_id => 1 } @$ignored_users;
    ....
       not ( $event->{type} eq "m.presence" and exists $ignored_userids{ $event->{content}{user_id} } )

tests/30rooms/13anonymousaccess.pl
```diff
@@ -120,15 +110,122 @@
...
+                  log_if_fail "chunk", $chunk;
+                  @{ $chunk } == 1 or die "Want exactly one event";
+                  my $event = $chunk->[0];
+                  $event->{type} eq "m.presence" or die "Wrong event type";
+                  $event->{content}->{user_id} eq $user->user_id or die "Wrong user";
+
+                  Future->done( $body->{end} );
+               }),
+
+               Future->needs_all(
```
leonerd: This inner `needs_all` is also redundant since it's nested inside another one.
`all( ONE, all( TWO, THREE ) )` is the same as `all( ONE, TWO, THREE )`

leonerd:
LGTM aside from the `needs_all` nesting comment.

Jenkins is unhappy but I believe that's an unrelated test.

tests/30rooms/13anonymousaccess.pl

PR #77, opened 2015-11-18
s/private_user_data/account_data/

NegativeMjark:
To match the updated matrix-org/matrix-doc#132


erikjohnston:
Come back when the checks don't fail? :smile:

erikjohnston:
Those looks like changes. That probably do what you want?

PR #78, opened 2015-11-19
Make sure that the timeline comes back in the right order

richvdh:
It turns out that it does, and I was just being a crank. But no harm in testing
for it explicitly.

### DIFF
--- a/tests/31sync/04timeline.pl
+++ b/tests/31sync/04timeline.pl
@@ -4,9 +4,9 @@ test "Can sync a room with a single message",
    check => sub {
       my ( $http ) = @_;
 
-      my ( $user, $filter_id, $room_id, $event_id );
+      my ( $user, $filter_id, $room_id, $event_id_1, $event_id_2 );
 
-      my $filter = { room => { timeline => { limit => 1 } } };
+      my $filter = { room => { timeline => { limit => 2 } } };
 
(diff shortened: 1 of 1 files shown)


richvdh:
The jenkins failure is a lie.

tests/31sync/04timeline.pl
```diff
@@ -16,10 +16,16 @@
          ( $room_id ) = @_;
 
          matrix_send_room_text_message( $user, $room_id,
-            body => "A test message",
+            body => "Test message 1",
          );
       })->then( sub {
-         ( $event_id ) = @_;
+         ( $event_id_1 ) = @_;
```
leonerd: The logic in these might read a little clearer, by way of associating the `$event_id_*` variables with their respective messages, by doing

       ...
          body => "Test message 1",
       )->on_done( sub { ( $event_id_1 ) = @_ } )
   })->then( sub {
          body => "Test message 2",
       )->on_done( sub { ( $event_id_2 ) = @_ } )
       ...

leonerd:
LGTM as it stands. Or adjust it to use the `->on_done` suggestion. I'll leave that up to you.

richvdh:
Thanks. I'll leave it for now, as none of the other tests in this area work in the way you suggest. It's a good idea though; I'll bear it in mind for the future.

PR #79, opened 2015-11-19
Update tests to match /sync changes

richvdh:
Update the tests so that they work with https://github.com/matrix-org/synapse/pull/389


leonerd:
LGTM. I notice that jenkins reports a failure - I guess just due to branch mismatch.

leonerd:
FYI - fails on jenkins but now passes locally if you `git merge develop` into synapse branch.

PR #81, opened 2015-11-19
Ensure that ->main::expect_http_* is applied correctly

leonerd:
Apply it just a single HTTP request, rather than the entire ->then chain.


illicitonion:
LGTM

PR #82, opened 2015-11-19
Add tests for setting account data (SYT-30)

NegativeMjark:
Tests the API from matrix-org/matrix-doc#168

### DIFF
--- a/tests/44account_data.pl
+++ b/tests/44account_data.pl
@@ -1,0 +1,269 @@
+push our @EXPORT, qw( matrix_add_account_data );
+
+=head2 matrix_add_account_data
+
+   matrix_add_account_data( $user, $type, $content )->get;
+
+Add account data for the user.
+
+=cut
+
+sub matrix_add_account_data
+{
+   my ( $user, $type, $content ) = @_;
+
+   do_request_json_for( $user,
+      method  => "PUT",
+      uri     => "/v2_alpha/user/:user_id/account_data/$type",
+      content => $content
+   );
+}
+
+=head2 matrix_add_room_account_data
+
+    matrix_add_account_data( $user, $room_id, $type, $content )->get;
+
+Add account data for the user for a room.
+
+=cut
+
+sub matrix_add_room_account_data
+{
+   my ( $user, $room_id, $type, $content ) = @_;
+
+   do_request_json_for( $user,
+      method  => "PUT",
+      uri     => "/v2_alpha/user/:user_id/rooms/$room_id/account_data/$type",
+      content => $content
+   );
+}
+
+
+test "Can add account data",
+   requires => [ local_user_fixture() ],
+
+   check => sub {
+      my ( $user ) = @_;
+
+      matrix_add_account_data( $user, "my.test.type", {} );
+   };
+
+
+test "Can add account data to room",
+   requires => [ local_user_and_room_fixtures() ],
+
+   check => sub {
+      my ( $user, $room_id ) = @_;
+
+      matrix_add_room_account_data( $user, $room_id, "my.test.type", {} );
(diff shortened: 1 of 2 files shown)


tests/44account_data.pl
```diff
@@ -0,0 +1,269 @@
...
+
+   check => sub {
+      my ( $user ) = @_;
+
+      Future->needs_all(
+         await_event_for( $user, filter => sub {
+            my ( $event ) = @_;
+
+            return unless $event->{type} eq "my.test.type"
+               and $event->{content}{cats_or_rats} eq "cats";
```
leonerd: This wants a terminating `return 1;` at the end.

tests/44account_data.pl
```diff
@@ -0,0 +1,269 @@
...
+   check => sub {
+      my ( $user, $room_id ) = @_;
+
+      Future->needs_all(
+         await_event_for( $user, filter => sub {
+            my ( $event ) = @_;
+
+            return unless $event->{type} eq "my.test.type"
+               and $event->{content}{cats_or_rats} eq "rats"
+               and $event->{room_id} eq $room_id;
```
leonerd: Ditto `return 1;`

tests/44account_data.pl
illicitonion: Or a dropped "unless" :)

tests/44account_data.pl
```diff
@@ -0,0 +1,269 @@
...
+         ),
+         setup_incremental_account_data(
+            $user, $room_id, "my.changing.type", "dogs", "frogs"
+         ),
+      )->then( sub {
+         matrix_sync( $user );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         $next_batch = $body->{next_batch};
```
illicitonion: I would possibly Future->done( $body->{next_batch} ) and grab it in the next sub, rather than having it scoped as a variable outside the futures. But possibly also I would do exactly what you've done.

tests/44account_data.pl
```diff
@@ -0,0 +1,269 @@
...
+         check_one_account_data_event(
+            $body->{account_data}{events}, "my.changing.type", "cats"
+         );
+
+         check_one_account_data_event(
+            $body->{rooms}{joined}{$room_id}{account_data}{events},
+            "my.changing.type", "rats"
+         );
+
+         Future->done(1);
```
illicitonion: Also check that my.unchanging.type isn't present?

tests/44account_data.pl
leonerd: The generally-prevailing style around most tests is some shared captured variables at the top of the test block, that get set in some stage and read later. For consistency it's probably best to stick to that style.

Neither of them is overly nice - ideally we'd want to use the `async`/`await` syntax for a lot of these tests instead, but I haven't got around to implementing that yet... ;)

illicitonion:
LGTM modulo comments

tests/44account_data.pl
```diff
@@ -0,0 +1,269 @@
...
+         matrix_initialsync( $user );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         check_one_account_data_event(
+            $body->{account_data}, "my.test.type", "cats"
+         );
+
+         check_one_account_data_event(
+            $body->{rooms}[0]->{account_data}, "my.test.type", "rats"
```
leonerd: This inner arrow isn't required; as per STYLE-GUIDE it should be omitteed.

    $body->{rooms}[0]{account_data} ...

tests/44account_data.pl
leonerd: Ahyes, that would be neater

PR #83, opened 2015-11-19
Make test assertion helper functions neater

leonerd:
Moves test assertion helper functions into their own module, and renames them to all start with `assert_*`.

Not done here is any of the large amount of migration work to move existing tests over to using these functions, but that can happen piecewise as and when updating them generally.

A huge portion of this linecount is due to the (automated and tested)

  $ findperl | xargs sed -i 's/require_json_/assert_json_/'

so it's not quite as scary as it first looks :)


lib/SyTest/Assertions.pm
```diff
@@ -0,0 +1,209 @@
...
+
+   defined $got && defined $want && $got eq $want or
+      croak "Got ${\ pp $got }, expected ${\ pp $want } for $name";
+}
+
+=head2 assert_json_object
+
+   assert_json_object( $obj )
+
+Fails the test if C<$obj> does not represent a JSON object (i.e. is a plain
```
illicitonion: The i.e. is confusing here, as it doesn't play nicely with the "not".

Does this fail the test if it *is* a plain HASH reference, or if it *isn't*?

leonerd: Reworded in 856fca2

lib/SyTest/Assertions.pm
```diff
@@ -0,0 +1,209 @@
...
+=head2 assert_base64_unpadded
+
+   assert_base64_unpadded( $str )
+
+Fails the test if C<$str> is not a plain string, contains any characters
+invalid in a Base64 encoding, or contains the trailing C<=> padding characters.
+
+=cut
+
+sub assert_base64_unpadded
```
illicitonion: As far as I can tell, the spec doesn't actually specify a base64 character set. Fancy specifying one? :)

illicitonion:
LGTM

lib/SyTest/Assertions.pm
leonerd: Done 1ffc195

PR #85, opened 2015-11-20
Wire-test federation remote room join handshake

leonerd:
Currently uses an internal hacked-up hybrid fake federation server. I'm in the process of migrating matrix-generic code out of there and into Protocol::Matrix and friends, but for now some of the federation core logic lives here.


leonerd:
Reässigning to @NegativeMjark because illicitonion is on holiday today

NegativeMjark:
LGTM

PR #86, opened 2015-11-20
Make more use of Fixtures rather than test env.

leonerd:
These changes move more of the data-sharing between tests out of the test env. and into fixtures. This helps support the eventual two-phase "load and run" ability mentioned in https://matrix.org/jira/browse/SYT-42

This is the first stage, which moves special-use things like federation and AS. The second stage, which moves the rest of the things (like knowledge of where the actual synapse instances are running) affects a lot more tests and will be done after this.


NegativeMjark:
LGTM

PR #87, opened 2015-11-27
Reduced use of 'api_client' values from test env.

leonerd:
Further advancement towards using fixtures for all inter-test value passing, which improves the ability to run just one file at a time.


NegativeMjark:
LGTM

PR #88, opened 2015-11-30
Share HTTP API clients using fixtures rather than test env.

leonerd:
More work towards the goal of using fixtures for everything, thus making it much easier to run just one file.


illicitonion:
LGTM

Please do the main:: thing for API_CLIENTS rather than needing the our $API_CLIENTS thing we discussed in person

PR #89, opened 2015-12-01
Use \$main::FOO qualification instead of our \$FOO lexicalisation

leonerd:
For consistency with the recent `$API_CLIENTS` addition


illicitonion:
LGTM

PR #90, opened 2015-12-01
Share all remaining inter-test information using fixtures instead of test env.

leonerd:
At this point, the Fixtures now lazily create any of the required test state when it's first needed. The only remaining values living in the environment are boolean ability-test guards.


illicitonion:
LGTM

PR #91, opened 2015-12-01
Pass all remaining test configuration into 05synapse.pl using shared variables

leonerd:
At this point, all the test env. contains is booleans, which I shall tidy up next.


NegativeMjark:
LGTM

PR #92, opened 2015-12-01
Kill test environment and the provide() function; rename 'provides' to 'proves'

leonerd:
Getting ever closer to one-file run ability...


NegativeMjark:
LGTM

PR #93, opened 2015-12-01
Nicer partial runs of selected file(s)

leonerd:
* Don't run tests in skipped files even if they 'prove' things; just mark their abilities as presumed
 * Print notification when later tests depend on those presumed abilities, so if it fails you might know why
 * We can't defer pushing to `@EXPORT` any more, because of skipped tests that would export functions that later files need.


NegativeMjark:
LGTM

PR #94, opened 2015-12-01
Fix operator precedence issues in 44account_data.pl

leonerd:
The `and` and `or` operators are very low precedence; even lower than `return`. The correct form for the intended behaviour here is

    return A && B;

or

    return ( A and B );

I opted for the former.


NegativeMjark:
LGTM

PR #95, opened 2015-12-02
Support lazy starting of homeservers

leonerd:
By storing `HOMESERVER_INFO` and `API_CLIENTS` as an array of fixtures, rather than one fixture returning an array(ref) of items, we can lazily start only the relevant homeserver for local tests, in the case of running only a few files that only contain "local" tests it doesn't have to spin up a second HS that will sit idle and do nothing.


illicitonion:
Seems reasonable :) LGTM

PR #96, opened 2015-12-02
Check that state updates in the C+S API are idempotent

NegativeMjark:


tests/30rooms/01state.pl
```diff
@@ -237,3 +237,48 @@
...
+
+      set_test_state( $user, $room_id )->then( sub {
+         ( $event_id_1 ) = @_;
+
+         set_test_state( $user, $room_id );
+      })->then( sub {
+         ( $event_id_2 ) = @_;
+
+         $event_id_1 eq $event_id_2 or die "Did not expect a new event when"
+             . " uploading the same state event a second time";
```
leonerd: You can avoid that awkward-to-read string concat in the message if you

    $event_id_1 eq $event_id_2 or
       die "Did not expect ....";

As per STYLE-GUIDE it doesn't overly matter if simply the end of the message string spills over 80 chars; a reader with a terminal this narrow can infer the end of string closing quote and `;`.

illicitonion: Or assert_eq( $event_id_1, $event_id_2, "Event IDs") :)

illicitonion:
LGTM with one comment

PR #97, opened 2015-12-02
Add test for /admin/whois

illicitonion:
I accidentally pushed this to develop and reverted it.

Spec: https://github.com/matrix-org/matrix-doc/pull/198
Synapse: https://github.com/matrix-org/synapse/pull/418

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -1,0 +1,35 @@
+test "/whois",
+   requires => [ $main::API_CLIENTS[0] ],
+
+   do => sub {
+      my ( $http ) = @_;
+
+      my $user;
+
+      # Register a user, rather than using a fixture, because we want to very
+      # tightly control the actions taken by that user.
+      # Conceivably this API may change based on the number of API calls the
+      # user made, for instance.
+      matrix_register_user( $http )
+      ->then( sub {
+         ( $user ) = @_;
+
+         do_request_json_for( $user,
+            method => "GET",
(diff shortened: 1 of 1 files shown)


tests/48admin.pl
```diff
@@ -0,0 +1,31 @@
+test "/whois",
+   requires => [ $main::API_CLIENTS[0] ],
+
+   do => sub {
+      my ( $http ) = @_;
```
leonerd: Any reason this isn't

    requires => [ local_user_fixture() ],
    do => sub {
       my ( $user ) = @_;

?

tests/48admin.pl
```diff
@@ -0,0 +1,31 @@
...
+         ( $user ) = @_;
+
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/api/v1/admin/whois/".$user->user_id,
+         )
+      })->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, "devices", "user_id" );
```
leonerd: For consistency with the other tests this wants to be

    assert_json_keys( $body, qw( devices user_id ));

tests/48admin.pl
illicitonion: Added comment

tests/48admin.pl
illicitonion: Done

leonerd:
OK, that comment makes sense.

Code LGTM aside from jenkins failure, which I presume is from branch mismatch.

PR #98, opened 2015-12-02
Two-phase "load then run" split (SYT-42)

leonerd:
This change makes `run-tests.pl` load all the files (or at least, all the files up to the latest one named on the commandline in case of a partial run), before attempting to run any of them. It simplifies some of the internal logic of doing so, while at the same time giving the added convenience of instant notification of syntax errors before any of the slow tests have started. It also avoids having to print the filenames of files that contain only preparation steps and fixtures, but don't run any actual tests.


NegativeMjark:
LGTM

PR #100, opened 2015-12-04
Update the tests for archived rooms to include the flag needed to mak…

NegativeMjark:
…e the server return them

This needs to land before matrix-org/synapse#423

The flag is documented in matrix-org/matrix-doc#210


leonerd:
LGTM

richvdh:
It would be nice to test the behaviour when the flag *isn't* set.

PR #102, opened 2015-12-04
Running status display in terminal output (SYT-41)

leonerd:
With the (default) terminal output format, show a running progress display at the bottom line.
This progress display shows the number of completed vs. total tests, and the count of any failed or skipped ones.

(Also contains some miscellaneous cleanup; deleting some unused methods I found on the way)


NegativeMjark:
Probably LGTM

PR #103, opened 2015-12-08
Check that filter JSON can be passed inline as a query parameter

NegativeMjark:
Tests changes made in matrix-org/synapse#431
Documented in matrix-org/matrix-doc#224

### DIFF
--- a/tests/31sync/13inline_filter.pl
+++ b/tests/31sync/13inline_filter.pl
@@ -1,0 +1,30 @@
+use JSON qw( encode_json );
+
+test "Can pass a JSON filter as a query parameter",
+   requires => [ local_user_and_room_fixtures() ],
+
+   check => sub {
+      my ( $user, $room_id ) = @_;
+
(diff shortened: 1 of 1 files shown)


tests/31sync/13inline_filter.pl
```diff
@@ -0,0 +1,30 @@
...
+
+         my $room = $body->{rooms}{join}{$room_id};
+
+         @{ $room->{timeline}{events} } == 0
+            or die "Expected no timeline events because limit is 0";
+
+         @{ $room->{state}{events} } == 1
+            or die "Expected a single state event because of the filter";
+
+         $room->{state}{events}[0]{type} eq "m.room.member"
```
leonerd: By the way, I'm cleaning up `eq` type tests to use `assert_eq` instead, as they'll have nicer output on failures.

leonerd:
LGTM. Though you could also use `assert_eq` to save me some time later on ;)

PR #106, opened 2015-12-11
Abstract out room-based federation logic to new ::Room class

leonerd:
A work-in-progress snapshot that is heading in the direction of making it easier to write more federation tests, so I can test things like backfill.

### DIFF
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -32,7 +32,6 @@ sub create
       auth_events => [],
       content     => { creator => $creator },
       depth       => 0,
-      room_id     => $room_id,
       sender      => $creator,
       state_key   => "",
    );
@@ -43,7 +42,6 @@ sub create
       auth_events => make_event_refs( $create_event ),
       content     => { membership => "join" },
       depth       => 0,
-      room_id     => $room_id,
       sender      => $creator,
       state_key   => $creator,
    );
@@ -54,7 +52,6 @@ sub create
       auth_events => make_event_refs( $create_event, $creator_member_event ),
       content     => { join_rule => "public" },
       depth       => 0,
-      room_id     => $room_id,
       sender      => $creator,
       state_key   => "",
    );
@@ -82,6 +79,7 @@ sub create_event
    my $event = $server->create_event(
       %fields,
 
+      room_id     => $self->room_id,
       prev_events => $self->{prev_events},
    );
 
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -1,0 +1,86 @@
+package SyTest::Federation::Room;
+
+use strict;
+use warnings;
+
+sub make_auth_events
+{
+   [ map { [ $_->{event_id}, $_->{hashes} ] } @_ ];
+}
+
+sub create
+{
+   my $class = shift;
+   my %args = @_;
+
+   my $server = $args{server};
+
+   my $creator = $args{creator};
+   my $room_id = $args{room_id} // $server->next_room_id;
+
+   my $create_event = $server->create_event(
+      type => "m.room.create",
+
+      auth_events => [],
+      prev_events => [],
+      prev_state  => [],
+      content     => { creator => $creator },
+      depth       => 0,
+      room_id     => $room_id,
+      sender      => $creator,
+      state_key   => "",
+   );
+
+   my $joinrules_event = $server->create_event(
+      type => "m.room.join_rules",
+
+      auth_events => make_auth_events( $create_event ),
+      prev_events => make_auth_events( $create_event ),
+      prev_state  => [],
+      content     => { join_rule => "public" },
+      depth       => 0,
+      room_id     => $room_id,
+      sender      => $creator,
+      state_key   => "",
+   );
+
+   return bless {
+      room_id => $room_id,
+      server  => $server,
+
+      # TODO: this will want to be better
+      auth_events => make_auth_events( $create_event, $joinrules_event ),
+   }, $class;
+}
+
+sub room_id
+{
+   return $_[0]->{room_id};
+}
+
+sub make_join_protoevent
+{
+   my $self = shift;
+   my %args = @_;
+
+   my $user_id = $args{user_id};
+
+   my $server = $self->{server};
+
+   return {
+      type => "m.room.member",
+
+      auth_events      => $self->{auth_events},
+      content          => { membership => "join" },
+      depth            => 0,
+      event_id         => my $join_event_id = $server->next_event_id,
+      origin           => $server->server_name,
+      origin_server_ts => $server->time_ms,
+      prev_events      => [],
+      room_id          => $self->room_id,
+      sender           => $user_id,
+      state_key        => $user_id,
+   };
+}
+
+1;
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -3,7 +3,7 @@ package SyTest::Federation::Room;
 use strict;
 use warnings;
 
-sub make_auth_events
+sub make_event_refs
 {
    [ map { [ $_->{event_id}, $_->{hashes} ] } @_ ];
 }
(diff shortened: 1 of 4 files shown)


tests/50federation/30room-join.pl
```diff
@@ -12,13 +9,18 @@ sub make_auth_events
       my ( $user, $inbound_server ) = @_;
       my $local_server_name = $inbound_server->server_name;
 
+      my $creator = '@50fed:' . $local_server_name;
+
+      my $room = SyTest::Federation::Room->create(
+         server  => $inbound_server,
```
illicitonion: As per discussion on matrix, this is quite a strange line of code which I don't know how to interpret.

lib/SyTest/Federation/Server.pm
```diff
@@ -96,6 +103,34 @@ sub get_event
...
+   my @event_ids = @_;
+
+   my %events_by_id = map { $_ => $self->get_event( $_ ) } @event_ids;
+
+   my @auth_chain = @event_ids;
+
+   while( @event_ids ) {
+      my $event = $events_by_id{shift @event_ids};
+
+      my @auth_ids = map { $_->[0] } @{ $event->{auth_events} };
```
illicitonion: It would be nice if auth_events were a struct with named values, rather than knowing by convention that the 0th element happens to be an array of auth event IDs.

lib/SyTest/Federation/Server.pm
```diff
@@ -96,6 +103,34 @@ sub get_event
...
+         next if $events_by_id{$id};
+
+         $events_by_id{$id} = $self->get_event( $id );
+         push @event_ids, $id;
+      }
+
+      unshift @auth_chain, @auth_ids;
+   }
+
+   # uniqify by first occurence
```
illicitonion: It could be a little nicer to build this up uniquely (the only difference would be de-duping on line 124, I think?) rather than build it it blindly and uniqueify at the end.

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
+package SyTest::Federation::Room;
+
+use strict;
+use warnings;
+
+sub make_event_refs
+{
+   [ map { [ $_->{event_id}, $_->{hashes} ] } @_ ];
```
illicitonion: Name these things, rather than relying on position?

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+}
+
+sub create
```
illicitonion: pod please

lib/SyTest/Federation/Room.pm
leonerd: That structure is part of Matrix federation protocol itself. The protocol says that the `prev_events` and `auth_events` are lists of 2-element lists, each element giving an event ID and a hash.

lib/SyTest/Federation/Server.pm
leonerd: It wouldn't quite be, because here I'm starting at the "most recent" events and walking backwards chronologically via each implied earlier event, building the full auth_chain list backwards. I then uniq it by first occurrence, ensuring that no event appears before the things that would authenticate it; as would happen were it uniq'ed as it went.

Though I suppose it might be nicer to do

    @auth_chain = uniq @auth_ids, @auth_chain;

incrementally here at least.

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+      depth       => 0,
+      sender      => $creator,
+      state_key   => "",
+   );
+
+   $self->create_event(
+      type => "m.room.member",
+
+      content     => { membership => "join" },
+      depth       => 0,
```
illicitonion: I would naively assume that the join and join_rules would be deeper than the creation. Is this not the case?

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+{
+   my $self = shift;
+   my %fields = @_;
+
+   my $server = $self->{server};
+
+   $fields{prev_state} = [] if defined $fields{state_key}; # TODO: give it a better value
+
+   my @auth_events = grep { defined }
+      $self->get_current_state_event( "m.room.create" ),
```
illicitonion: ()s around the array you're ?implicitly constructing would be nice, if they're allowed

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+sub get_current_state_event
+{
+   my $self = shift;
+   my ( $type, $state_key ) = @_;
+   $state_key //= "";
+
+   return $self->{current_state}{ join "\0", $type, $state_key };
+}
+
+sub make_join_protoevent
```
illicitonion: pod please

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+{
+   my $self = shift;
+   my %args = @_;
+
+   my $user_id = $args{user_id};
+
+   my $server = $self->{server};
+
+   my @auth_events = grep { defined }
+      $self->get_current_state_event( "m.room.create" ),
```
illicitonion: ()s if allowed

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+sub make_join_protoevent
+{
+   my $self = shift;
+   my %args = @_;
+
+   my $user_id = $args{user_id};
+
+   my $server = $self->{server};
+
+   my @auth_events = grep { defined }
```
illicitonion: Surely if these aren't defined, it's an error?

lib/SyTest/Federation/Room.pm
```diff
@@ -0,0 +1,138 @@
...
+      $self->get_current_state_event( "m.room.create" ),
+      $self->get_current_state_event( "m.room.join_rules" );
+
+   return {
+      type => "m.room.member",
+
+      auth_events      => make_event_refs( @auth_events ),
+      content          => { membership => "join" },
+      depth            => 0,
```
illicitonion: Surely depth => max { map { { $_->{depth} }, @all_events } } + 1 ?

lib/SyTest/Federation/Server.pm
leonerd: That would indeed be nice. :) However, it's a data structure baked into the Matrix federation protocol so there's not a lot we can do about that here.

lib/SyTest/Federation/Room.pm
leonerd: Added much docs.

lib/SyTest/Federation/Room.pm
leonerd: Done 3257c70

lib/SyTest/Federation/Room.pm
leonerd: Done 3257c70

lib/SyTest/Federation/Room.pm
leonerd: Most likely; but I was trying to be super-defensive here. Sometime we'll want to use this code to generate known-bad data to check that the SUT rejects it.

lib/SyTest/Federation/Room.pm
leonerd: You would think. :) However, it's part of the ongoing list of bugs in SYN-507 to note that currently this isn't being checked. I'm slowly fixing up "bugs" in the tests such as this simultaneously with fixing up the implementation to check for them.

illicitonion: Add a comment, then?

leonerd: Comments added in 14d9ce5

lib/SyTest/Federation/Server.pm
leonerd: Adjusted in ea1737b

leonerd:
PTAL

illicitonion:
LGTM

PR #108, opened 2015-12-16
Split "data storage" out of SY:F:Server into new ::Datastore

leonerd:
The `::Server` class was getting increasingly messy, storing events and rooms, allocating event and room IDs, associating room aliases, etc... This series of commits makes a new `::Datastore` object to store all the mocked testing data, and returns `::Server` to the simpler state of just being a server.

What's more, the new methods even have docs *gasp* *applause* etc...

### DIFF
--- a/lib/SyTest/Federation/Datastore.pm
+++ b/lib/SyTest/Federation/Datastore.pm
@@ -52,8 +52,7 @@ sub key_id { $_[0]->{key_id} }
 
    $key = $store->secret_key
 
-Return the public or secret halves of the signing key the server is currently
-using
+Return the public or secret halves of the signing key of the local homeserver.
 
 =cut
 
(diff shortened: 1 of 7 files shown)


lib/SyTest/Federation/Datastore.pm
```diff
@@ -0,0 +1,250 @@
...
+
+=head2 put_key
+
+   $store->put_key( server_name => $name, key_id => $id, key => $key )
+
+Accessor and mutator for federation key storage
+
+=cut
+
+sub get_key
```
illicitonion: How is get_key different from public_key or secret_key?

leonerd: It relates to our local cache of remote server (federation) signing keys. I'll add to docs.

leonerd: Done adc1eeb

illicitonion:
LGTM

PR #109, opened 2015-12-16
Test sending and receiving transactions containing room messages

leonerd:
An initial hack at a pair of tests to send or receive a transaction containing an `m.room.message` for a federated room. Deliberately engineered with most of the fancy parts in the reusable library implementation so that we can have lots more tests without too much more fuss.

### DIFF
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -84,7 +84,7 @@ sub do_request_json
    );
 }
 
-sub send_edu
+sub send_transaction
 {
    my $self = shift;
    my %params = @_;
@@ -95,15 +95,8 @@ sub send_edu
       origin           => $self->server_name,
       origin_server_ts => JSON::number( $ts ),
       previous_ids     => [], # TODO
-      pdus             => [],
-      edus             => [
-         {
-            edu_type => $params{edu_type},
-            content  => $params{content},
-            origin   => $self->server_name,
-            destination => $params{destination},
-         }
-      ],
+      pdus             => $params{pdus} // [],
+      edus             => $params{edus} // [],
    );
 
    $self->do_request_json(
@@ -113,6 +106,35 @@ sub send_edu
 
       content => \%transaction,
    )->then_done(); # response body is empty
+}
+
+sub send_edu
+{
+   my $self = shift;
+   my %params = @_;
+
+   $self->send_transaction(
+      %params,
+      edus => [
+         {
+            edu_type => $params{edu_type},
+            content  => $params{content},
+            origin   => $self->server_name,
+            destination => $params{destination},
+         }
+      ],
+   );
+}
+
+sub send_event
+{
+   my $self = shift;
+   my %params = @_;
+
+   $self->send_transaction(
+      %params,
+      pdus => [ $params{event} ],
+   );
 }
 
 sub join_room
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -5,6 +5,8 @@ use warnings;
 
 use base qw( SyTest::Federation::_Base SyTest::HTTPClient );
 
+use List::UtilsBy qw( uniq_by );
+
 use MIME::Base64 qw( decode_base64 );
 use HTTP::Headers::Util qw( join_header_words );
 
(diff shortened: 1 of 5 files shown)


lib/SyTest/Federation/Client.pm
```diff
@@ -105,14 +124,78 @@ sub send_edu
...
+         $join_body = $join_body->[1] if ref $join_body eq "ARRAY";
+
+         my $room = SyTest::Federation::Room->new(
+            datastore => $store,
+            room_id   => $room_id,
+         );
+
+         my %done_event;
+         foreach my $event ( @{ $join_body->{auth_chain} }, @{ $join_body->{state} } ) {
+            $done_event{ $event->{event_id} }++ or
```
illicitonion: What are you trying to do here?

lib/SyTest/Federation/Datastore.pm
```diff
@@ -246,4 +251,71 @@ sub next_room_id
...
+=head2 get_alias
+
+   $room_id = $store->get_alias( $room_alias )
+
+Returns the room ID associated with the given room alias, if one exists, or
+C<undef> if not.
+
+=cut
+
+sub get_alias
```
illicitonion: Maybe lookup_alias? Or get_room_by_alias? I generally expect get functions to return the thing named after get, but this doesn't get an alias, it gets a room.

lib/SyTest/Federation/Room.pm
```diff
@@ -167,14 +167,36 @@ sub create_event
...
+{
+   my $self = shift;
+   my ( $event ) = @_;
+
+   $self->_insert_event( $event );
+
+   my $prev_events = $self->{prev_events};
+   push @$prev_events, $event;
+
+   my %remove = map { $_->[0] => 1 } @{ $event->{prev_events} };
```
illicitonion: I've stared at these two lines for a while and I can't work out why you're doing them. Could you explain? :)

lib/SyTest/Federation/Server.pm
```diff
@@ -225,6 +226,67 @@ sub on_request_key_v2_server
...
 }
 
+sub on_request_federation_v1_query_directory
+{
+   my $self = shift;
+   my ( $req, $alias ) = @_;
+
+   my $room_id = $self->{datastore}->get_alias( $alias ) or
+      return Future->done( response => HTTP::Response->new(
+         404, "Not found", [ Content_length => 0 ], "",
```
illicitonion: Does federation not require JSON responses, like client-server does?

lib/SyTest/Federation/Server.pm
```diff
@@ -225,6 +226,67 @@ sub on_request_key_v2_server
...
+      ) );
+
+   Future->done( json => {
```
illicitonion: Why json => ?

tests/50federation/31room-send.pl
```diff
@@ -0,0 +1,93 @@
+test "Outbound federation can send events",
+   requires => [ local_user_fixture(), $main::INBOUND_SERVER ],
+
+   do => sub {
+      my ( $user, $inbound_server ) = @_;
+
+      my $local_server_name = $inbound_server->server_name;
+      my $datastore         = $inbound_server->datastore;
+
+      my $creator = '@50fed:' . $local_server_name;
```
illicitonion: Why a hard-coded user?

lib/SyTest/Federation/Client.pm
leonerd: Add each event only once. There'll possibly be duplicates between the auth chain and the current state, so this deduplicates them by event ID. Alternatively I suppose I could

    uniq_by { $_->{event_id} } @events...

lib/SyTest/Federation/Datastore.pm
leonerd: Done f1b1ea8

lib/SyTest/Federation/Room.pm
leonerd: Commented/adjusted code in 7d3137b

lib/SyTest/Federation/Server.pm
leonerd: The spec is rather vague on the subject - neither the word `failure` nor `error` appears at all in http://matrix.org/docs/spec/r0.0.0/server_server.html
Empirically it is observed to be happy with empty-length errors.

lib/SyTest/Federation/Server.pm
leonerd: To tell the responder to JSON encode this result. Not all results are JSON-encoded (because of the media server, etc...)

https://github.com/matrix-org/sytest/blob/develop/lib/SyTest/Federation/Server.pm#L109

tests/50federation/31room-send.pl
leonerd: It's a (in fact the) user on the "internal" fake federation server that SyTest creates. There isn't another way to create those currently.

lib/SyTest/Federation/Client.pm
illicitonion: I would definitely prefer the uniq_by

illicitonion:
LGTM except for uniq_by

lib/SyTest/Federation/Client.pm
leonerd: Done 549fb1e

PR #110, opened 2015-12-17
Add a test for paginated search results

richvdh:
This tests the fix added at matrix-org/synapse#447

### DIFF
--- a/tests/43search.pl
+++ b/tests/43search.pl
@@ -132,6 +132,31 @@ test "Can back-paginate search results",
             $results->[9]{result}{event_id} eq $event_ids[0]
                 or die "Final result incorrect";
 
+            my $next_batch = $room_events->{next_batch};
+
+            do_request_json_for( $user,
+                                 method  => "POST",
+                                 uri     => "/api/v1/search",
+                                 params  => { next_batch => $next_batch },
+                                 content => $search_query,
+            );
+        })->then( sub {
+            my ( $body ) = @_;
+
+            log_if_fail "Third search result:", $body;
+
+            assert_json_keys( $body, qw( search_categories ) );
+            assert_json_keys( $body->{search_categories}, qw ( room_events ) );
+            my $room_events = $body->{search_categories}{room_events};
+
+            assert_json_keys( $room_events, qw( count results ) );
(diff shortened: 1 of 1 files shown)


richvdh:
... and it fails because the fix isn't on develop yet.

tests/43search.pl
```diff
@@ -53,3 +54,109 @@
...
+            assert_json_keys( $body->{search_categories}, qw ( room_events ) );
+            my $room_events = $body->{search_categories}{room_events};
+
+            assert_json_keys( $room_events, qw( count results next_batch ) );
+
+            $room_events->{count} == 20 or die "Expected count == 20";
+
+            my $results = $room_events->{results};
+            scalar @$results == 10 or die "Expected 10 search results";
+            $results->[0]{result}{event_id} eq $event_ids[19]
```
leonerd: These can be done with

    assert_eq( $results->[0]{result}{event_id}, $event_ids[19],
       'first result event_id' );

which prints the expected/actual values on failure

tests/43search.pl
```diff
@@ -53,3 +54,109 @@
...
+                }
+            }
+        };
+
+        repeat( sub {
+            my $msgnum = $_[0];
+            matrix_send_room_text_message( $user, $room_id,
+                                           body => "Message number $msgnum" )
+                ->on_done( sub { ($event_ids[$msgnum]) = @_ } )
+        }, foreach => [ 0 .. 19 ] ) -> then ( sub {
```
leonerd: Whitespace:

    ( $event_ids[$msgnum] ) = @_

and

    )->then( sub {

leonerd:
Overall: the test appears to be expecting that search results come in batches of 10 each, but the search parameters didn't request exactly 10. This feels like it's a fragile assumption of implementation. Is there perhaps a limit parameter that can be specified?

tests/43search.pl
richvdh: that's new!

Done.

tests/43search.pl
richvdh: Done.

richvdh:
good point about the implicit limit. Now made explicit.

leonerd:
LGTM

richvdh:
ta

PR #114, opened 2015-12-17
Added a test for inbound federation's /event/:event_id endpoint

leonerd:
Can't easily test outbound version because it's hard to provoke synapse into making such a request using only the client APIs. But at least we can test inbound version.

### DIFF
--- a/tests/50federation/32room-getevent.pl
+++ b/tests/50federation/32room-getevent.pl
@@ -1,0 +1,49 @@
+test "Inbound federation can return events",
+   requires => [ $main::OUTBOUND_CLIENT, $main::HOMESERVER_INFO[0],
+                 local_user_and_room_fixtures() ],
+
+   do => sub {
+      my ( $outbound_client, $info, $user, $room_id ) = @_;
+      my $first_home_server = $info->server_name;
+
+      my $local_server_name = $outbound_client->server_name;
+
+      my $user_id = "\@50fed-user:$local_server_name";
+
+      my $member_event;
+
+      $outbound_client->join_room(
+         server_name => $first_home_server,
+         room_id     => $room_id,
+         user_id     => $user_id,
+      )->then( sub {
+         my ( $room ) = @_;
+
+         $member_event = $room->get_current_state_event( "m.room.member", $user_id );
+         log_if_fail "Member event", $member_event;
+
+         $outbound_client->do_request_json(
+            method   => "GET",
+            hostname => $first_home_server,
+            uri      => "/event/$member_event->{event_id}/",
+         );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( origin origin_server_ts pdus ));
+         assert_json_list( my $events = $body->{pdus} );
+
+         @$events == 1 or
+            die "Expected 1 event, found " . scalar(@$events);
+         my ( $event ) = @$events;
+
+         log_if_fail "Retrieved event", $event;
+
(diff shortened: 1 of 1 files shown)


tests/50federation/32room-getevent.pl
```diff
@@ -0,0 +1,49 @@
...
+   requires => [ $main::OUTBOUND_CLIENT, $main::HOMESERVER_INFO[0],
+                 local_user_and_room_fixtures() ],
+
+   do => sub {
+      my ( $outbound_client, $info, $user, $room_id ) = @_;
+      my $first_home_server = $info->server_name;
+
+      my $local_server_name = $outbound_client->server_name;
+
+      my $user_id = "\@50fed-user:$local_server_name";
```
illicitonion: We should really make this be a constant somewhere, rather than repeating it eeeeeeverywhere

tests/50federation/32room-getevent.pl
```diff
@@ -0,0 +1,49 @@
...
+            hostname => $first_home_server,
+            uri      => "/event/$member_event->{event_id}/",
+         );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( origin origin_server_ts pdus ));
+         assert_json_list( my $events = $body->{pdus} );
+
+         @$events == 1 or
```
illicitonion: log_if_fail?

illicitonion: Oh, you do it after the die, not before. Maybe do it before the die?

tests/50federation/32room-getevent.pl
```diff
@@ -0,0 +1,49 @@
...
+            die "Expected 1 event, found " . scalar(@$events);
+         my ( $event ) = @$events;
+
+         log_if_fail "Retrieved event", $event;
+
+         # Check that the string fields seem right
+         assert_eq( $event->{$_}, $member_event->{$_},
```
illicitonion: As far as I can tell, the only reason for use of the C-S API here is to verify that the C-S and S-S APIs return the same.

Vaguely open-ended thought: It's possible a server may implement one of the APIs but not the other. Could it be nice to try to keep their tests independent where possible, and then have a separate set of tests saying "when the C-S does foo, the S-S does bar"?

illicitonion:
LGTM

illicitonion:
Also, presumably you could test outbound by doing a join, then sending a message, and making sure the message appears on the other server?

tests/50federation/32room-getevent.pl
leonerd: Hmm.. There's no great requirement for it to be the same though. I guess either make it a constant so it actually is, or dynamically allocate them so it's clear they're different. Probably easier to do the latter in fact. Would you mind if I do that on a separate PR?

tests/50federation/32room-getevent.pl
leonerd: Moved 075aa33

tests/50federation/32room-getevent.pl
leonerd: There's no C-S API here. The call to `$room->get_current_state_event` is a call on the Federation::Room object to recall the state event it was initialised with in the federation client.

leonerd:
That would still be testing inbound federation, in that synapse would be responding to an `/event` request made by sytest. However, Mjark suggests a way to provoke synapse into making such a request itself, so I'll add that in another test.

PR #115, opened 2015-12-18
Include branch coverage for synapse

NegativeMjark:


leonerd:
LGTM

NegativeMjark:
Turns out that it was harder than I thought it would be to get the branch coverage into our CI. I'm abandoning this until I work out how to do this without breaking our CI.

PR #117, opened 2015-12-18
Silently ignore otherwise-unhandled incoming m.presence EDUs

leonerd:
Otherwise these spam up the output of federation tests and make it harder to see the real errors.


NegativeMjark:
LGTM

PR #118, opened 2015-12-18
Explicitly set an avatar url for the tests that check for it

NegativeMjark:


tests/10apidoc/01register.pl
```diff
@@ -136,6 +136,17 @@ sub local_user_fixture
             )->then_done( $user );
          })->then_with_f( sub {
             my $f = shift;
+            return $f unless defined( my $avatar_url = $args{avatar_url} );
+
+            my $user = $f->get;
+            do_request_json_for( $user,
```
leonerd: If we were optimising for overall runtime speed we might consider `Future->needs_all` for concurrency here; but this is fine enough to solve the problem initially.

tests/31sync/05presence.pl
```diff
@@ -34,6 +34,8 @@
 test "User is offline if they set_presence=offline in their sync",
    requires => [ local_user_fixture( with_events => 0 ),
                  qw( can_sync ) ],
+   # This test passes if we set a displayname for the user.
+   expect_fail => 1,
```
leonerd: Hmm. Can we raise a bug (SYN or SYT) for this failure, and change it to `bug => "BUGNAME"`? That indicates more accurately what the failure actually is.

leonerd:
LGTM as a quick hack but please raise bug soon and adjust.

PR #119, opened 2015-12-21
Test the federation `/backfill` API

leonerd:


tests/50federation/33room-get-missing-events.pl
```diff
@@ -0,0 +1,140 @@
...
+         $missing_event = $room->create_event(
+            type => "m.room.message",
+
+            sender  => $user_id,
+            content => {
+               body => "Message 1",
+            },
+         );
+
+         # Generate another one and do send it so it will refer to the
```
illicitonion: bleh, I don't like that I have no reason to believe that this will add a created-but-unsent event to the prev_events of this event. Could this be made explicit, perhaps by taking an additional_prev_events kwarg, and explicitly passing $missing_event->event_id to it?

tests/50federation/33room-get-missing-events.pl
```diff
@@ -0,0 +1,140 @@
...
+         Future->needs_all(
+            $inbound_server->await_get_missing_events( $room_id )
+            ->then( sub {
+               my ( $req ) = @_;
+               my $body = $req->body_from_json;
+
+               assert_json_keys( $body, qw( earliest_events latest_events limit ));
+               # TODO: min_depth but I have no idea what it does
+
+               assert_json_list( $body->{earliest_events} );
```
illicitonion: Surely we should be asserting on their values, not just their presence?

lib/SyTest/Federation/Datastore.pm
```diff
@@ -200,6 +200,64 @@ sub create_event
...
+=over 4
+
+=item limit => $limit
+
+Gives the maximum number of events that should be returned.
+
+=item stop_before => \@ids
+
+Gives a list of event IDs that should not be entered into or returned (most
+likely because the caller already has them). These events do not count towards
```
illicitonion: Do the start_at events count towards the limit? Presumably so?

lib/SyTest/Federation/Datastore.pm
```diff
@@ -200,6 +200,64 @@ sub create_event
...
+
+   my @events;
+   while( @event_ids and @events < $limit ) {
+      my $id = shift @event_ids;
+      my $event = eval { $self->get_event( $id ) }
+         or next;
+
+      push @events, $event;
+
+      push @event_ids, grep { !$stop_before{$_} }
```
illicitonion: These should be deduped, right?

lib/SyTest/Federation/Datastore.pm
```diff
@@ -200,6 +200,64 @@ sub create_event
...
+                       map { $_->[0] } @{ $event->{prev_events} };
+      $stop_before{$id}++;
```
illicitonion: I'm not sure why this is getting incremented?

tests/50federation/33room-get-missing-events.pl
leonerd: Done 4551fad

tests/50federation/33room-get-missing-events.pl
leonerd: Oops; I think I intended to test those and then forgot. Added b328cd6

lib/SyTest/Federation/Datastore.pm
leonerd: See below :)

lib/SyTest/Federation/Datastore.pm
leonerd: By adding each encountered event into the `%stop_before` hash, we'll ensure we don't recurse into it a second time, effectively applying the deduplication.

lib/SyTest/Federation/Datastore.pm
leonerd: Yes. The intention here was more to suggest that the `start_at` list was included in the output, but the `stop_before` list was not.

lib/SyTest/Federation/Datastore.pm
illicitonion: Perhaps change this to:

```perl
my %exclude = map { $_ => 1 } @{ $params{stop_before} // [] };
```
because you're changing the meaning of the stop_before hash to really be a set of things not to include.

And here, $exclude{$id} = 1; rather than ++, because your semantics here are adding to the set (boolean) not counting.

leonerd:
Documentation for the API being tested here is being added at  https://github.com/matrix-org/matrix-doc/pull/240

lib/SyTest/Federation/Datastore.pm
leonerd: Done 1cd4c87

illicitonion:
LGTM

leonerd:
Merging anyway though Jenkins is unhappy (upset about an unrelated flaky test that fails due to `m.presence` non-deterministic event ordering)

PR #121, opened 2015-12-21
Tests for returning room avatar URLs in /publicRooms

illicitonion:
Spec: https://github.com/matrix-org/matrix-doc/pull/244
Impl: https://github.com/matrix-org/synapse/pull/453

### DIFF
--- a/tests/30rooms/60anonymousaccess.pl
+++ b/tests/30rooms/60anonymousaccess.pl
@@ -745,6 +745,87 @@ test "GET /publicRooms lists rooms",
       });
    };
 
+test "GET /publicRooms includes avatar URLs",
+   requires => [ $main::API_CLIENTS[0], local_user_fixture() ],
+
+   check => sub {
+      my ( $http, $user ) = @_;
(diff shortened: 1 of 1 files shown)


NegativeMjark:
LGTM

tests/30rooms/60anonymousaccess.pl
```diff
@@ -687,6 +687,87 @@ sub check_events
...
+test "GET /publicRooms includes avatar URLs",
+   requires => [ $main::API_CLIENTS[0], local_user_fixture() ],
+
+   check => sub {
+      my ( $http, $user ) = @_;
+
+      Future->needs_all(
+         matrix_create_room( $user,
+            visibility => "public",
+            room_alias_name => "nonworldreadable",
```
leonerd: There's a slight danger of these "unique" alias names becoming fragile across many test files.
I don't have any better suggestion currently, but perhaps something to think about for the future. Maybe a fixture for allocating new unique alias names?

PR #122, opened 2015-12-22
Check that guest users can call /sync

NegativeMjark:
Needs matrix-org/synapse#455

### DIFF
--- a/tests/30rooms/60anonymousaccess.pl
+++ b/tests/30rooms/60anonymousaccess.pl
@@ -215,7 +215,7 @@ test "Annonymous user can call /sync on a world readable room",
    do => sub {
       my ( $anonymous_user, $user ) = @_;
 
-      my ($room_id, $sent_event_id);
+      my ( $room_id, $sent_event_id );
 
       matrix_create_and_join_room( [ $user ] )
       ->then( sub {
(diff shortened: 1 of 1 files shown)


tests/30rooms/60anonymousaccess.pl
```diff
@@ -189,6 +209,46 @@ sub await_event_not_presence_for
       });
    };
 
+test "Annonymous user can call /sync on a world readable room",
+   requires => [ anonymous_user_fixture(), local_user_fixture() ],
+
+   do => sub {
+      my ( $anonymous_user, $user ) = @_;
+
+      my ($room_id, $sent_event_id);
```
leonerd: Whitespace inside `( parens )`

tests/30rooms/60anonymousaccess.pl
```diff
@@ -189,6 +209,46 @@ sub await_event_not_presence_for
...
+               timeline => { types => ["m.room.message"] },
+            },
+            presence => { types => [] }
+         }));
+      })->then( sub {
+         my ( $sync_body ) = @_;
+
+         assert_eq(
+            $sent_event_id,
+            $sync_body->{rooms}{join}{$room_id}{timeline}{events}[0]{event_id},
```
leonerd: This is quite a deep search. I'd be tempted to break it up a little so if it fails, you get some idea why:

    assert_json_object( my $room = $sync_body->{rooms}{join}{$room_id} );
    assert_json_list( my $events = $room->{timeline}{events} );
    assert_eq( $events->[0]{event_id}, $sent_event_id, 'event id' );

It's pretty arbitrary but if it fails the line number might help determine which bit

leonerd:
LGTM

PR #125, opened 2015-12-23
Fix broken guest access test.

NegativeMjark:
The test could only pass because an auth check was being skipped in
synapse.


richvdh:
lgtm

PR #126, opened 2015-12-23
List users to allow presence for, rather than ignore

NegativeMjark:
Hopefully should make some of the tests less flaky

### DIFF
--- a/tests/30rooms/60anonymousaccess.pl
+++ b/tests/30rooms/60anonymousaccess.pl
@@ -62,13 +62,17 @@ test "Anonymous user cannot call /events on non-world_readable room",
 
 sub await_event_not_presence_for
 {
-   my ( $user, $room_id, $ignored_users ) = @_;
+   my ( $user, $room_id, $allowed_users ) = @_;
    await_event_for( $user,
       room_id => $room_id,
       filter  => sub {
          my ( $event ) = @_;
-         return not( $event->{type} eq "m.presence" and
-            any { $event->{content}{user_id} eq $_->user_id } @$ignored_users );
+
+         # Include all events where the type is not m.presence.
+         # If the type is m.presence, then only include it if it is for one of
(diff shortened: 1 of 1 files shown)


tests/30rooms/60anonymousaccess.pl
```diff
@@ -62,13 +62,13 @@
...
-   my ( $user, $room_id, $ignored_users ) = @_;
+   my ( $user, $room_id, $allowed_users ) = @_;
    await_event_for( $user,
       room_id => $room_id,
       filter  => sub {
          my ( $event ) = @_;
-         return not( $event->{type} eq "m.presence" and
-            any { $event->{content}{user_id} eq $_->user_id } @$ignored_users );
+         return ((not $event->{type} eq "m.presence") or
+            any { $event->{content}{user_id} eq $_->user_id } @$allowed_users);
```
richvdh: this makes my head hurt. Can you document the behaviour of allowed_users?

tests/30rooms/60anonymousaccess.pl
```diff
@@ -62,13 +62,17 @@
...
    await_event_for( $user,
       room_id => $room_id,
       filter  => sub {
          my ( $event ) = @_;
-         return not( $event->{type} eq "m.presence" and
-            any { $event->{content}{user_id} eq $_->user_id } @$ignored_users );
+
+         # Include all events where the type is not m.presence.
+         # If the type is m.presence, then only include it if it is for one of
+         # the allowed users
```
richvdh: It would be nice if we could add parameter documentation at the top of the function. This is better, though - at least I can see the intention now.

richvdh:
as per our IRL convo, I'm not loving the fact that we're now ignoring unexpected presence events; but fine, it's better we don't have flaky tests to make the test suite useless.

LGishTM

PR #127, opened 2016-01-04
Fixtures for allocating room aliases (and localpart names)

leonerd:
More robust tests by allocating aliases (and alias names) automatically rather than hardcoding them in the files and risking a collision, or a mismatch.

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -122,3 +122,31 @@ sub matrix_create_room
       Future->done( $body->{room_id}, $body->{room_alias} );
    });
 }
+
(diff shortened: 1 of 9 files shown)


illicitonion:
These things seem strange to be fixtures rather than just global functions. Why are they fixtures?

leonerd:
Partly for symmetry with everything else, but also because there's a few places we have a combined `check` and `do` block, so it would be nice to allocate the ID outside the block of code there.

While it could be done with a variable outside the test, that would mean it's always run even if the test is skipped. No great problem in this case other than skipped numbers, but it might seem odd to be different to all the others.

illicitonion:
LGTM

PR #128, opened 2016-01-05
Test support for inviting 3pids in /createRoom

erikjohnston:
ihavenoidea

https://github.com/matrix-org/matrix-doc/pull/247
https://github.com/matrix-org/synapse/pull/460

### DIFF
--- a/tests/30rooms/12thirdpartyinvite.pl
+++ b/tests/30rooms/12thirdpartyinvite.pl
@@ -81,8 +81,8 @@ test "Can invite existing 3pid in createRoom",
                my ( $body ) = @_;
 
                log_if_fail "Body", $body;
-               $body->{membership} eq "invite" or
-                  die "Expected invited user membership to be 'invite'";
+               assert_eq( $body->{membership}, "invite",
+                  'invited user membership' );
             });
          });
       });
--- a/tests/30rooms/12thirdpartyinvite.pl
+++ b/tests/30rooms/12thirdpartyinvite.pl
@@ -66,8 +66,8 @@ test "Can invite existing 3pid in createRoom",
       $id_server->bind_identity( undef, "email", $invitee_email, $invitee )
       ->then( sub {
          my $invite_info = {
-            medium => "email",
-            address => $invitee_email,
+            medium    => "email",
+            address   => $invitee_email,
             id_server => $id_server->name,
          };
          matrix_create_room( $inviter, invite_3pid => [ $invite_info ] )
--- a/tests/30rooms/12thirdpartyinvite.pl
+++ b/tests/30rooms/12thirdpartyinvite.pl
@@ -53,6 +53,42 @@ test "Can invite existing 3pid",
       });
    };
 
+test "Can invite existing 3pid in createRoom",
+   requires => [ local_user_fixtures( 2 ), id_server_fixture() ],
+
+   do => sub {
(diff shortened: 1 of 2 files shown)


tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -53,6 +53,42 @@
...
+      my $invitee_mxid = $invitee->user_id;
+
+      my $room_id;
+
+      $id_server->bind_identity( undef, "email", $invitee_email, $invitee )
+      ->then( sub {
+         my $invite_info = {
+            medium => "email",
+            address => $invitee_email,
+            id_server => $id_server->name,
```
leonerd: Please align the three '=>'s

tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -53,6 +53,42 @@
...
+            ( $room_id ) = @_;
+
+            matrix_get_room_state( $inviter, $room_id,
+               type      => "m.room.member",
+               state_key => $invitee_mxid,
+            )->on_done( sub {
+               my ( $body ) = @_;
+
+               log_if_fail "Body", $body;
+               $body->{membership} eq "invite" or
```
leonerd: This line is better done using `assert_eq` now, because it prints clearer messages when it fails

    assert_eq( $body->{membership}, "invite",
       'invited user membership' );

erikjohnston: Like so?

I just cargoculted from... somewhere

tests/30rooms/12thirdpartyinvite.pl
erikjohnston: Done

tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -53,6 +53,42 @@
...
+
+      $id_server->bind_identity( undef, "email", $invitee_email, $invitee )
+      ->then( sub {
+         my $invite_info = {
+            medium    => "email",
+            address   => $invitee_email,
+            id_server => $id_server->name,
+         };
+         matrix_create_room( $inviter, invite_3pid => [ $invite_info ] )
+         ->then( sub {
```
leonerd: It would avoid the excessive indentation here, to end this `then` block and start another; by doing

       matrix_create_room( ... );
    })->then( sub {
       ( $room_id ) = @_;

tests/30rooms/12thirdpartyinvite.pl
```diff
@@ -53,6 +53,42 @@
...
+            ( $room_id ) = @_;
+
+            matrix_get_room_state( $inviter, $room_id,
+               type      => "m.room.member",
+               state_key => $invitee_mxid,
+            )->on_done( sub {
```
leonerd: And similar here with the `on_done`

tests/30rooms/12thirdpartyinvite.pl
leonerd: Yeah - that set of test functions is fairly new, so there'll probably be a lot of old code not yet using them. I might go around and fix them up sometime, but at least writing new tests should use them.

leonerd:
Aside from the indentation/reflow (which can be fixed up later) this LGTM.
Still has a red mark from testing, presumably because the related `synapse` branch isn't merged yet

PR #129, opened 2016-01-05
Neaten up code. Equivalent semantics.

erikjohnston:


leonerd:
Code LGTM
I might be tempted to alter the description a little though - it's not just a formatting change, it's an "equivalent semantics" one that neatens up the code a bit.

PR #134, opened 2016-01-06
Markjh/syn 482 room initialsync

erikjohnston:
https://matrix.org/jira/browse/SYN-482


erikjohnston:
(This PR by @negativemjark)

tests/90jira/SYN-482.pl
```diff
@@ -0,0 +1,52 @@
+multi_test "Limit on room/initialSync is reached over federation (SYN-482)",
+    requires => [qw( make_test_room local_users remote_users)],
```
leonerd: This looks like a merge of some very old code; because it's using `local_users` and `remote_users`. That hasn't been around for a month or so at least.

Take a look around instead at some of the other tests, how they use `local_user_fixture()` and `remote_user_fixture()` - you probably want to use those instead.

tests/90jira/SYN-482.pl
```diff
@@ -0,0 +1,52 @@
...
+        })->then( sub {
+            Future->needs_all(map {
+                matrix_send_room_text_message( $user_a, $room_id,
+                    body => "Message #$_",
+                )
+            } 1..3)
+        })->then( sub {
+            matrix_join_room( $user_b, $room_id )
+        })->then( sub {
+            do_request_json_for( $user_b,
```
leonerd: I believe there's a `matrix_initialsync()` helper function you can use here instead.

tests/90jira/SYN-482.pl
```diff
@@ -0,0 +1,52 @@
...
+                method => "GET",
+                uri    => "/api/v1/rooms/$room_id/initialSync",
+                params => {limit => 10},
+            )
+        })->then( sub {
+            my ($body) = @_;
+            $sync_events = $body->{messages}->{chunk};
+            do_request_json_for( $user_b,
```
leonerd: Ditto a `matrix_get_messages()`

tests/90jira/SYN-482.pl
```diff
@@ -0,0 +1,52 @@
...
+                method => "GET",
+                uri    => "/api/v1/rooms/$room_id/messages",
+                params => {limit => 10, dir => "b"},
+            )
+        })->then( sub {
+            my ($body) = @_;
+            $messages_events = ${body}->{chunk};
```
leonerd: The braces around `${body}`, while harmless, are also unnecessary here. They should be removed.

tests/90jira/SYN-482.pl
```diff
@@ -0,0 +1,52 @@
...
+        my ( $make_test_room, $local_users, $remote_users ) = @_;
+        my $user_a = $local_users->[0];
+        my $user_b = $remote_users->[0];
+        my $room_id;
+        my $messages_events;
+        my $sync_events;
+
+        $make_test_room->( [ $user_a ] )->then( sub {
+            ( $room_id ) = @_;
+            do_request_json_for( $user_a,
```
leonerd: There's a `matrix_put_room_state()` you can use here.

leonerd:
You may want to have a quick read of https://github.com/matrix-org/sytest/blob/develop/STYLE-GUIDE.rst regarding whitespace/formatting/etc.. style.

NegativeMjark:
Eh. This was mostly for the benefit of someone looking at SYN-482.
If I recall correctly I had a go at fixing SYN-482 by making room initialSync do backfill, but then @erikjohnston talked me out of it because it might cause problems if the server started backfilling without a client explicitly asking for it by hitting /messages.

NegativeMjark:
I think I've fixed the formatting. Unfortunately the test doesn't seem to fail if other federation tests have been run. @leonerd PTAL.

leonerd:
Yeees. Test (correctly) fails when run on its own.

Test unexpectedly passes when run in combination such as
    ./run-tests-in-tox tests/90jira/SYN-4*   (adds SYN-442.pl)
    ./run-tests-in-tox tests/90jira/SYN-115.pl tests/90jira/SYN-482.pl
    ./run-tests-in-tox tests/10apidoc/01register.pl tests/90jira/SYN-482.pl

leonerd:
When run with SYN-4* so the test itself passes, I get:

 * One single `/initialSync` event, being a `m.room.member` for the requesting user
 * One single `/messages` event, being the very same member event.

When run on its own I get

 * One single `/initialSync` event, being a `m.room.member` for the requesting user
 * Many `/messages` events; including the `m.room.member` but also many others after it

NegativeMjark:
We've decided to not fix SYN-482 so closing this PR

PR #135, opened 2016-01-06
Guest users must be joined to a room to see it in /sync

NegativeMjark:
Tests changes made in matrix-org/synapse#469


illicitonion:
LGTM

PR #136, opened 2016-01-11
Move 400-series /events tests to their own 'test' blocks

leonerd:
Also remove that test step from the later test.

Additionally, one existing test had a plausible-sounding title but the wrong test body, which seemed to be a copy-paste of another one. This change also fixes the behaviour to actually match the title.


illicitonion:
LGTM

PR #138, opened 2016-01-12
Check that calling /events doesn't always return events

erikjohnston:
Tests: https://github.com/matrix-org/synapse/pull/480

### DIFF
--- a/tests/30rooms/60anonymousaccess.pl
+++ b/tests/30rooms/60anonymousaccess.pl
@@ -484,7 +484,6 @@ test "Anonymous users can get individual state for world_readable rooms after le
 
 test "Annonymous user calling /events doesn't tightloop",
    requires => [ anonymous_user_fixture(), local_user_fixture() ],
-   bug => 'SYN-582',
 
    do => sub {
       my ( $anonymous_user, $user ) = @_;
--- a/tests/30rooms/60anonymousaccess.pl
+++ b/tests/30rooms/60anonymousaccess.pl
@@ -1,6 +1,5 @@
 use Future::Utils qw( try_repeat_until_success repeat );
 use JSON qw( encode_json );
-use Data::Dumper;
 
 test "Anonymous user cannot view non-world-readable rooms",
    requires => [ anonymous_user_fixture(), local_user_fixture() ],
@@ -485,6 +484,7 @@ test "Anonymous users can get individual state for world_readable rooms after le
 
 test "Annonymous user calling /events doesn't tightloop",
    requires => [ anonymous_user_fixture(), local_user_fixture() ],
+   bug => 'SYN-582',
 
    do => sub {
       my ( $anonymous_user, $user ) = @_;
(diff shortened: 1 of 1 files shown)


tests/30rooms/60anonymousaccess.pl
```diff
@@ -482,6 +483,69 @@ sub await_event_not_presence_for
...
+         );
+      })->then( sub {
+         my ( $sync_body ) = @_;
+         my $sync_from = $sync_body->{messages}->{end};
+
+         repeat( sub {
+            my ( undef, $f ) = @_;
+
+            my $body = $f ? $f->get : undef;
+            my $end_token = $body ? $body->{end} : $sync_from;
```
leonerd: You could just squash these into

    my $end_token = $f ? $f->get->{end} : $sync_from;

tests/30rooms/60anonymousaccess.pl
```diff
@@ -482,6 +483,69 @@ sub await_event_not_presence_for
...
+
+            log_if_fail "Events body", $body;
+
+            get_events_no_timeout( $anonymous_user, $room_id, $end_token );
+         }, foreach => [ 0 .. 5 ], until => sub {
+            my ( $res ) = @_;
+            $res->failure or not @{$res->get->{chunk}};
```
leonerd: Wants whitespace to draw attention to the complex expression inside the braces.

    @{ $res->get->{chunk} }

tests/30rooms/60anonymousaccess.pl
```diff
@@ -1,5 +1,6 @@
-use Future::Utils qw( try_repeat_until_success );
+use Future::Utils qw( try_repeat_until_success repeat );
 use JSON qw( encode_json );
+use Data::Dumper;
```
leonerd: Delete this extraneous `use`.

Also, be aware that `Data::Dump`'s `pp` function is imported by `run-tests.pl`, so that's free to use anywhere in a test.

leonerd:
The test does currently fail. Are you intending to fix the server implementation before merging it?
If not, you could annotate the test with a `bug` argument, informing sytest that we expect it to fail because that bug isn't fixed yet. See also the `fixes` commandline argument

erikjohnston:
Ooooooooh, how fancy.

tests/30rooms/60anonymousaccess.pl
erikjohnston: Ta

tests/30rooms/60anonymousaccess.pl
erikjohnston: Ta

tests/30rooms/60anonymousaccess.pl
erikjohnston: Ta

erikjohnston:
I think that should be everything :)

leonerd:
LGTM

PR #139, opened 2016-01-12
Test for SYN-589, that old left rooms don't get sent down new links

erikjohnston:


erikjohnston:
https://github.com/matrix-org/synapse/pull/481

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
    };
 
 
+test "Oldly left rooms don't appear in the leave section of sync",
```
leonerd: "Oldly" ??

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
...
+         matrix_join_room( $user2, $room_id_1 );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         $next = $body->{next_batch};
+
+          Future->done(1);
+      })->then( sub {
```
leonerd: The combination of the above two lines is redundant; they can be deleted.

    $next = $body->{next_batch};

    matrix_leave_room( ...

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
...
+         matrix_leave_room( $user, $room_id_1 );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         $next = $body->{next_batch};
+
+          Future->done(1);
+      })->then( sub {
```
leonerd: Ditto these

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
...
+         matrix_leave_room( $user, $room_id_1 );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id );
```
leonerd: Did you mean to provide a `since` argument here? Otherwise, you're just about to overwrite the `$next` variable, having not actually used it since it was written first on line 153.

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
...
+
+          Future->done(1);
+      })->then( sub {
+         Future->needs_all( map {
+            matrix_put_room_state( $user2, $room_id_1,
+               content  => { "filler" => $_, membership => "join" },
+               type      => "m.room.member",
+               state_key => $user2->user_id,
+            )
+         } 0 .. 20 );
```
leonerd: This is going to create 21 items of filler state - is that number really necessary? Would the test work with fewer, or even just one?

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,69 @@
...
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id, since => $next );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body->{rooms}{leave} , qw( ) );
```
leonerd: You probably don't want to use `assert_json_keys` here. What that function does is assert that it's given a JSON object, which contains *at least* the named keys. It won't mind if there's extra.

Instead you likely want something like:

    assert_json_object( my $leave = $body->{rooms}{leave} );
    keys %$leave == 0 or die "Expected no rooms in 'leave' state";

erikjohnston: *cough*

tests/31sync/09archived.pl
erikjohnston: I've lowered it and added back a comment.

tests/31sync/09archived.pl
erikjohnston: Ta

tests/31sync/09archived.pl
erikjohnston: Hush

tests/31sync/09archived.pl
erikjohnston: Ta

tests/31sync/09archived.pl
erikjohnston: Ta

erikjohnston:
I think that should be everything now

tests/31sync/09archived.pl
```diff
@@ -121,6 +121,77 @@
...
+
+         $next = $body->{next_batch};
+
+         matrix_leave_room( $user, $room_id_1 );
+      })->then( sub {
+         matrix_sync( $user, filter => $filter_id, since => $next );
+      })->then( sub {
+         my ( $body ) = @_;
+
+         $next = $body->{next_batch};
```
leonerd: Do you need to assert that you definitely did see the newly-left room here, or is that checked already in another test?

erikjohnston: Its tested elsewhere.

I'm not sure how much I want to be complicating tests testing things that are tested elsewhere. I can do either.

leonerd: If it's tested elsewhere then no need to add it here, indeed.

leonerd:
LGTM

PR #142, opened 2016-01-13
Tests that AS users must be registered before use

illicitonion:


Kegsay:
Probably LGTM? I don't know if the perl is any good. You're making some requests.

tests/60app-services/02ghost.pl
```diff
@@ -157,3 +157,35 @@
...
+   requires => [ $main::AS_USER, local_user_and_room_fixtures() ],
+
+   check => sub {
+      my ( $as_user, undef, $room_id ) = @_;
+
+      do_request_json_for( $as_user,
+         method => "POST",
+         uri    => "/api/v1/rooms/$room_id/join",
+         params => {
+            user_id => "@".$unregistered_as_user_localpart.":localhost:".$HOMESERVER_PORTS[0],
```
leonerd: You can get this a better way by requiring `$HOMESERVER_INFO[0]`, unpacking it into `$info` and then using

    $info->server_name

which will give you a string looking like `localhost:$port`

illicitonion: Done

leonerd:
LGTM

PR #143, opened 2016-01-13
Neater handling of v2 'sync' batch tokens

leonerd:
Abstract away the manual handling of sync batch tokens into the `matrix_sync` and newly-added `matrix_sync_again` helper functions. Avoids having to capture and share a lexical to store the token between multiple stages of a test.


NegativeMjark:
LGTM

PR #144, opened 2016-01-13
Bugfix presence+sync test

leonerd:
This fix removes an expected-fail condition. The previous test logic was in fact broken, and the synapse behaviour seems to be correct.

This PR actually contains two independent changes - the added helper functions, and a change to the test that would use them but doesn't yet due to a different bug. I hope to fix that one next.

### DIFF
--- a/tests/10apidoc/20presence.pl
+++ b/tests/10apidoc/20presence.pl
@@ -1,5 +1,28 @@
 my $fixture = local_user_fixture();
 
+push our @EXPORT, qw( matrix_set_presence_status );
+
+=head2 matrix_set_presence_status
+
+   matrix_set_presence_status( $user, $presence, %params )
+
+Sets the presence status of the given C<$user> to C<$presence>, with optional
+additional parameters (such as C<status_msg>) given in C<%params>.
+
+=cut
+
+sub matrix_set_presence_status
+{
+   my ( $user, $presence, %params ) = @_;
+
+   do_request_json_for( $user,
+      method => "PUT",
+      uri    => "/api/v1/presence/:user_id/status",
+
+      content => { presence => $presence, %params }
+   )->then_done();
+}
+
 test "GET /presence/:user_id/status fetches initial status",
    requires => [ $fixture ],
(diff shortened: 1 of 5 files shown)


tests/10apidoc/20presence.pl
```diff
@@ -1,15 +1,55 @@
...
+sub matrix_set_presence_status
+{
+   my ( $user, $presence, %params ) = @_;
+
+   do_request_json_for( $user,
+      method => "PUT",
+      uri    => "/api/v1/presence/:user_id/status",
+
+      content => { presence => $presence, %params }
+   )->then_done();
```
illicitonion: Why the ->then_done() ?

tests/31sync/05presence.pl
```diff
@@ -51,13 +49,25 @@
...
+         # I shouldn't be "online"
+
          my $events = $body->{presence}{events};
-         my $presence = first { $_->{type} eq "m.presence" } @$events
-            or die "Expected to see our own presence";
 
-         $presence->{sender} eq $user->user_id or die "Unexpected sender";
-         $presence->{content}{presence} eq "offline"
-            or die "Expected to be offline";
+         if( my $presence = first { $_->{type} eq "m.presence" } @$events ) {
```
illicitonion: Surely the $presence->{sender} eq $user->user_id check should be in this filter? It may possibly be valid for other users' presence to be reported.

(Maybe we would auto-populate other people on your team/company/the whole server/whatever into your presence list, in the future. Or Myspace Tom.)

tests/10apidoc/20presence.pl
leonerd: Conceptually a "set" operation shouldn't return a result, so this just throws away the returned JSON body of the HTTP response.

tests/31sync/05presence.pl
leonerd: Good point. Logic adjusted in 365e62f

illicitonion:
LGTM

PR #147, opened 2016-01-14
Use the new format for configuring listeners

NegativeMjark:
Rather than old method of supplying bind port and unsecure port


leonerd:
LGTM

PR #149, opened 2016-01-18
Add tests for ASes being able to join rooms

illicitonion:


leonerd:
LGTM

Also retested locally against latest synapse and it tests fine.

PR #150, opened 2016-01-18
Ensure that newly-registered users have some defined presence status (SYT-34)

leonerd:
Originally I wrote this test because synapse wasn't doing this and so the test failed. Before I could properly investigate it, apparently the server implementation passes now. Ohwell - I'd like to commit the test anyway then because it's still useful.

(see also https://matrix.org/jira/browse/SYT-34)


illicitonion:
LGTM

PR #151, opened 2016-01-18
Use exists test on a hash key instead of 'all' iteration

leonerd:
Rewrote test logic (I just found it because I was looking for other places where `all { ... }` is used).

Also adjusted formatting of later lines to elide the implied `->` deref arrow


leonerd:
Bah. Latest commit appears to fail, but the failure looks unrelated and one of the slightly-unreliable presence-related ones.

We should fix that...

illicitonion:
LGTM

PR #152, opened 2016-01-18
Add `assert_deeply_eq()` comparison assertion

leonerd:
A handy test for deep-nested structure, which prints the path to the differing element if one is found.


NegativeMjark:
LGTM

PR #153, opened 2016-01-20
Poll synapse's https listener to check if it has started.

NegativeMjark:
Rather than polling the log file

### DIFF
--- a/lib/SyTest/Synapse.pm
+++ b/lib/SyTest/Synapse.pm
@@ -260,7 +260,7 @@ sub start
                $polling_period *= 2;
                $polling_count -= 1;
 
-               if ($polling_count == 0) {
+               if ( $polling_count == 0 ) {
                   die "Synapse failed to start within ${polling_period}s";
                }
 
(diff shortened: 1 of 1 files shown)


lib/SyTest/Synapse.pm
```diff
@@ -239,6 +239,37 @@ sub start
...
+               my ( $connection ) = @_;
+
+               $connection->close;
+
+               $self->started_future->done;
+            }, sub {
+               $polling_period *= 2;
+               $polling_count -= 1;
+
+               if ($polling_count == 0) {
```
illicitonion: nit: spaces around ()s

illicitonion:
LGTM

lib/SyTest/Synapse.pm
```diff
@@ -239,6 +239,37 @@ sub start
             )
          );
 
+         my $polling_period = 0.01;
+         my $polling_count = 20;
+         my $poll;
+         $poll = sub {
```
leonerd: This is the closure self-capturing antipattern that `Future::Utils::repeat` is designed to help you avoid. You probably want to use that instead.

(the primary reason being it memory-leaks; though not a major disaster in this case because it's once at startup)

PR #155, opened 2016-01-21
Use a longer timeout, 60s, for test setup.

NegativeMjark:
Test setup can include starting synapse. On slower hardware, e.g my
workstation, this can take ~10s causing the first tests to timeout.

Using a separate timeout for the setup part of running a test and by
not counting the test setup time when tracking test exectuion should
prevent this from happening


leonerd:
LGTM

PR #156, opened 2016-01-21
Add tests for real users peeking (in multiple rooms)

erikjohnston:


erikjohnston:
Okay, that is one large merge conflict.

PR #159, opened 2016-01-26
Use r0 not api/v1 or v2_alpha

illicitonion:
This covers everything except /register calls, which use v1 registration


leonerd:
LGTM

Looks to be a lot of similar calls to "/events" for example here; I might pull them out into another helper function to make this an easier edit next time. :)

PR #160, opened 2016-01-27
Fix tests for adding rules relative to other rules

NegativeMjark:
The tests were incorrect because they assumed that rules were added
to the end of the list by default, rather than to the beginning of the
list.


dbkr:
lgtm

PR #162, opened 2016-01-27
Add option to run the tests against dendron rather than synapse.

NegativeMjark:


lib/SyTest/Synapse.pm
```diff
@@ -187,32 +205,66 @@ sub start
...
+      @synapse_command, "--generate-config", "--report-stats=no"
+   );
+
+   my @command;
+
+   if( $self->{dendron} ) {
+      $db_type eq "pg" or die "Dendron can only run against postgres";
+
+      my @db_arg_pairs = ();
+      while( my ( $key, $value ) = each %db_args ) {
```
leonerd: Don't use `each()` on a hash. It's silly and needs to be taken out and shot.

Instead, especially in this case, you might find `List::Util::pairmap` useful. It's like a regular `map` only takes its input list as an even-sized list of pairs, assigning values to `$a` and `$b` for each iteration:

    use List::Util qw( pairmap );

    my @db_arg_pairs = pairmap { $a eq "database" ? "dbname=$b" : "$a=$b" } %db_args;

lib/SyTest/Synapse.pm
```diff
@@ -187,32 +205,66 @@ sub start
...
+         }
+         push @db_arg_pairs, $key . '=' . $value;
+      }
+
+      @command = (
+         $self->{dendron},
+         "--synapse-python" => $self->{python},
+         "--synapse-config" => $config_path,
+         "--synapse-url" => "http://127.0.0.1:$self->{unsecure_port}",
+         "--synapse-postgres" => join(" ", @db_arg_pairs),
```
leonerd: Whitespace `join( " ", @db_arg_pairs )`

NegativeMjark: Done dbcaef9

lib/SyTest/Synapse.pm
NegativeMjark: Done dbcaef9

leonerd:
LGTM

PR #164, opened 2016-01-28
Fix some /sync tests.

erikjohnston:
See https://github.com/matrix-org/synapse/pull/534

### DIFF
--- a/tests/31sync/06state.pl
+++ b/tests/31sync/06state.pl
@@ -499,8 +499,6 @@ test "When user joins a room the state is included in the next sync",
 test "A change to displayname should not result in a full state sync",
    requires => [ local_user_fixture( with_events => 0 ),
                  qw( can_sync ) ],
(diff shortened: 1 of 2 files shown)


tests/31sync/06state.pl
```diff
@@ -637,6 +634,7 @@ sub wait_for_event_in_room {
             timeline  => { types => [] },
             state     => { types => [ "a.madeup.test.state" ] },
             ephemeral => { types => [] },
+            include_leave => JSON::true,
```
leonerd: FYI this doesn't *need* to be JSON::true; I believe a simple 1 would suffice.

leonerd:
LGTM

tests/31sync/06state.pl
erikjohnston: Ah, cool, thanks. I c+p'd from elsewhere :)

PR #166, opened 2016-01-28
Test that users can login with just the localpart of their ID

NegativeMjark:


tests/10apidoc/02login.pl
```diff
@@ -94,6 +94,38 @@
...
+test "POST /login can log in as a user with just the local part of the id",
+   requires => [ $main::API_CLIENTS[0], $registered_user_fixture,
+                 qw( can_login_password_flow )],
+
+   proves => [qw( can_login )],
+
+   do => sub {
+      my ( $http, $user_id ) = @_;
+
+      my ( $user_localpart ) = ( $user_id =~ /@([^:]*):/ );
```
leonerd: Stylistically I always put the leading `m/` on a regexp match, just because otherwise that dangling `/` can occasionally lead to parsing ambiguities.

leonerd:
LGTM besides the `m/.../` comment

PR #167, opened 2016-01-28
Adjust AS code so it could support multiple instances

leonerd:
Various code adjustments to get to the point where all AS-related behaviour comes via fixtures.

This doesn't yet give us multiple-AS ability, but that should be straightforward after this by just turning the `$AS_*` scalar variables into `@AS_` arrays.

### DIFF
--- a/lib/SyTest/ApplicationService.pm
+++ b/lib/SyTest/ApplicationService.pm
@@ -6,11 +6,12 @@ use warnings;
 sub new
 {
    my $class = shift;
-   my ( $info, $await_http ) = @_;
+   my ( $info, $await_http, $await_event ) = @_;
 
    return bless {
       info       => $info,
       await_http => $await_http,
+      await_event => $await_event,
    }, $class;
 }
 
@@ -20,6 +21,13 @@ sub await_http_request
    my ( $path, @args ) = @_;
 
    $self->{await_http}->( $self->{info}->path . $path, @args );
+}
+
+sub await_event
+{
+   my $self = shift;
+
+   $self->{await_event}->( @_ );
 }
 
 1;
--- a/lib/SyTest/ApplicationService.pm
+++ b/lib/SyTest/ApplicationService.pm
@@ -1,0 +1,25 @@
+package SyTest::ApplicationService;
+
+use strict;
+use warnings;
+
+sub new
+{
+   my $class = shift;
+   my ( $info, $await_http ) = @_;
+
+   return bless {
+      info       => $info,
+      await_http => $await_http,
+   }, $class;
+}
+
+sub await_http_request
+{
+   my $self = shift;
+   my ( $path, @args ) = @_;
+
+   $self->{await_http}->( $self->{info}->path . $path, @args );
+}
+
+1;
--- a/lib/SyTest/ApplicationService.pm
+++ b/lib/SyTest/ApplicationService.pm
@@ -5,6 +5,32 @@ use warnings;
 
 use Future::Utils qw( repeat );
 
+=head1 NAME
+
+C<SyTest::ApplicationService> - abstraction of a single application service
+
+=head1 DESCRIPTION
+
+An instance of this class represents an abstracted application service for the
+homeserver to talk to. It provides the test scripts a way to receive inbound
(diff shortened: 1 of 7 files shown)


leonerd:
Reässign to Mjark since Dan appears not to be around today.

lib/SyTest/ApplicationService.pm
```diff
@@ -0,0 +1,91 @@
...
+   my ( $info, $await_http ) = @_;
+
+   my $path = $info->path;
+
+   # Map event types to ARRAYs of Futures
+   my %futures_by_type;
+
+   my $f = repeat {
+      $await_http->( qr{^\Q$path\E/transactions/\d+$}, sub { 1 },
+         timeout => 0,
```
illicitonion: What does this timeout => 0 do?

lib/SyTest/ApplicationService.pm
```diff
@@ -0,0 +1,91 @@
+package SyTest::ApplicationService;
```
illicitonion: Some documentation on what this is would be nice

lib/SyTest/ApplicationService.pm
```diff
@@ -0,0 +1,91 @@
...
+
+         foreach my $event ( @{ $request->body_from_json->{events} } ) {
+            my $type = $event->{type};
+
+            my $queue = $futures_by_type{$type};
+
+            # Ignore any cancelled ones
+            shift @$queue while $queue and @$queue and $queue->[0]->is_cancelled;
+
+            if( $queue and my $f = shift @$queue ) {
```
illicitonion: Call this something other than $f because you're shadowing?

lib/SyTest/ApplicationService.pm
```diff
@@ -0,0 +1,91 @@
...
+      )->then( sub {
+         my ( $request ) = @_;
+
+         # Respond immediately to AS
+         $request->respond_json( {} );
+
+         foreach my $event ( @{ $request->body_from_json->{events} } ) {
+            my $type = $event->{type};
+
+            my $queue = $futures_by_type{$type};
```
illicitonion: Call this $expected_events or something rather than $queue?

tests/02as-info.pl
```diff
@@ -0,0 +1,25 @@
+push our @EXPORT, qw( AS_INFO );
+
+sub gen_token
+{
+   my ( $length ) = @_;
+   return join "", map { chr 64 + rand 63 } 1 .. $length;
+}
+
+struct ASInfo => [qw( localpart user_id as2hs_token hs2as_token path )];
```
illicitonion: Throw in an id, because ASes have IDs now (even if they're not quite mandatory yet)

illicitonion:
I'd like to take a look at the documentation you're about to write, but otherwise LGTM

leonerd:
PTAL. Now with added docs.

illicitonion:
Docs look good, four other comments are outstanding

lib/SyTest/ApplicationService.pm
leonerd: Done 0757e71

lib/SyTest/ApplicationService.pm
leonerd: Not *strictly* a shadowing, because the outer `$f` isn't in scope until after the statement that introduced it, which is after the entire `repeat` call. But I agree it's still confusing to the human reader, so I've renamed it. dd15368

tests/02as-info.pl
leonerd: Done ac9a00f

lib/SyTest/ApplicationService.pm
leonerd: The same as the clearer `undef`; now changed to that plus added comment. 65e7b66

lib/SyTest/ApplicationService.pm
leonerd: Done bb12f30

leonerd:
PTAL

illicitonion:
Make the ID non-null and I'm happy

PR #171, opened 2016-02-02
Fix up sync tests

erikjohnston:


erikjohnston:
matrixbot: test this please

tests/31sync/09archived.pl
```diff
@@ -219,7 +222,7 @@
 
       my $filter = {
          room => {
-            timeline => { types => [ "m.room.message" ] },
+            timeline => { types => [ "m.room.message", "a.madeup.test.state" ] },
```
leonerd: This might have been neater using `qw( ... )` but then so many of the other tests aren't using that, so... hmmm

leonerd:
LGTM

PR #173, opened 2016-02-03
Use $output->diag() to avoid messing up progressive terminal output

leonerd:
Raw prints to `STDERR` tend to break it.

### DIFF
--- a/lib/SyTest/Identity/Server.pm
+++ b/lib/SyTest/Identity/Server.pm
@@ -125,7 +125,7 @@ sub on_request
       $req->respond_json( \%resp );
    }
    else {
-      print STDERR "Unexpected request to Identity Service for $path";
+      warn "Unexpected request to Identity Service for $path";
       $req->respond( HTTP::Response->new( 404, "Not Found", [ Content_Length => 0 ] ) );
    }
 }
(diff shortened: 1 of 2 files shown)


NegativeMjark:
You should probably fix the other places in the file that talk to STDERR otherwise I will cargocult them again.

NegativeMjark:
$ grep -rn STDERR lib/ tests
    lib/SyTest/Identity/Server.pm:128:      print STDERR "Unexpected request to Identity Service for $path";
    lib/SyTest/Synapse.pm:267:            print STDERR $stderr;
    lib/SyTest/Synapse.pm:337:      print STDERR "Process failed ($exitcode)\n";
    lib/SyTest/Synapse.pm:339:      print STDERR "\e[1;35m[server $self->{port}]\e[m: $_\n"
    lib/SyTest/Synapse.pm:383:            print STDERR "\e[1;35m[server $self->{port}]\e[m: $line\n";
    lib/SyTest/Synapse.pm:412:      print STDERR "\e[1;35m[server $self->{port}]\e[m: $_\n"

FWIW

leonerd:
Some of those fixed. The remaining probably want a bit of extra support in the $output instances, which would put it past the realm of "tiny-fixes" for this branch. ;) I'll neaten those up in a bit

NegativeMjark:
LGTM

PR #177, opened 2016-02-08
Always PUT not POST when sending

illicitonion:


tests/10apidoc/34room-messages.pl
```diff
@@ -21,6 +21,8 @@
       });
    };
 
+our $global_txn_id = 0;
```
leonerd: This isn't being shared with any other files, so it can just be `my`

illicitonion: Done

leonerd:
LGTM

PR #184, opened 2016-02-15
The membership property moved to content long ago

illicitonion:


tests/60app-services/03passive.pl
```diff
@@ -78,17 +78,19 @@
...
                die "Expected room_id to be $room_id";
             $event->{user_id} eq $local_user->user_id or
                die "Expected user_id to be ${\ $local_user->user_id }";
-            $event->{membership} eq "join" or
-               die "Expected membership to be 'join'";
             $event->{state_key} eq $local_user->user_id or
                die "Expected state_key to be ${\ $local_user->user_id }";
 
+            assert_json_keys( $event->{content}, qw( membership ));
+            $event->{content}->{membership} eq "join" or
```
leonerd: Extraneous arrow; after the first one it's not required. Also we want to be using `assert_eq()` these days as it gives nicer messages. So this wants to be

    assert_eq( $event->{content}{membership}, "join",
       'event content membership' );

leonerd:
LGTM besides one comment

tests/60app-services/03passive.pl
illicitonion: Done (also tidied up the above ones)

PR #185, opened 2016-02-15
Check that the correct predefined rules are returned

NegativeMjark:
Checks that the push rules match what's in matrix-org/matrix-doc#274


dbkr:
What's the current standing on sytest testing synapse-specific behaviour?

leonerd:
@dbkr there's already a number of synapse-specific tests; but generally they're regression tests for specific bugs, having specific SYN-nnn bug numbers, so they live in 90jira/...

NegativeMjark:
I'd say that the default push rules probably need to be part of the specification for them to be useful to clients, I can see a couple of options:

1) We make the default push rules be part of the spec.
2) We mark the test as being synapse specific.
3) Rather than checking that all the ".m.*" rules exist and match, we just check that all of the ".m.*" rules returned match their definition in the spec.

dbkr:
OK, what do we want to do with this? I'm not really happy with it in its current state, or at least if we're mandating this exact behaviour, I'd expect to see this reflected very explicitly in the sister doc change (ie. "A new user account must start with exactly this set of rules", "An attempt to add a rule before a server defined rule gives X matrix error". I'm also not sure I see the value in specifying the latter, since server can safely allow this if they wish.

NegativeMjark:
Hmm, it doesn't look like I'm going to get round to merging this anytime soon.

PR #190, opened 2016-02-19
Check that the homeserver rejects receipts with the wrong type

NegativeMjark:


tests/30rooms/21receipts.pl
```diff
@@ -157,3 +157,25 @@ sub find_receipt
          })
       });
    };
+
+
```
leonerd: No need for two blank lines here

leonerd:
LGTM besides extraneous blank line

PR #191, opened 2016-02-24
Check that getting the sub-paths for push rules works

NegativeMjark:


illicitonion:
Looks no worse than what exists to me.

PR #193, opened 2016-02-24
Check that push rules API gives correct 4xx responses for invalid input

NegativeMjark:


tests/61push/80torture.pl
```diff
@@ -0,0 +1,89 @@
+my $user_fixture = local_user_fixture();
```
illicitonion: Why share across tests?

illicitonion:
*shrug* LGTM I guess (with one comment)

PR #194, opened 2016-02-25
Check getting the enabled attribute of a push rule

NegativeMjark:


dbkr:
seems reasonable. lgtm.

PR #197, opened 2016-03-01
Check that aliases can be created and deleted

erikjohnston:


tests/30rooms/05aliases.pl
```diff
@@ -101,3 +101,53 @@
...
+test "Creators can delete alias",
+   requires => [ $creator_fixture, $room_fixture,
+                 qw( can_create_room_alias )],
+
+   bug => "SYN-83",
+
+   do => sub {
+      my ( $user, $room_id ) = @_;
+      my $server_name = $user->http->server_name;
+      $room_alias = "#Kuke9aic:$server_name";
```
leonerd: Rather than this hardcoded value in the test, a more robust way would be to use a `room_alias_fixture()`; see 10apidoc/30room-create.pl

tests/30rooms/05aliases.pl
```diff
@@ -101,3 +101,53 @@
...
+   };
+
+test "Users can't delete other's aliases",
+   requires => [ $creator_fixture, $room_fixture, local_user_fixture(),
+                 qw( can_create_room_alias )],
+
+   do => sub {
+      my ( $user, $room_id, $other_user ) = @_;
+      my $server_name = $user->http->server_name;
+      $room_alias = "#ta8Shu3z:$server_name";
```
leonerd: ditto

erikjohnston:
@leonerd PTAL

leonerd:
LGTM

PR #199, opened 2016-03-02
Annotate log_if_fail lines with elapsed timestamps

richvdh:
For context: I was trying to understand what was taking so long in a test which was timing out.


richvdh:
ok to test

leonerd:
LGTM

PR #200, opened 2016-03-04
Make it possible to configure the logging for synapse

richvdh:


lib/SyTest/Synapse.pm
```diff
@@ -115,7 +115,8 @@ sub start
    }
 
    if( defined $db_type ) {
-      $self->${\"clear_db_$db_type"}( %db_args );
+      my $clear_meth = "clear_db_${db_type}";
+      $self->${\$clear_meth}( %db_args );
```
leonerd: If you're going to pull it out to a separate variable then it can just be

    $self->$clear_meth( %db_args );

lib/SyTest/Synapse.pm
```diff
@@ -160,12 +161,17 @@ sub start
...
+   if( ! -f $log_config_file) {
+      $log_config_file = undef;
+   }
 
    my $macaroon_secret_key = "secret_$self->{port}";
 
    my $config_path = $self->write_yaml_file( config => {
         "server_name" => "localhost:$port",
         "log_file" => "$log",
+        "log_config" => $log_config_file,
```
leonerd: This does mean you'll end up setting the `log_config` key anyway, even if it just gets a YAML undef value. It might be nicer to do

    ( -f $log_config_file ) ? ( log_config => $log_config_file ) : (),

and then also not bother with the `-f` test earlier

lib/SyTest/Synapse.pm
richvdh: Thanks. TBH the main reason I pulled it out was because my editor was confused by `\"`.

lib/SyTest/Synapse.pm
richvdh: done

richvdh:
@leonerd ptal

leonerd:
Code LGTM. Waiting on jenkins but if that's happy then merge it.

PR #201, opened 2016-03-04
Suppress undef warnings when using local postgres

richvdh:
It's valid to configure a postgres database without a host (it means we will
connect over a UNIX socket), so don't spam warnings when that happens.

### DIFF
--- a/lib/SyTest/Synapse.pm
+++ b/lib/SyTest/Synapse.pm
@@ -433,7 +433,7 @@ sub clear_db_pg
    my $self = shift;
    my %args = @_;
 
-   my $host = $args{host} || '';
+   my $host = $args{host} // '';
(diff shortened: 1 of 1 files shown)


lib/SyTest/Synapse.pm
```diff
@@ -433,12 +433,13 @@ sub clear_db_pg
    my $self = shift;
    my %args = @_;
 
-   $self->{output}->diag( "Clearing Pg database $args{database} on $args{host}" );
+   my $host = $args{host} || '';
```
leonerd: Use the `//` operator as we care about definedness, not truth (just in case for some bizarre reason we want to use "0" as a host)

richvdh: done

leonerd:
LGTM

PR #203, opened 2016-03-08
Update comment to match actual code behaviour

leonerd:


erikjohnston:
ah, yes, LGTM, thanks

erikjohnston:
@matrixbot retest this please

leonerd:
Merging even though Jenkins complains, because the failing test is elsewhere and fails for all branches, and also this is only a comment change anyway.

PR #207, opened 2016-03-11
Implement logout tests

erikjohnston:


tests/10apidoc/02login.pl
```diff
@@ -206,3 +206,29 @@
          }
       );
    };
+
+
```
leonerd: Excessive blank lines - 1 is fine

tests/10apidoc/02login.pl
```diff
@@ -206,3 +206,29 @@
...
+our @EXPORT = qw( matrix_login_again_with_user );
+
+
+sub matrix_login_again_with_user {
```
leonerd: Opening `{` of a named function should be on its own line in column 1

tests/10apidoc/02login.pl
```diff
@@ -206,3 +206,29 @@
...
+         type     => "m.login.password",
+         user     => $user->user_id,
+         password => $user->password,
+      },
+   )->then( sub {
+      my ( $body ) = @_;
+
+      assert_json_keys( $body, qw( access_token home_server refresh_token ));
+
+      my $new_user = User( $user->http, $user->user_id, $user->password, $body->{access_token}, $body->{refresh_token}, undef, undef, [], undef );
```
leonerd: Rather than allocating a new `User` object, you can note that the existing one is mutable. So just mutate its fields:

    $user->access_token = $body->{access_token};
    $user->refresh_token = $body->{refresh_token};

tests/10apidoc/02login.pl
```diff
@@ -206,3 +206,29 @@
...
+
+      Future->done( $new_user );
```
leonerd: Then don't bother returning the user

tests/13logout.pl
```diff
@@ -0,0 +1,48 @@
...
+         ( $other_login ) = @_;
+
+         do_request_json_for( $user,
+            method => "POST",
+            uri    => "/r0/logout",
+            content => {},
+         )
+      })->then( sub {
+         matrix_sync( $user );
+      })->main::expect_http_401->then( sub {
```
leonerd: The `expect_http_401` should be attached to the sync request directly, so it only checks the result of that. The way it's written here it would catch (and consider as "OK") a 401 failure from any of the preceding logic; e..g the login or logout requests. This wants to be

    ->then( sub {
       matrix_sync( $user )->main::expect_http_401
    })->then( sub { ...

tests/13logout.pl
```diff
@@ -0,0 +1,48 @@
...
+         ( $other_login ) = @_;
+
+         do_request_json_for( $user,
+            method => "POST",
+            uri    => "/r0/logout/all",
+            content => {},
+         )
+      })->then( sub {
+         matrix_sync( $user );
+      })->main::expect_http_401->then( sub {
```
leonerd: Same here x2

tests/10apidoc/02login.pl
leonerd: Ah actually now I read the later tests, I see why it's making a new one.

erikjohnston:
@leonerd PTAL

leonerd:
LGTM

PR #209, opened 2016-03-11
Check that already present members cannot be invited to a room

leonerd:


tests/30rooms/06invite.pl
```diff
@@ -271,3 +271,25 @@ sub invited_user_can_reject_invite_for_empty_room
...
+   requires => [ local_user_and_room_fixtures() ],
+
+   do => sub {
+      my ( $creator, $room_id ) = @_;
+
+      matrix_invite_user_to_room( $creator, $creator, $room_id )
+         ->main::expect_http_403;
+   };
+
+test "Users cannot invite others to a room they're already in",
```
NegativeMjark: "Users cannot invite a user that is already in the room." might be clearer.

NegativeMjark:
Other than wording for the test name it LGTM.

PR #211, opened 2016-03-14
Set new config option invite_3pid_guest

erikjohnston:
See https://github.com/matrix-org/synapse/pull/645 for context


NegativeMjark:
LGTM

PR #212, opened 2016-03-15
Invite rejections: test that invitee no longer gets the invite down /sync

erikjohnston:
Tests: https://github.com/matrix-org/synapse/pull/646

### DIFF
--- a/tests/30rooms/06invite.pl
+++ b/tests/30rooms/06invite.pl
@@ -142,6 +142,14 @@ sub invited_user_can_reject_invite
          die "Expected membership to be 'leave'";
 
       Future->done(1);
+   })->then( sub {
+      matrix_sync( $invitee )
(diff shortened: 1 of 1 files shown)


tests/30rooms/06invite.pl
```diff
@@ -142,6 +142,14 @@ sub invited_user_can_reject_invite
          die "Expected membership to be 'leave'";
 
       Future->done(1);
+   })->then( sub {
+      matrix_sync( $invitee )
+   })->then( sub {
+      my ( $body ) = @_;
+
```
NegativeMjark: I realise that this would be somewhat unprecedented in this code base, but perhaps a short comment to explain why the dict should be empty might be useful here.

NegativeMjark:
LGTM

PR #219, opened 2016-03-23
Don't assume public rooms are fine

erikjohnston:
... and by fine i mean have an aliases key


NegativeMjark:
Why would rooms in the public room list omit the "aliases" key rather than having an empty list?

erikjohnston:
> Why would rooms in the public room list omit the "aliases" key rather than having an empty list?

I think either is reasonable, given the other keys are optional.

NegativeMjark:
LGTM

PR #221, opened 2016-04-05
Add reject invite over federation test.

erikjohnston:


NegativeMjark:
LGTM from a matrix perspective, might be worth prodding @leonerd to check that the perl makes sense.

lib/SyTest/HTTPServer/Request.pm
```diff
@@ -48,6 +48,19 @@ sub respond_json
    $self->respond( $response );
 }
 
+sub respond_error_json
```
leonerd: Rather than add a new method, I'd suggest a named option on the existing `respond_json`. Change it to

    my ( $json, %opts ) = @_;

    my $response = HTTP::Response->new( $opts{code} // 200 );

Then elsewhere you can just

    ->respond_json( {}, code => 404 );

erikjohnston: That's much neater, yes.

Done.

leonerd:
LGTM

PR #222, opened 2016-04-06
Test to check we can see our own leave events

richvdh:
(test for https://github.com/matrix-org/synapse/pull/699)

### DIFF
--- a/tests/31sync/09archived.pl
+++ b/tests/31sync/09archived.pl
@@ -47,7 +47,6 @@ test "Newly left rooms appear in the leave section of incremental sync",
          matrix_create_room( $user );
       })->then( sub {
          ( $room_id ) = @_;
-
          matrix_sync( $user, filter => $filter_id );
       })->then( sub {
          matrix_leave_room( $user, $room_id );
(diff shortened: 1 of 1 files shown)


tests/31sync/09archived.pl
```diff
@@ -56,16 +55,67 @@
...
+         my $ev = $room->{timeline}{events}[0];
+         assert_eq($ev->{type}, "m.room.member", "event type");
+         assert_eq($ev->{content}{membership}, "leave", "membership");
+
          Future->done(1);
       });
    };
 
+test "We should see our own leave event, even if history_visibility is " .
+    "restricted (SYN-662)",
```
leonerd: BTW, don't feel you have to stick to a strict 80-column limit just for strings like this; they're fine to spill over as there's nothing else semantically "important" out there.

tests/31sync/09archived.pl
```diff
@@ -56,16 +55,67 @@
...
          @{ $room->{state}{events} } == 0
             or die "Expected no state events";
 
+         # we should see our own leave event
+         assert_eq(scalar @{ $room->{timeline}{events} }, 1,
+                   "timeline events");
+         my $ev = $room->{timeline}{events}[0];
+         assert_eq($ev->{type}, "m.room.member", "event type");
+         assert_eq($ev->{content}{membership}, "leave", "membership");
+
```
leonerd: Wants whitespace inside argument parens - `func( p1, p2, ... )`

tests/31sync/09archived.pl
```diff
@@ -56,16 +55,67 @@
...
+         log_if_fail "sync result", $body;
+
+         my $room = $body->{rooms}{leave}{$room_id};
+         assert_json_keys( $room, qw( timeline state ));
+
+         @{ $room->{state}{events} } == 0
+            or die "Expected no state events";
+
+         assert_eq(scalar @{ $room->{timeline}{events} }, 1,
+                   "timeline events");
```
leonerd: Could you pick just one style of testing these; rather than using two different ones for different fields in the same test?

tests/31sync/09archived.pl
richvdh: they make the file ugly in my editor :)

tests/31sync/09archived.pl
richvdh: mmyes, c&p for the loss.

tests/31sync/09archived.pl
richvdh: argh. This is the problem with having completely different coding styles to every other project. Now fixed.

richvdh:
ptal?

leonerd:
LGTM

PR #224, opened 2016-04-07
Test unread counts in pushes

dbkr:
Tests that unread counts increase and decrease correctly once messages and read receipts are sent respectively. Tests existing behaviour but uses event_id field added in https://github.com/matrix-org/synapse/pull/705/files

### DIFF
--- a/tests/61push/01message-pushed.pl
+++ b/tests/61push/01message-pushed.pl
@@ -47,6 +47,8 @@ multi_test "Test that a message is pushed",
          matrix_join_room( $bob, $room_id )
       })->then( sub {
          await_event_for( $alice, filter => sub {
+            my ( $event ) = @_;
+            return unless $event->{type} eq "m.room.member";
             return 1;
          })->then( sub {
             my ( $event ) = @_;
--- a/tests/61push/01message-pushed.pl
+++ b/tests/61push/01message-pushed.pl
@@ -93,7 +93,7 @@ multi_test "Test that a message is pushed",
             })->then( sub {
                my ( $request ) = @_;
 
-               $request->respond( HTTP::Response->new( 200, "OK", [], "" ) );
+               $request->respond_json( {} );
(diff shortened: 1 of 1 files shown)


tests/61push/01message-pushed.pl
```diff
@@ -40,11 +40,21 @@
...
 
             matrix_invite_user_to_room( $alice, $bob, $room_id ),
+            flush_events_for( $alice ),
          )
       })->then( sub {
          # Bob accepts the invite by joining the room
          matrix_join_room( $bob, $room_id )
       })->then( sub {
+         await_event_for( $alice, filter => sub {
+            return 1;
```
leonerd: You probably don't want to wait for any possible event, because that leads to fragile tests that get upset about things like spurious `m.presence` updates. Consider looking for just `m.room.receipt` events maybe?

tests/61push/01message-pushed.pl
```diff
@@ -112,6 +123,40 @@
...
+            await_http_request( "/alice_push", sub {
+               my ( $request ) = @_;
+               my $body = $request->body_from_json;
+
+               return unless $body->{notification}{counts};
+               return 1;
+            })->then( sub {
+               my ( $request ) = @_;
+
+               $request->respond( HTTP::Response->new( 200, "OK", [], "" ) );
```
leonerd: Does this need to be an actually-empty body? You didn't supply a `Content-Length` or connection close or chunked encoding, so bad things will happen. Probably easiest if you don't need the actually-empty response is to

    $request->respond_json( {} );

to have it fill in the content length at least.

dbkr:
ptal! :)

leonerd:
LGTM. Though Jenkins appears unhappy.

dbkr:
yeah, it needs https://github.com/matrix-org/synapse/pull/705 so will wait until that's merged

PR #232, opened 2016-04-25
Increase the timeouts on a couple of problem tests

NegativeMjark:
These tests keep randomly failing due to timeouts on jenkins. It might be an idea to bump the timeouts for now so that we get more reliable test output.


leonerd:
LGTM

PR #234, opened 2016-05-05
Add test for a openidish mechanism for proving to third parties that …

NegativeMjark:
…you own a given user_id

Tests the APIs added to synapse in matrix-org/synapse#765


erikjohnston:
Could do with a test to makes sure the federation thing rejects incorrect tokens, but otherwise LGTM

PR #238, opened 2016-06-03
Wait for /sync to be ready in the tests.

NegativeMjark:
We want to do this because we want to test /sync requests that are handled by a separate process.
How it works:
  * Before performing an update sytest makes a /sync request to get a token from synapse. This /sync request doesn't affect the token stored on the user object.
  * Sytest then performs an action that updates.
  * Sytest then polls /sync until the change it was expecting is visible.
  * When the change is visible it returns result of the update.
  * Sytest then can call sync for real and check that the result matches it's expectations.

Alternatives:
  1) Have the tests call /sync repeatedly until they observe whatever the test is expecting. This doesn't work because some of the tests need to check that the timeline is limited appropriately between /sync invocations.
  2) Do the action, then repeatedly call /sync. This doesn't work because the action could be returned to the client before the first /sync that was checking for it.

### DIFF
--- a/tests/31sync/02sync.pl
+++ b/tests/31sync/02sync.pl
@@ -2,14 +2,14 @@ use Future::Utils qw( repeat );
 
 push our @EXPORT, qw(
    matrix_do_and_wait_for_sync
-   matrix_send_room_text_message_and_wait_for_sync
-   matrix_send_room_message_and_wait_for_sync
-   matrix_create_room_and_wait_for_sync
-   matrix_join_room_and_wait_for_sync
-   matrix_leave_room_and_wait_for_sync
-   matrix_invite_user_to_room_and_wait_for_sync
-   matrix_put_room_state_and_wait_for_sync
-   matrix_advance_room_receipt_and_wait_for_sync
+   matrix_send_room_text_message_synced
+   matrix_send_room_message_synced
+   matrix_create_room_synced
+   matrix_join_room_synced
+   matrix_leave_room_synced
+   matrix_invite_user_to_room_synced
+   matrix_put_room_state_synced
+   matrix_advance_room_receipt_synced
    sync_timeline_contains
 );
 
@@ -104,7 +104,7 @@ sub sync_timeline_contains
    sync_room_contains( $sync_body, $room_id, "timeline", $check );
 }
 
-sub matrix_send_room_text_message_and_wait_for_sync
+sub matrix_send_room_text_message_synced
 {
    my ( $user, $room_id, %params ) = @_;
 
@@ -122,7 +122,7 @@ sub matrix_send_room_text_message_and_wait_for_sync
    );
 }
 
-sub matrix_send_room_message_and_wait_for_sync
+sub matrix_send_room_message_synced
 {
    my ( $user, $room_id, %params ) = @_;
 
@@ -140,7 +140,7 @@ sub matrix_send_room_message_and_wait_for_sync
    );
 }
 
-sub matrix_create_room_and_wait_for_sync
+sub matrix_create_room_synced
 {
    my ( $user, %params ) = @_;
 
@@ -152,7 +152,7 @@ sub matrix_create_room_and_wait_for_sync
    );
 }
 
-sub matrix_join_room_and_wait_for_sync
+sub matrix_join_room_synced
 {
    my ( $user, $room_id, %params ) = @_;
 
@@ -164,7 +164,7 @@ sub matrix_join_room_and_wait_for_sync
    );
 }
 
-sub matrix_leave_room_and_wait_for_sync
+sub matrix_leave_room_synced
 {
    my ( $user, $room_id, %params ) = @_;
 
@@ -176,7 +176,7 @@ sub matrix_leave_room_and_wait_for_sync
    );
 }
 
-sub matrix_put_room_state_and_wait_for_sync
+sub matrix_put_room_state_synced
 {
    my ( $user, $room_id, %params ) = @_;
 
@@ -195,7 +195,7 @@ sub matrix_put_room_state_and_wait_for_sync
    );
 }
 
-sub matrix_invite_user_to_room_and_wait_for_sync
+sub matrix_invite_user_to_room_synced
 {
    my ( $inviter, $invitee, $room_id, %params ) = @_;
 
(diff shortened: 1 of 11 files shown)


tests/31sync/02sync.pl
```diff
@@ -1,3 +1,222 @@
...
+
+   my $room =  $sync_body->{rooms}{join}{$room_id};
+
+   foreach my $event ( @{ $room->{$section}{events} } ) {
+      if ( $check->( $event ) ) {
+         return 1;
+      }
+   }
+
+   return 0;
```
leonerd: This whole chunk could just be

    return any { $check->( $_ ) } @{ $room->{$section}{events} };

tests/31sync/02sync.pl
```diff
@@ -1,3 +1,222 @@
...
+{
+   my ( $user, %params ) = @_;
+
+   my $check = delete $params{check} or die "Must supply a 'check' param";
+   my $do = delete $params{do} or die "Must supply a 'do' param";
+   $params{timeout} = $params{timeout} // 1000;
+
+   my $next_batch;
+
+   matrix_sync( $user,
```
leonerd: Could do with a small amount of documentation to explain what the `do` and `check` functions do; especially with the two arguments to `check`.

tests/31sync/02sync.pl
```diff
@@ -1,3 +1,222 @@
...
+      },
+      check => sub {
+         sync_timeline_contains( $_[0], $room_id, sub {
+            $_[0]->{type} eq "m.room.member"
+               and $_[0]->{state_key} eq $invitee->user_id
+               and $_[0]->{content}{membership} eq "invite"
+         });
+      },
+   );
+}
```
leonerd: Rather than two nested calls, could you use `->then` chaining instead?

tests/31sync/03joined.pl
```diff
@@ -252,7 +254,13 @@
 
          matrix_sync( $user_a );
       })->then( sub {
-         matrix_join_room( $user_b, $room_id );
+         matrix_send_room_text_message_and_wait_for_sync( $user_a, $room_id,
+            body => "Wait for presence changes cause by the first sync to trickle through",
```
leonerd: s/cause/&d/

leonerd:
Overall the direction seems OK, but I worry about the lengths of some of these function names. How about using `synced` instead; one of the following alternative patterns?

    matrix_join_room_synced
    matrix_synced_join_room

tests/31sync/06state.pl
```diff
@@ -258,7 +260,12 @@ sub wait_for_event_in_room {
                content => { "filler" => $_ },
                type    => "a.made.up.filler.type",
             )
-         } 0 .. 20 );
+         } 0 .. 19 );
+      })->then( sub {
+         matrix_send_room_message_and_wait_for_sync( $user, $room_id,
```
leonerd: Why this addition?

tests/31sync/06state.pl
```diff
@@ -542,6 +560,10 @@ sub wait_for_event_in_room {
             state_key => $user->user_id,
          );
       })->then( sub {
+         matrix_send_room_text_message_and_wait_for_sync( $user, $room_id,
```
leonerd: If tracking down this bug is going to hold up progress too long, could you at least put a `TODO(mjark)` marker in somewhere, so we can hunt that down later?

tests/31sync/09archived.pl
```diff
@@ -129,9 +133,10 @@
       matrix_create_filter( $user, {} )->then( sub {
          ( $filter_id ) = @_;
 
-         Future->needs_all(
-            matrix_create_room( $user )->on_done( sub { ( $room_id_1 ) = @_; } ),
-            matrix_create_room( $user )->on_done( sub { ( $room_id_2 ) = @_; } ),
+         matrix_create_room( $user )->on_done( sub { ( $room_id_1 ) = @_; } );
+      })->then( sub {
+         matrix_create_room_and_wait_for_sync($user )->on_done( sub {
+            ( $room_id_2 ) = @_; }
```
leonerd: Any reason these two have to now be serialised, instead of the concurrent form they were in before?

tests/31sync/09archived.pl
```diff
@@ -129,9 +133,10 @@
       matrix_create_filter( $user, {} )->then( sub {
          ( $filter_id ) = @_;
 
-         Future->needs_all(
-            matrix_create_room( $user )->on_done( sub { ( $room_id_1 ) = @_; } ),
-            matrix_create_room( $user )->on_done( sub { ( $room_id_2 ) = @_; } ),
+         matrix_create_room( $user )->on_done( sub { ( $room_id_1 ) = @_; } );
+      })->then( sub {
+         matrix_create_room_and_wait_for_sync($user )->on_done( sub {
```
leonerd: Missing whitespace `( $user`

tests/31sync/02sync.pl
NegativeMjark: Not really, otherwise we lose the guarantee that the first sync happens before the action. This would complicate the check.

tests/31sync/06state.pl
NegativeMjark: We need to have a ``wait_for_sync`` so that before we call ``matrix_sync_again`` below. However invoking ``matrix_send_room_message_and_wait_for_sync`` in parallel is dangerous because we might skip over messages in a gapped sync causing ``wait_for_sync`` to wait forever.

tests/31sync/06state.pl
NegativeMjark: I've added another sytest and filed a synapse bug for now.

NegativeMjark:
@leonerd I've renamed the methods from ``and_wait_for_sync`` to ``synced``. I think I've responded to your other comments. PTAL

tests/31sync/02sync.pl
```diff
@@ -1,3 +1,246 @@
...
+      },
+   )->get;
+
+
+Does something and waits for the result to appear in an incremental sync.
+Doesn't affect the next_batch token used by matrix_sync_again.
+
+The do parameter is the a subroutine with the action to perform that returns
+a future.
+The check parameter is a subroutine that receives the body of an incremental
```
leonerd: "is the a"??

Also you might like to use `C<do>` etc.. to notate the names of params in fixedwidth font.

tests/31sync/02sync.pl
leonerd: (explained IRL; this seems the best we can do).

leonerd:
LGTM

PR #239, opened 2016-06-03
Optionally run tests against a dendron with a split out synchrotron

NegativeMjark:
Checks that splitting out the /sync to a separate process works.

### DIFF
--- a/lib/SyTest/Synapse.pm
+++ b/lib/SyTest/Synapse.pm
@@ -305,14 +305,13 @@ sub start
       );
 
       if ( $self->{pusher} ) {
-         @command = ( @command, "--pusher-config" => $pusher_config_path );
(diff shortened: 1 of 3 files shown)


lib/SyTest/Synapse.pm
```diff
@@ -278,6 +307,13 @@ sub start
       if ( $self->{pusher} ) {
          @command = ( @command, "--pusher-config" => $pusher_config_path );
       }
+
+      if ( $self->{synchrotron} ) {
+         @command = ( @command,
```
leonerd: I'd be tempted by

    push @command, "--synchrotron-config", ...

and also fix up the other cases above it.

NegativeMjark:
@leonerd I've used ``push`` in the places you asked me to. PTAL

leonerd:
LGTM

PR #240, opened 2016-06-08
Globally disable any and all HTTP/1.1 pipelining for all HTTP clients…

leonerd:
…, because it's almost never what we want in the presence of longpolling


NegativeMjark:
LGTM

PR #242, opened 2016-06-13
Only wait for the last filler message, otherwise it times out on jenkins

NegativeMjark:


leonerd:
L{OK}TM

It'd be nice to pull out this filler logic into another function but.. eh.. This particular commit doesn't overly make the problem worse.

PR #243, opened 2016-06-13
Add test for createUser endpoint

negzi:
Signed-off-by: Negar Fazeli negar.fazeli@ericsson.com

### DIFF
--- a/tests/10apidoc/01register.pl
+++ b/tests/10apidoc/01register.pl
@@ -61,6 +61,30 @@ test "POST /register can create a user",
       });
    };
 
+test "POST /createUser can get or create a user",
+   requires => [ $main::API_CLIENTS[0] ],
+
+   do => sub {
+      my ( $http ) = @_;
+
+      $http->do_request_json(
+         method => "POST",
+         uri    => "/unstable/createUser",
+
+         content => {
+            localpart     => "01register-user"
+            displayname     => "01register_user x",
+            duration_seconds => 200,
+         },
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( user_id access_token ));
+
+         Future->done( 1 );
+      });
+   };
+
 push our @EXPORT, qw( localpart_fixture );
 
 my $next_anon_uid = 1;


leonerd:
Do you have any documentation on what the `createRoom` API is supposed to do? I don't see it on the official spec doc

negzi:
the createUser implementation feature is on development branch still. Therefore there are no API docs about it yet. 
Basically what it does is generating a new  access token for a user if it exists otherwise it registers the user and send an access token.

negzi:
Erik knows about this also, I will poke him on vector.

tests/10apidoc/01register.pl
```diff
@@ -61,6 +61,32 @@
...
+   requires => [ $main::API_CLIENTS[0] ],
+
+   critical => 1,
+
+   do => sub {
+      my ( $http ) = @_;
+
+      $http->do_request_json(
+         method => "POST",
+         uri    => "/api/v1/createUser",
```
erikjohnston: This should probably be using the `/unstable/createUser` url, or whatever it is.

tests/10apidoc/01register.pl
```diff
@@ -61,6 +61,32 @@
       });
    };
 
+test "POST /createUser can get or create a user",
+   requires => [ $main::API_CLIENTS[0] ],
+
+   critical => 1,
```
erikjohnston: I have literally no idea what this flag means, but this test probably isn't. @leonerd ?

erikjohnston:
@leonerd This is a completely out of spec addition to synapse, but is nonetheless something that I would like to test if people are relying on. Not sure where it should go though, if its vaguely synapse specific

tests/10apidoc/01register.pl
leonerd: Yes it doesn't sound like this is a critical test.

A `critical` test is one that is deemed so important to the entire remaining test suite, that if it fails we should just bail out and stop here. That's marked on things like the regular user registration test, because if sytest can't register users at all, there's very little point continuing at all.

leonerd:
@erikjohnston Eh.. right now there's probably quite a lot of fairly synapse-specific testing going on. At this point I don't have a good answer to how to split it out, so for now just leaving it here is ... well, not OK, but no worse than all the other files anyway.

tests/10apidoc/01register.pl
erikjohnston: Ok, then @negzi you can probably just remove this line entirely

erikjohnston:
@matrixbot  ok to test? (Will that work?)

tests/10apidoc/01register.pl
negzi: Do I need to change this or can it stay here?

erikjohnston: Please :)

tests/10apidoc/01register.pl
negzi: Sure, and thanks for the comments.

negzi:
Please ignore my last change. I did a whitespace cleanup in emacs and everything got messed up. I will fix this and push

negzi:
OK so now it should be fine :)

tests/10apidoc/01register.pl
```diff
@@ -61,6 +61,30 @@
...
+      my ( $http ) = @_;
+
+      $http->do_request_json(
+         method => "POST",
+         uri    => "/unstable/createUser",
+
+         content => {
+            localpart     => "01register-user"
+            displayname     => "01register_user x",
+            duration_seconds => 200,
```
leonerd: These arrows don't line up, nor do they have consistent spacing. Please insert whitespaces before the `=>` to line them up.

leonerd:
LGTM aside one small whitespace comment.
(Feel free to merge after that's fixed)

tests/10apidoc/01register.pl
```diff
@@ -61,6 +61,30 @@
...
+
+   do => sub {
+      my ( $http ) = @_;
+
+      $http->do_request_json(
+         method => "POST",
+         uri    => "/unstable/createUser",
+
+         content => {
+            localpart     => "01register-user"
```
erikjohnston: Missing `,`

erikjohnston:
Just tested this locally, and the tests fails due to synapse returning a `400` due to not including an AS token.

@leonerd Do we have an easy way of getting at an AS token?

leonerd:
AS token?
Oh, is this related to application services? If so, then it probably wants to live in the `60app-services` directory. Once it's that late, it can get at the AS information in `our $AS_INFO` - see several other files around that area.

negzi:
Yes sorry about that, its not cool that I cant run the test. Maybe I put some more time and see if I can get the sytest to work. But in the mean time I need to add the app service token in the requests.args

I will take a look at 60app-services now

tests/10apidoc/01register.pl
negzi: Sure I will fix this and also I should say I have never used perl before. Sorry about my very clumsy mistakes here.

negzi:
@Leonerd
Here is an example of how to send a post request to create user feature.
````
curl -i -H "Accept: application/json" -H "Content-Type:application/json" -X POST --data '{"localpart": "user_local_part", "displayname": "user spla\
yname", "duration_seconds": 200 }' "http://192.168.99.101:8008/_matrix/client/api/v1/createUser?access_token=this_is_the_app_service_token"
basically the payload is :
 '{"localpart": "user_local_part", "displayname": "user displayname", "duration_seconds": 200 }'
```
and as you can see in the url I need to add  ?access_token=this_is_the_app_service_token"
parameter.
And this is an example of what I get back.
````
{
    "access_token": "MDAxN2xvY2F0aW9uIGxv",
    "home_server": "localhost",
    "user_id": "@user_local_part:localhost"
}
```
I have looked at the code under the 60app-services  directory and think the best  place to add my code would be in 01as-create.pl
I have a question though, I didn’t get how to add  the  app service  token. In the test, could you please let me know.
Also if 01as-create.pl is a correct place to add my test.

negzi:
closing this in favour of #252

PR #244, opened 2016-06-14
More fixes needed to test the synchrotron.

NegativeMjark:
It's a bit of a messy PR because I had to move all the .*_synced subs to 10apidoc so that I could use them in some of the earlier tests.


tests/90jira/SYN-627.pl
```diff
@@ -23,7 +25,9 @@
             my $room_id = $_;
 
             matrix_send_room_text_message( $user, $room_id, body => "$room_id" );
-         } @rooms );
+         } ( head -1, @rooms) );
```
leonerd: If this is the only use of `head` here, you could just use

    $rooms[-1]

to save an entire new CPAN import of a module. `head` is more intended for cases like slicing the result of another function call, or getting more than just the final element.

NegativeMjark: ``head -1, @list`` returns everything *except* the last element?

NegativeMjark: I suppose I could do something like ``@rooms[0..($rooms - 2)]``?

NegativeMjark: Done b36584f

leonerd: Ohyes; head is not tail. Oops. :)
In that case, you could use the  `@rooms[0..@rooms-2]` notation,or just stick with `List::Slice` as it was.

leonerd:
LGTM

PR #245, opened 2016-06-15
Test that backfilled events don't come down stream

erikjohnston:
See https://github.com/matrix-org/synapse/pull/869

### DIFF
--- a/tests/30rooms/04messages.pl
+++ b/tests/30rooms/04messages.pl
@@ -306,7 +306,8 @@ test "Message history can be paginated over federation",
       })->then( sub {
          my ( $event_id ) = @_;
 
-         # Wait for the message we just sent.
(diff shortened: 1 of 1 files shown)


tests/30rooms/04messages.pl
```diff
@@ -298,6 +300,22 @@
...
             'chunk[4] content body' );
 
-         Future->done(1);
+         matrix_send_room_text_message( $creator, $room_id,
+            body => "Marker message"
+         )
+      })->then( sub {
+         my ( $event_id ) = @_;
+
+         # Wait for the message we just sent.
```
NegativeMjark: Maybe say why we are waiting for the message we just sent?

NegativeMjark:
LGTM

PR #246, opened 2016-06-17
Change how the workers are configured to deduplicate common fields.

NegativeMjark:


NegativeMjark:
Needs matrix-org/synapse#874

PR #248, opened 2016-06-20
Allocate TCP ports from a predictable pool

leonerd:
Allocate TCP ports from a pool given as a commandline argument, instead of inventing arbitrary offsets from the main server ports. This allows multiple instances run concurrently on one machine with much more predictability between them, so they won't clash in allocated port numbers.

### DIFF
--- a/jenkins/prep_sytest_for_postgres.sh
+++ b/jenkins/prep_sytest_for_postgres.sh
@@ -18,19 +18,14 @@ if [ -z "$POSTGRES_DB_2" ]; then
     exit 1
 fi
 
-if [ -z "$PORT_BASE" ]; then
-    echo >&2 "Variable PORT_BASE not set"
-    exit 1
-fi
-
-mkdir -p "localhost-$(($PORT_BASE + 1))"
-mkdir -p "localhost-$(($PORT_BASE + 2))"
+mkdir -p "localhost-0"
+mkdir -p "localhost-1"
 
 : PGUSER=${PGUSER:=$USER}
 
 # We leave user, password, host blank to use the defaults (unix socket and
 # local auth)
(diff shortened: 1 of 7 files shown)


leonerd:
Ugh. :( Of course, jenkins will fail this because I've changed the actual commandline arguments that jenkins itself will use to run it.

I'll poke it tomorrow...

NegativeMjark:
@leonerd Jenkins uses the scripts in https://github.com/matrix-org/sytest/tree/paul/port-pool/jenkins to run the tests. You may just be able to update those scripts.

However you will probably need to change the PORT_BASE environment variable in jenkins, it's currently configured based on the current port allocation "strategy".

leonerd:
Updated jenkins tests; PTAL

NegativeMjark:
Couple of problems:

1) I can't run the tests with dendron because we don't allocate an ``unsecure_port`` for dendron to proxy to.

2) I have scripts for running sytest against different database backends which work by modifying the database config files before the tests are run. These fail because I don't know where the *second* synapse's config directory is, so I can't modify the files.

3) The default ranges of ports starts at 8000. This collides with the default unsecure homeserver port of 8008.

NegativeMjark:
You will probably want to fix https://github.com/matrix-org/sytest/blob/paul/port-pool/jenkins/prep_sytest_for_postgres.sh as well.

leonerd:
@NegativeMjark point 3 fixed by d90a142

leonerd:
Point 2 fixed by 36be1d4

leonerd:
Point 1 possibly fixed by 8875578 (untested with dendron as I don't have it to hand)

leonerd:
I've probably fixed the remaining `jenkins` script in 242ca85

NegativeMjark:
LGTM

PR #249, opened 2016-06-27
Paginated sync tests

erikjohnston:
https://github.com/matrix-org/synapse/pull/893


tests/31sync/20paginated.pl
```diff
@@ -0,0 +1,857 @@
...
+         } @{$room->{state}{events}} or die "Expected creation event";
+
+         pass "Got full state for room that had been seen but was then limited";
+
+         Future->done( 1 );
+      });
+   };
+
+
+multi_test "Can request unsen room",
```
NegativeMjark: Spelling "unseen".
Maybe say what API this is using?

NegativeMjark:
I think the tests tiles could be clearer about what the test is testing.
Perhaps some short comment outlining what each test is checking would help here?

PR #251, opened 2016-06-30
Dockerfile

superdump:
A quickly put together Dockerfile for building a functional SyTest test environment containing SyTest and synapse from their respective develop branches at the time of building.

You should be able to use git to update those clones as you please.

Once built, if the image name is `sytest` for example, you can run `docker run --rm sytest ./run-tests.pl` to run the tests in a new container created from the image.


NegativeMjark:
Firstly, thanks for the Dockerfile, its good to see contributions that make it easier for people to test changes. However when reviewing it I realised that we forgot to add a license or contributing.rst to the project. We've now added an Apache License version 2 and have the same contributing guidelines as synapse. If you are happy with that then please add this line to the PR description:

    Signed-off-by: Your Name <your@email.example.org>

Thanks, Mark

PR #252, opened 2016-06-30
Add test for createUser endpoint

negzi:
The createUser is working like:
```
Here is an example of how to send a post request to create user feature.

curl -i -H "Accept: application/json" -H "Content-Type:application/json" -X POST --data '{"localpart": "user_local_part", "displayname": "user spla\
yname", "duration_seconds": 200 }' "http://192.168.99.101:8008/_matrix/client/api/v1/createUser?access_token=this_is_the_app_service_token"
basically the payload is :
 '{"localpart": "user_local_part", "displayname": "user displayname", "duration_seconds": 200 }'
and as you can see in the url I need to add ?access_token=this_is_the_app_service_token"
parameter.
And this is an example of what I get back.

{
    "access_token": "MDAxN2xvY2F0aW9uIGxv",
    "home_server": "localhost",
    "user_id": "@user_local_part:localhost"
}
```

### DIFF
--- a/tests/60app-services/01as-create.pl
+++ b/tests/60app-services/01as-create.pl
@@ -27,6 +27,41 @@ test "AS can create a user",
       });
    };
 
+test "AS can get or create user and return an access_token",
+   requires => [ $main::AS_USER[0], $main::API_CLIENTS[0] ],
+
+   do => sub {
+      my ( $as_user, $http ) = @_;
+
+      do_request_json_for( $as_user,
+         method => "POST",
+         uri    => "/unstable/createUser",
+
+         content => {
+            localpart        => "user_localpart",
+            displayname      => "user_displayname",
+            duration_seconds => 200,
+         }
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( access_token user_id  home_server ));
+
+         my $user =User( $http , $body->{user_id}, "", $body->{access_token}, undef, undef, undef, [], undef );
+
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/r0/profile/:user_id/displayname",
+         )}
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_eq( $body->{displayname}, qw( user_displayname ));
+
+         Future->done(1);
+      });
+   };
+
 test "AS cannot create users outside its own namespace",
    requires => [ $main::AS_USER[0] ],
 


erikjohnston:
@matrixbot ok to test

erikjohnston:
Oh, matrixbot doesn't run against PRs? :cry:

negzi:
:(

negzi:
Oh by the way now I was able to run the tests easily using the dockerfile Rob made today :)

erikjohnston:
@matrix-bot ok to test

erikjohnston:
@matrixbot ok to test

erikjohnston:
@matrixbot test this please

erikjohnston:
@matrixbot test this please

erikjohnston:
@matrixbot test this please

erikjohnston:
@matrixbot test this please

erikjohnston:
@matrixbot test this please

erikjohnston:
@matrixbot test this please

erikjohnston:
Right, looks like the tests are now up and running. They're just busy installing (and testing) all the CPAN modules.

erikjohnston:
@negzi Might be worth checking that the access token works? E.g. you could try hitting the profile API?

negzi:
@erikjohnston That sounds great. Would like to test if the access token works. But do you mean writing a test case in syTest. In that case where could I write that test. Is there a specific file where i can write the test

erikjohnston:
I'd just add it to that test. So something like:

```perl
   )->then( sub {
	 my ( $body ) = @_;

	 log_if_fail "Body", $body;

	 assert_json_keys( $body, qw( access_token user_id  home_server ));

	 # Create a new user object, as in `matrix_register_user` 
	 my $user = User( $http, $body->{user_id}, "", $access_token, undef, undef, undef, [], undef );

	 do_request_json_for( $user,
	    method => "GET",
	    uri    => "/r0/profile/:user_id/displayname",
	 )
   });
```

You could also then check that the display name returned matches the display name sent, but that's a minor thing.

negzi:
@erikjohnston:

I have made the changes, I also check if the returned display name matches. Please take a look when you have time. Thank you!

tests/60app-services/01as-create.pl
```diff
@@ -27,6 +27,40 @@
...
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( access_token user_id  home_server ));
+
+         my $user =User( $http , $body->{user_id}, "", $body->{access_token}, undef, undef, undef, [], undef );
+
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/r0/profile/:user_id/displayname",
+         )})->then(sub {
```
erikjohnston: This doesn't need to be nested, as `->then( ... )` chains things correctly. For example:

```perl
foo()
->then( sub {
   bar();
})->then( sub {
   my ( $bar_result ) = @_;

   ...
});
```

Does the right thing, i.e. `$bar_result` will be the result of the `bar()` invocation, assuming it returns a promise/deferred.

(In perl, the last expression in a `sub { ... }` is used as the return value.)

This is really nice as it means you don't end up with tonnes of nesting, and generally makes it easier to read.

---

You also want a space between the parenthesis and `sub`, lest you make leo sad.

negzi: @erikjohnston 
I think here the indentation on line 55 is wrong. Am I wrong so if I change the indentation like below, then it should be fine or?
```
         content => {
            localpart        => "user_localpart",
            displayname      => "user_displayname",
            duration_seconds => 200,
         }
      )->then( sub {
         my ( $body ) = @_;

         assert_json_keys( $body, qw( access_token user_id  home_server ));

         my $user =User( $http , $body->{user_id}, "", $body->{access_token}, undef, undef, undef, [], undef );

         do_request_json_for( $user,
            method => "GET",
            uri    => "/r0/profile/:user_id/displayname",
         )}
      )->then( sub {
         my ( $body ) = @_;

         assert_eq( $body->{displayname}, qw( user_displayname ));

         Future->done(1);
      });
   };
```

negzi: By the way you are explaining really good, I find it really helpful :)

erikjohnston: Ah, yes, there's a missing newline. If you commit that change then we should be good to go :)

>  By the way you are explaining really good, I find it really helpful :)

Thanks :D

negzi: cool!

negzi:
Signed-off-by: Negar Fazeli negar.fazeli@ericsson.com

erikjohnston:
Woo! Thanks! Sorry how much of a faff this has been to do :(

negzi:
No worries, I have learned a bit of Perl ;p

tests/60app-services/01as-create.pl
```diff
@@ -27,6 +27,41 @@
...
+            localpart        => "user_localpart",
+            displayname      => "user_displayname",
+            duration_seconds => 200,
+         }
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( access_token user_id  home_server ));
+
+         my $user =User( $http , $body->{user_id}, "", $body->{access_token}, undef, undef, undef, [], undef );
```
leonerd: Whitespace after the `=`

tests/60app-services/01as-create.pl
```diff
@@ -27,6 +27,41 @@
...
+
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/r0/profile/:user_id/displayname",
+         )}
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_eq( $body->{displayname}, qw( user_displayname ));
```
leonerd: Stylistically I wouldn't use `qw()` here. That would lead the reader into suspecting that the string is a "structural" string - such as a hash key or field name. Instead, this is just general string data, expected to match the `displayname` parameter earlier passed. I'd write this

    assert_eq( $body->{displayname}, "user_displayname" );

PR #256, opened 2016-07-04
Extract a base `SyTest::Homeserver` class

leonerd:
In case of creating other kinds of homeserver wrapping in the future (e.g. dendron, ruma, ...)


NegativeMjark:
Looks like perl to me. LGTM I guess?

PR #258, opened 2016-07-05
mkdir() the hs_dir earlier on in HS setup in case of other config fil…

leonerd:
…es that need writing before we call ->start


NegativeMjark:
LGTM

PR #259, opened 2016-07-06
Test registering via shared secret

erikjohnston:


cpanfile
```diff
@@ -4,6 +4,7 @@ requires 'Class::Method::Modifiers';
 requires 'Data::Dump';
 requires 'DBD::Pg';
 requires 'Digest::SHA';
+requires 'Digest::HMAC_SHA1';
```
leonerd: Can you keep these alphabetically sorted?

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+
+sub matrix_register_user_via_secret
+{
+   my ( $http, $uid, %opts ) = @_;
+
+   my $password = $opts{password} // "an0th3r s3kr1t";
+   my $admin = $opts{admin} // 0;
+
+   defined $uid or
+      croak "Require UID for matrix_register_user";
```
leonerd: Name of the function in the error message needs updating

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+
+   my $hmac = Digest::HMAC_SHA1->new("reg_secret");
+   $hmac->add($uid);
+   $hmac->add($password);
+   $hmac->add($admin ? "admin" : "notadmin");
```
leonerd: Whitespaces inside parens in these four lines above

leonerd: Also, more of a spec thought - isn't it usual to include some sort of separator character between these fields, so as to avoid certain collision attacks?

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+
+      assert_json_keys( $body, qw( user_id access_token ));
+
+      my $access_token = $body->{access_token};
+
+      my $user = User( $http, $body->{user_id}, $password, $access_token, undef, undef, undef, [], undef );
+
+      my $f = Future->done;
+
+      return $f->then_done( $user )
```
leonerd: Not sure what these two lines are doing. You're creating an already-done future, then making a new one from it with `->then_done`. Why not just have

   return Future->done( $user )
      ->on_done ...

in the first place?

tests/10apidoc/01register.pl
erikjohnston: Err, yes.

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+       matrix_register_user_via_secret( $http, $uid, admin => 0 );
+   };
+
+test "POST /register admin with shared secret",
+   requires => [ $main::API_CLIENTS[0], localpart_fixture() ],
+
+   do => sub {
+       my ( $http, $uid ) = @_;
+
+       matrix_register_user_via_secret( $http, $uid, admin => 1 );
```
leonerd: Is there some sort of `check` block that can be done to check that the user really is an admin-enabled user?

erikjohnston: Ah, I was copy pasting without thought.

tests/10apidoc/01register.pl
erikjohnston: Ah, I was copy pasting without thought.

tests/10apidoc/01register.pl
erikjohnston: Not at the moment─I want to add one─but right now I just want to ensure that it doesn't implode when you set `admin` to true.

erikjohnston:
Right, the "Merged PR" tests aren't happy because the associated synapse PR hasn't landed: https://github.com/matrix-org/synapse/pull/909

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+      croak "Require UID for matrix_register_user_via_secret";
+
+   my $hmac = Digest::HMAC_SHA1->new("reg_secret");
+   $hmac->add( $uid );
+   $hmac->add( "\0" );
+   $hmac->add( $password );
+   $hmac->add( "\0" );
+   $hmac->add( $admin ? "admin" : "notadmin" );
+
+   my $mac = $hmac->hexdigest;
```
leonerd: Actually, looking at the module there's a more convenient "all-in-one" function of `hmac_sha1_hex()` you could use instead.

    use Digest::HMAC_SHA1 qw( hmac_sha1_hex );

    my $mac = hmac_sha1_hex(
       join( "\0", $uid, $password, $admin ? "admin" : "notadmin" ),
       "reg_secret"
    );

That might look shorter.

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
+{
+   my ( $http, $uid, %opts ) = @_;
+
+   my $password = $opts{password} // "an0th3r s3kr1t";
+   my $admin = $opts{admin} // 0;
+
+   defined $uid or
+      croak "Require UID for matrix_register_user_via_secret";
+
+   my $hmac = Digest::HMAC_SHA1->new("reg_secret");
```
leonerd: Hrm. This shared-secret has to be the same value as the one given at synapse configuration time? This seems a bit fragile to just keep two string literals that must have equal value - maybe we can store it somewhere?

tests/10apidoc/01register.pl
```diff
@@ -126,7 +128,75 @@ sub matrix_register_user
...
 
-push @EXPORT, qw( local_user_fixture local_user_fixtures );
+push @EXPORT, qw( matrix_register_user_via_secret );
+
+sub matrix_register_user_via_secret
+{
+   my ( $http, $uid, %opts ) = @_;
+
+   my $password = $opts{password} // "an0th3r s3kr1t";
+   my $admin = $opts{admin} // 0;
```
leonerd: Purely stylistically - I tend to name boolean variables `$is_*`; so `$is_admin` in this case.

tests/10apidoc/01register.pl
erikjohnston: Agreed. Any suggestions where?

leonerd:
LGTM. Jenkins still complains about the added tests but other ones are fine.

erikjohnston:
@matrixbot retest this please

PR #263, opened 2016-07-14
Add test to send and fetch image events

erikjohnston:


leonerd:
After a long Jenkins wait... LGTM. finally.

PR #267, opened 2016-07-20
Kill the old listeners before running sytest

NegativeMjark:


jenkins/kill_old_listeners.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+# Kill any stray processes that left over from a previous test run
+# that are holding onto one of the ports we want to use.
+for port in $(seq $PORT_BASE $((PORT_BASE+PORT_COUNT-1))); do
+    lsof -i TCP:$port | grep LISTEN | awk '{print $2}' | xargs kill -9
```
leonerd: Presumably this is running as as "jenkins"-type user, so there's no chance of accidentally killing essential bits of system process?

NegativeMjark: Yep

leonerd:
LGTM

PR #269, opened 2016-07-21
Implement some tests for the initial devices APIs

richvdh:
Check that devices are recorded when we login and register, and that the GET
apis return the new devices

### DIFF
--- a/tests/10apidoc/12device_management.pl
+++ b/tests/10apidoc/12device_management.pl
@@ -1,12 +1,12 @@
 our @EXPORT = qw( matrix_get_device );
 
 sub matrix_get_device {
-   my ($user, $device_id) = @_;
+   my ( $user, $device_id ) = @_;
 
    return do_request_json_for(
       $user,
       method => "GET",
-      uri => "/unstable/devices/${device_id}",
+      uri    => "/unstable/devices/${device_id}",
    );
 }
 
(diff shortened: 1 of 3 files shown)


tests/10apidoc/12device_management.pl
```diff
@@ -0,0 +1,101 @@
+our @EXPORT = qw( matrix_get_device );
+
+sub matrix_get_device {
+   my ($user, $device_id) = @_;
```
leonerd: Whitespace `( $user, $device_id )`

tests/10apidoc/12device_management.pl
```diff
@@ -0,0 +1,101 @@
...
+
+   return do_request_json_for(
+      $user,
+      method => "GET",
+      uri => "/unstable/devices/${device_id}",
```
leonerd: Whitespace

tests/10apidoc/12device_management.pl
```diff
@@ -0,0 +1,101 @@
...
+             uri => "/unstable/devices",
+          );
+      })->then( sub {
+         my ( $devices ) = @_;
+         log_if_fail ("/devices response: ", $devices);
+         assert_json_keys(
+            $devices,
+            qw( devices ),
+           );
+         assert_json_list($devices->{devices});
```
leonerd: Whitespace inside parens

tests/10apidoc/12device_management.pl
```diff
@@ -0,0 +1,101 @@
...
+
+         # check each of the devices we logged in with is returned
+         for my $id ( keys( %DEVICES )) {
```
leonerd: For consistency this wants to be `foreach`; as all the others are

tests/10apidoc/12device_management.pl
```diff
@@ -0,0 +1,101 @@
...
+            my $matched = 0;
+            RESULT: foreach my $result ( @{ $devices->{devices}} ) {
```
leonerd: FYI: Rather than a second inner `foreach` loop it might be neater to do something like

    my $result = first { $_->{device_id} eq $id } @{ $devices->{devices} };

then test that you got something, and that it was the right thing. That way you avoid keeping most of the code double-nested inside both loops.

Just a thought; how it's written now is fine.

leonerd:
Code LGTM. Jenkins result still pending

PR #271, opened 2016-07-27
Update sytests to match e2e key upload changes

richvdh:
The e2e endpoints have never been specced, so we should be accessing them on an
/unstable endpoint for now.

The 'deviceId' parameter on e2e key upload is now optional (and where it
exists, it must match that from the login) - so we can no longer just make it
up. Stash the device_id when we register or login, and use it in the e2e keys.

GET /keys/upload has never even been proposed as an API, and has now been
removed from synapse. Test that keys have been claimed using a second call to
/claim instead.

### DIFF
--- a/tests/10apidoc/00prepare.pl
+++ b/tests/10apidoc/00prepare.pl
@@ -2,7 +2,7 @@ our @EXPORT = qw( User is_User do_request_json_for );
 
 # A handy little structure for other scripts to find in 'user' and 'more_users'
 struct User =>
-   [qw( http user_id password access_token refresh_token eventstream_token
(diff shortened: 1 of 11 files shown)


richvdh:
(corresponds to changes in https://github.com/matrix-org/synapse/pull/956)

leonerd:
LGTM

I wonder if the `User` struct is getting sufficiently large now that we should switch it to using named args to the constructor. I might do that in a separate commit.

richvdh:
> I wonder if the User struct is getting sufficiently large now that we should switch it to using named args to the constructor. I might do that in a separate commit.

I couldn't believe how much of a pita it was to add device_id. Something Must Be Done.

PR #272, opened 2016-07-27
Test that ASes may begin user IDs with '_'; normal users cannot

leonerd:
Tests the code submitted in https://github.com/matrix-org/synapse/pull/958


leonerd:
Tests code in https://github.com/matrix-org/sytest/pull/272

NegativeMjark:
LGTM

PR #275, opened 2016-08-01
Don't rely on signal handlers to fire during END blocks

leonerd:
Because of

  https://rt.perl.org/Public/Bug/Display.html?id=128774

Which additionally I didn't notice myself because my machine runs using `IO::Async::Loop::Epoll`, and manages to sidestep this issue because it is working around its own annoyance with perl core bugs:

  https://metacpan.org/source/PEVANS/IO-Async-Loop-Epoll-0.17/lib/IO/Async/Loop/Epoll.pm#L377


NegativeMjark:
LGTM

NegativeMjark:
retest this please

PR #277, opened 2016-08-02
Factor out some of the code shared between sytest scripts

NegativeMjark:


run-tests.pl
```diff
@@ -94,7 +94,7 @@
 
    'bind-host=s' => \$BIND_HOST,
 
-   'p|port-range=s' => \(my $PORT_RANGE = "8800:8819"),
+   'p|port-range=s' => \(my $PORT_RANGE = "8800:8900"),
```
richvdh: `port-max` appears to be inclusive so this might be better as `8899` ?

jenkins/install_and_run.sh
```diff
@@ -14,6 +14,21 @@ cd "`dirname $0`/.."
...
+: ${BIND_HOST=localhost}
+
+export PORT_BASE
+export PORT_COUNT
+export BIND_HOST
+
 ./jenkins/kill_old_listeners.sh
 
-./run-tests.pl --bind-host ${BIND_HOST:-localhost} -O tap --all "$@" > results.tap
+export TOX_BIN=$WORKSPACE/.tox/py27/bin
```
richvdh: is there a reason to prefer `$WORKSPACE` over `pwd` here? seems like this unnecessarily ties it to jenkins.

NegativeMjark: Yes, using WORKSPACE means that this script will also work when sytest is checked out as a subfolder in the synapse and dendron jenkins tests.

run-tests.pl
NegativeMjark: Done 3ea0c89

jenkins/clone.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/bash
+
+NAME=$1
+PROJECT=$2
+BASE=".$NAME-base"
+
+# update our clone
+if [ ! -d .$NAME-base ]; then
```
richvdh: this script could do with some quotes around its $vars, so as to reduce breakage in the event of spaces in filenames

jenkins/clone.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/bash
```
richvdh: could do with a comment saying what this does and what the args are

jenkins/clone.sh
```diff
@@ -0,0 +1,24 @@
...
+git clone $BASE $NAME --shared
+
+: ${GIT_BRANCH:="origin/$(git rev-parse --abbrev-ref HEAD)"}
+cd $NAME
+# check out the relevant branch
+git checkout "${GIT_BRANCH}" || (
+    echo >&2 "No ref ${GIT_BRANCH} found, falling back to develop"
+    git checkout "origin/develop"
+)
+git clean -df .
```
richvdh: this doesn't feel like it belongs here

NegativeMjark:
@richvdh PTAL

jenkins/install_and_run.sh
richvdh: ugh. fair enough

richvdh:
lgtm

PR #279, opened 2016-08-03
Test querying devices for a user with no devices

richvdh:
This should return an empty dict, rather than no dict. (per
https://github.com/matrix-org/synapse/pull/977)

### DIFF
--- a/tests/41end-to-end-keys/01-upload-key.pl
+++ b/tests/41end-to-end-keys/01-upload-key.pl
@@ -123,9 +123,9 @@ test "Can query device keys using GET",
    };
 
 test "query for user with no keys returns empty key dict",
-  requires => [ local_user_fixture() ],
+   requires => [ local_user_fixture() ],
 
-  check => sub {
+   check => sub {
       my ( $user ) = @_;
 
       do_request_json_for( $user,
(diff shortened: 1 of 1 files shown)


tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -122,6 +122,39 @@
       })
    };
 
+test "query for user with no keys returns empty key dict",
+  requires => [ local_user_fixture() ],
```
leonerd: Indentation. sw=3
(Do you use vim, you could just set that automatically :) )

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -122,6 +122,39 @@
...
+
+         assert_json_keys( $content, "device_keys" );
+
+         my $device_keys = $content->{device_keys};
+         assert_json_keys( $device_keys, $user->user_id );
+
+         my $alice_keys = $device_keys->{ $user->user_id };
+
+         assert_json_object( $alice_keys );
+         assert_ok( !%{$alice_keys}, "unexpected keys" );
```
leonerd: FYI braces aren't strictly needed here, but they're fine to leave in.

leonerd:
LGTM besides the indentation

PR #280, opened 2016-08-03
keys/query should include device display name

richvdh:
Per https://github.com/matrix-org/synapse/pull/978

### DIFF
--- a/tests/10apidoc/12device_management.pl
+++ b/tests/10apidoc/12device_management.pl
@@ -1,4 +1,4 @@
-our @EXPORT = qw( matrix_get_device );
+our @EXPORT = qw( matrix_get_device matrix_set_device_display_name );
 
 sub matrix_get_device {
    my ( $user, $device_id ) = @_;
@@ -8,6 +8,19 @@ sub matrix_get_device {
       method => "GET",
       uri    => "/unstable/devices/${device_id}",
    );
+}
+
+sub matrix_set_device_display_name {
+    my ( $user, $device_id, $display_name ) = @_;
+
+    return do_request_json_for(
+        $user,
+        method => "PUT",
+        uri    => "/unstable/devices/${device_id}",
+        content => {
+            display_name => $display_name,
+        },
+    );
 }
 
 test "GET /device/{deviceId}",
--- a/tests/41end-to-end-keys/01-upload-key.pl
+++ b/tests/41end-to-end-keys/01-upload-key.pl
@@ -53,6 +53,8 @@ test "Can query device keys using POST",
       )->then( sub {
          my ( $content ) = @_;
 
+         log_if_fail( "/query response", $content );
+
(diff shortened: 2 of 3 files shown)


tests/10apidoc/12device_management.pl
```diff
@@ -10,6 +10,19 @@ sub matrix_get_device {
    );
 }
 
+sub matrix_set_device_display_name {
```
leonerd: Hrm. This brace should have been on its own line, but then so should the other functions in the file. Don't worry about it; I'll do a style fixup later.

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -53,14 +53,29 @@
...
+
+         my $alice_device_keys = $alice_keys->{ $user->device_id };
+         assert_json_keys( $alice_device_keys, "unsigned" );
+
+         my $unsigned = $alice_device_keys->{unsigned};
+
+         # display_name should be null by default
+         exists $unsigned->{device_display_name} or
+           die "Expected to get a (null) device_display_name";
+         defined $unsigned->{device_display_name} and
```
leonerd: This tests that the key exists but has an undefined value. Are we OK with adding that kind of distinction? I seem to recall it being quite a rare thing in our API that we make use of `null` values in JSON. Maybe @NegativeMjark has a thought here?

tests/41end-to-end-keys/04-query-key-federation.pl
```diff
@@ -27,7 +29,14 @@
 
          my $alice_keys = $device_keys->{ $user->user_id };
          assert_json_keys( $alice_keys, $user->device_id );
+
+         my $alice_device_keys = $alice_keys->{ $user->device_id };
+
          # TODO: Check that the content matches what we uploaded.
+
+         assert_eq( $alice_device_keys->{"unsigned"}->{"device_display_name"},
```
leonerd: FYI, second arrow here is unnecessary; perl doesn't care after the first.

tests/41end-to-end-keys/01-upload-key.pl
NegativeMjark: I dunno. I'd guess that we should handle this in the same way that user displaynames are handled.

richvdh: Yeah, fair enough. The instances of that which I've checked in the spec omit the key when no displayname is set.

richvdh: updated in https://github.com/matrix-org/synapse/pull/981.

richvdh:
(tests are failing because the fix is on a different branch)

richvdh:
matrixbot: please retest

richvdh:
matrixbot: retest this please

even

PR #293, opened 2016-08-25
Extend 3PE lookup APIs for metadata query

leonerd:
Tests for updated API and implementation in
  https://github.com/matrix-org/synapse/pull/1041


NegativeMjark:
LGTM

PR #294, opened 2016-08-25
Root the 3PE lookup API within /_matrix/app/unstable

leonerd:
Because newly added paths ought to go here, not at toplevel.


NegativeMjark:
LGTM

PR #295, opened 2016-08-25
Add reliability tests for direct-to-device messaging

NegativeMjark:
Adds tests for matrix-org/synapse#1046

### DIFF
--- a/tests/46direct/01directmessage.pl
+++ b/tests/46direct/01directmessage.pl
@@ -1,30 +1,30 @@
-my $matrix_send_direct_message_txn_id = 0;
+my $matrix_send_to_device_message_txn_id = 0;
 
-push our @EXPORT, qw( matrix_send_direct_message );
+push our @EXPORT, qw( matrix_send_to_device_message );
 
-sub matrix_send_direct_message
+sub matrix_send_to_device_message
 {
    my ( $user, %params ) = @_;
    exists $params{type} or die "Expected a type";
    exists $params{messages} or die "Expected messages";
    my $type = delete $params{type};
-   my $txn_id = delete $params{txn_id} // ++$matrix_send_direct_message_txn_id;
+   my $txn_id = delete $params{txn_id} // ++$matrix_send_to_device_message_txn_id;
 
    do_request_json_for( $user,
       method  => "PUT",
-      uri     => "/unstable/sendDirect/$type/$txn_id",
+      uri     => "/unstable/sendToDevice/$type/$txn_id",
       content => \%params,
    );
 }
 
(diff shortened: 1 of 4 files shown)


tests/46direct/01directmessage.pl
```diff
@@ -0,0 +1,143 @@
+use Future::Utils qw( repeat );
+
+my $next_device_message_txn_id = 0;
+
+push our @EXPORT, qw( matrix_send_to_device_message );
+
+sub matrix_send_to_device_message
```
leonerd: A bold choice of name there. Maybe `send_message_to_device`; or perhaps is the "message" part implied? Could it be just `send_to_device` ?

tests/46direct/01directmessage.pl
```diff
@@ -0,0 +1,143 @@
...
+{
+   my ( $user, %params ) = @_;
+   exists $params{type} or die "Expected a type";
+   exists $params{messages} or die "Expected messages";
+   my $type = delete $params{type};
```
leonerd: Is `type` ever false? If not it might be more direct to

    my $type = delete $params{type} or die "Expected a 'type'";

tests/46direct/01directmessage.pl
```diff
@@ -0,0 +1,143 @@
...
+
+sub matrix_recv_and_ack_to_device_message
+{
+   my ( $user ) = @_;
+
+   matrix_recv_to_device_message( $user )->then( sub {
+      my ( $messages, $next_batch ) = @_;
+
+      matrix_ack_to_device_message( $user, $next_batch )->then( sub {
+         Future->done( $messages );
```
leonerd: Since `$messages` is already in scope and has its value by now, you can write this neater as

    ...->then_done( $messages );

leonerd:
Having got to the end I now see that *to device message* is in fact a compound noun, not part of the verb structure. I wonder if this code would be a bit more readable were these renamed to something else, perhaps a "device-directed message" or similar? If that name sounds a bit long, I don't think it would be unreasonable to document/specify these at the protocol level and introduce the initialism *DDM* to refer to it; then we can name functions

    sub matrix_send_DDM
    sub matrix_recv_DDM
    sub matrix_recv_and_ack_DDM
    etc...

NegativeMjark:
@leonerd I've called it a ``device_message`` for now since DDM is probably a bit meaningless for someone trying to read the code.

```
sub matrix_send_device_message
sub matrix_recv_device_message
sub matrix_recv_and_ack_device_message
```

leonerd:
LGTM

PR #297, opened 2016-08-30
Check room names in push are correct

erikjohnston:


tests/61push/01message-pushed.pl
```diff
@@ -226,3 +226,158 @@
...
+         ( $room_id ) = @_;
+
+         do_request_json_for( $bob,
+            method => "PUT",
+            uri    => "/r0/directory/room/$room_alias",
+
+            content => { room_id => $room_id },
+         )
+      })->then( sub {
+         do_request_json_for( $bob,
```
leonerd: I believe there's a helper function for this; `matrix_put_room_state()` in `tests/10apidoc/31room-state.pl`.

tests/61push/01message-pushed.pl
```diff
@@ -226,3 +226,158 @@
          Future->done(1);
       });
    };
+
+
+sub setup_push
```
leonerd: Could be worth adding a small line of documentation to remark that the returned future yields a new room ID; normally I'd expect a `setup_*` future to be uninteresting.

erikjohnston:
@leonerd PTAL

leonerd:
LGTM

PR #298, opened 2016-09-01
Truncate the worker log files as well as the main log file

NegativeMjark:


leonerd:
LGTM.

I slightly wonder whether it would lead to confusion by creating files for synchrotron/dendron/fed-reader if those parts aren't even in use, but probably not important.

PR #299, opened 2016-09-02
Add a new_User sub that takes named parameters

NegativeMjark:


tests/10apidoc/00prepare.pl
```diff
@@ -26,3 +28,19 @@ sub do_request_json_for
       %args,
    );
 }
+
+
+sub new_User
```
leonerd: Rather than create this as a function you could just use the `-named_constructor` form; see https://metacpan.org/pod/Struct::Dumb#CONSTRUCTOR-FORMS

leonerd: On second inspection, that would complicate the optional nature of most of the parameters and the defaulting arrayref for `saved_events`; so may be best to leave it as it is.

leonerd:
LGTM, presuming Jenkins is happy.

PR #300, opened 2016-09-02
Check that joining a room is idempotent

NegativeMjark:


tests/30rooms/01state.pl
```diff
@@ -235,6 +235,20 @@
    };
 
 
+sub get_test_state {
```
leonerd: Opening `{` of a new toplevel function wants to be on a line of its own.

leonerd:
LGTM other than the one linefeed style issue

PR #301, opened 2016-09-06
Check that direct to device messages work over federation

NegativeMjark:
Tests matrix-org/synapse#1074


tests/46direct/03federation.pl
```diff
@@ -0,0 +1,57 @@
...
+
+         assert_deeply_eq( $messages, [{
+            sender  => $local_user->user_id,
+            type    => "my.test.type",
+            content => {
+               message => "first",
+            },
+         }]);
+
+         matrix_send_device_message( $local_user,
```
leonerd: It's not immediately obvious from the test why you're basically doing the same thing twice. Can you put a comment here to suggest why?

PR #302, opened 2016-09-07
Apply the same filtering to incremental test as was used for initial test

NegativeMjark:
When checking that for presence updates.


leonerd:
LGTM, presuming jenkins is happy.

PR #305, opened 2016-09-09
Test improvements to 3PE lookup API

leonerd:
Tests for new functionality added in https://github.com/matrix-org/synapse/pull/1091

### DIFF
--- a/tests/60app-services/05lookup3pe.pl
+++ b/tests/60app-services/05lookup3pe.pl
@@ -1,15 +1,29 @@
+use List::UtilsBy qw( sort_by );
+
 use constant AS_PREFIX => "/_matrix/app/unstable";
 
+sub stub_empty_result
+{
+   my ( $appserv, $path ) = @_;
+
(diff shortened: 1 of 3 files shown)


leonerd:
retest this please

NegativeMjark:
LGTM

tests/60app-services/05lookup3pe.pl
```diff
@@ -33,20 +68,80 @@
...
+test "HS can provide query metadata on a single protocol",
+   requires => [ local_user_fixture(), $main::APPSERV[0], $main::APPSERV[1] ],
+
+   proves => [qw( can_get_3pe_metadata )],
+
+   check => sub {
+      my ( $user, $appserv1, $appserv2 ) = @_;
+
+      # Awkwardly, this test relies on the caching within synapse of the
+      # results obtained in the previous test.
```
NegativeMjark: Why does it need to depend on the previous caching?

Could you explain what would be needed to avoid that dependency?

leonerd: Ah, will do that directly in `develop`, since I've already merged.

PR #306, opened 2016-09-09
Test clients requesting the whole of a single state event

leonerd:
Tests the change added in https://github.com/matrix-org/synapse/pull/1094


leonerd:
Jenkins SQLite test failure looks randomly spurious - possibly a timing race failure or something. Doesn't look related to this change.

NegativeMjark:
LGTM

PR #312, opened 2016-09-22
Add test for 3pid invite over federation

erikjohnston:
Tests that a 3pid invite is correctly handled when the invitee registers
with a server that is already in the room, other than the inviter's.

Tests: https://github.com/matrix-org/synapse/pull/1136


leonerd:
LGTM
I presume the failed "Merged PR" tests are because this needs corresponding synapse change?

erikjohnston:
Yup

erikjohnston:
@matrixbot retest this please

PR #315, opened 2016-09-27
Fix flakey presence /initialSync test

erikjohnston:


leonerd:
LGTM. (Sorry to have taken all day to get round to it; slack AS was being busy)

PR #317, opened 2016-10-12
Maybe make the push rule in sync test more reliable 

NegativeMjark:
By filtering out presence updates.

I think some of the spurious sytest failures are due to sync request returning prematurely because of presence updates caused by the sync requests. This causes the "push rules wake up sync" tests to fail because the sync request returns but doesn't contain any push rule events.

### DIFF
--- a/tests/61push/06_push_rules_in_sync.pl
+++ b/tests/61push/06_push_rules_in_sync.pl
@@ -1,6 +1,7 @@
 use List::Util qw( first );
 
-my $check_for_push_rules = sub {
+
+sub check_for_push_rules {
    my ( $sync_body ) = @_;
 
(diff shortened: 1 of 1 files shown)


tests/61push/06_push_rules_in_sync.pl
```diff
@@ -1,6 +1,7 @@
 use List::Util qw( first );
 
-my $check_for_push_rules = sub {
+
```
leonerd: Extraneous blank line

leonerd:
Overall LGTM.
I wonder given the growing number of sync tests that have to discard presence messages whether it might be a general pattern to extract into a new sync function, or parameter thereto

PR #318, opened 2016-10-13
User-interactive auth on device deletion

richvdh:
Device deletion now uses interactive auth, so check that and do it right.

### DIFF
--- a/tests/10apidoc/02login.pl
+++ b/tests/10apidoc/02login.pl
@@ -132,7 +132,7 @@ test "POST /login as non-existing user is rejected",
    requires => [ $main::API_CLIENTS[0],
                  qw( can_login_password_flow )],
 
-   bug => "SYN-680",
+   # bug => "SYN-680", # now fixed
 
    do => sub {
       my ( $http ) = @_;
--- a/tests/10apidoc/02login.pl
+++ b/tests/10apidoc/02login.pl
@@ -132,8 +132,6 @@ test "POST /login as non-existing user is rejected",
    requires => [ $main::API_CLIENTS[0],
                  qw( can_login_password_flow )],
 
-   # bug => "SYN-680", # now fixed
-
    do => sub {
       my ( $http ) = @_;
 
--- a/tests/10apidoc/12device_management.pl
+++ b/tests/10apidoc/12device_management.pl
@@ -1,3 +1,4 @@
+use JSON qw( decode_json );
 use Future::Utils qw( try_repeat_until_success );
 
 our @EXPORT = qw( matrix_get_device matrix_set_device_display_name );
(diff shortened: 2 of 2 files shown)


richvdh:
retest this please please please

tests/10apidoc/02login.pl
```diff
@@ -132,7 +132,7 @@
    requires => [ $main::API_CLIENTS[0],
                  qw( can_login_password_flow )],
 
-   bug => "SYN-680",
+   # bug => "SYN-680", # now fixed
```
leonerd: Any reason to keep this as a comment? I'd just outright delete it.

tests/10apidoc/12device_management.pl
```diff
@@ -25,6 +26,17 @@ sub matrix_set_device_display_name {
     );
 }
 
+sub matrix_delete_device {
+    my ($user, $device_id, $request_body) = @_;
```
leonerd: Wants whitespace inside parens:

    my ( $user, $device_id, $request_body ) = ...

tests/10apidoc/12device_management.pl
```diff
@@ -199,12 +211,46 @@ sub matrix_set_device_display_name {
...
          );
       })->then( sub {
-         # delete the device
-         do_request_json_for(
-            $user,
-            method => "DELETE",
-            uri    => "/unstable/devices/${DEVICE_ID}",
-         );
+         # attempt request with empty auth dict
+         matrix_delete_device($user, $DEVICE_ID, {});
```
leonerd: Ditto whitespace inside parens

tests/10apidoc/02login.pl
richvdh: thought it might be useful to keep a record of the bug we are testing against?

leonerd: If it's a true "regression test" for a known bug then it can live in `90jira/` named after the bug it asserts no longer exists. If it's just a test for generically specified behaviour then it's probably less important that it used to not work due to a bug.

richvdh:
I hope I've fixed all the issues ... ptal @leonerd ?

tests/10apidoc/12device_management.pl
```diff
@@ -199,12 +211,43 @@ sub matrix_set_device_display_name {
...
+
+         my $body = decode_json $resp->content;
+         log_if_fail( "Response to wrong password", $body );
+
+         assert_json_keys( $body, qw( error errcode session params flows ));
+
+         assert_eq( $body->{errcode}, "M_FORBIDDEN", 'errcode' );
+
+         # one more time with the right password
+         matrix_delete_device($user, $DEVICE_ID, {
```
leonerd: One more missing whitespace

leonerd:
LGTM besides a single missing space

leonerd:
Though while I look, jenkins appears to be upset... I shall look into it

leonerd:
The failing test there appears to be
    258 - User sees updates to presence from other users in the incremental sync.

which doesn't sound related to this change. I suspect it's another flaky one :(

tests/10apidoc/02login.pl
richvdh: whatevs. done.

tests/10apidoc/12device_management.pl
richvdh: done.

richvdh:
Yeah, I'm going to blame that on a flaky test.

PR #319, opened 2016-10-25
Remove unused flags from dendron command line

NegativeMjark:
These flags aren't used by dendron anymore, and will be removed at somepoint.


leonerd:
"Whatevs" (LGTM)

PR #320, opened 2016-11-03
Test we can register with a non-ascii password

dbkr:
Tests https://github.com/matrix-org/synapse/pull/1191

### DIFF
--- a/tests/11register.pl
+++ b/tests/11register.pl
@@ -289,8 +289,6 @@ test "registration accepts non-ascii passwords",
    do => sub {
       my ( $http, $localpart ) = @_;
 
-      my $session;
-
       $http->do_request_json(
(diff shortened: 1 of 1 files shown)


tests/11register.pl
```diff
@@ -282,3 +282,49 @@
...
+            initial_device_display_name => "display_name",
+         },
+      )->main::expect_http_401->then( sub {
+         my ( $response ) = @_;
+
+         my $body = decode_json $response->content;
+
+         assert_json_keys( $body, qw( session ));
+
+         $session = $body->{session};
```
leonerd: `$session` doesn't seem to be used outside this block so you might as well just declare it here; `my $session`.

leonerd:
LGTM

PR #321, opened 2016-11-11
Add PUT /rooms/$room_id/send/m.room.message/$txn_id tests 

Kegsay:
- A general one (we only had one for POST)
- One to test idempotency of PUTs

Also moved the `proves` statement to the `PUT` endpoint rather than the `POST` endpoint as the send message helper function uses the `PUT` form.


leonerd:
LGTM

PR #323, opened 2016-11-28
Make expected failures clearer

richvdh:
I can never tell which tests are expected to fail, so put some text in the
test name.


richvdh:
retest this please

leonerd:
LGTM

PR #328, opened 2016-12-07
Test AS specific room lists

erikjohnston:
Tests: https://github.com/matrix-org/synapse/pull/1676

### DIFF
--- a/tests/60app-services/06publicroomlist.pl
+++ b/tests/60app-services/06publicroomlist.pl
@@ -1,0 +1,232 @@
+use constant AS_PREFIX => "/_matrix/app/unstable";
+
+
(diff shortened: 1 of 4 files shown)


tests/60app-services/06publicroomlist.pl
```diff
@@ -0,0 +1,238 @@
+use Future::Utils qw( try_repeat_until_success );
+
+
+use constant AS_PREFIX => "/_matrix/app/unstable";
+
+
+sub matrix_get_room_list_synced
```
NegativeMjark: If you aren't going to export this, then maybe don't prefix it with ``matrix_``? I'm somewhat ambivalent on this though.

PR #329, opened 2016-12-08
Add test for /joined_rooms and /joined_members

erikjohnston:


leonerd:
Might be worth trying to combine this with #332

leonerd:
I've moved the code over to #332 and merged the logic with my other ones.

PR #330, opened 2016-12-12
Make E2E key tests use specced endpoints

richvdh:
https://github.com/matrix-org/synapse/pull/1694 removes the unspecced
endpoints, so don't try to use them.


leonerd:
LGTM, though I wonder why one of the four tests failed on Jenkins...

richvdh:
Looks like it failed to talk to github.

PR #333, opened 2016-12-14
Add retry to test because dendron

erikjohnston:


tests/50federation/37public-rooms.pl
```diff
@@ -1,9 +1,12 @@
+use Future::Utils qw( try_repeat_until_success );
+
+
 test "Inbound federation can get public room list",
    requires => [ $main::OUTBOUND_CLIENT, $main::HOMESERVER_INFO[0],
                  local_user_and_room_fixtures(),
                  federation_user_id_fixture() ],
 
-   do => sub {
+  do => sub {
```
leonerd: Indenting got disturbed on this line

leonerd:
What's the expected latency to success in the normal case? I.e. I'm wondering if some sort of delay before retry might be wanting...

tests/50federation/37public-rooms.pl
erikjohnston: Ugh, yeah, the indentation of this file was all over the place.

erikjohnston:
It should be fairly instant as we're waiting for a single replication round trip.

erikjohnston:
It might be nice to have a `try_repeat_until_success` do some backoff, e.g. try 3 times, then add some delays?

leonerd:
Possibly; though the abstract `Future::Utils` has no source of timing, so it wouldn't be able to do that. Such ability would have to be in some more concrete wrapping that has access to the `$loop`.

Either way; this PR LGTM

PR #337, opened 2017-01-24
Test ability for room members to override displayname on a per-room basis

leonerd:
This tests the synapse code added by https://github.com/matrix-org/synapse/pull/1852


NegativeMjark:
LGTM

PR #340, opened 2017-02-20
Add extra push test

erikjohnston:


richvdh:
what's going on here? lgtm, will merge unless anyone shouts

PR #341, opened 2017-02-20
Test /context/ visibility

erikjohnston:


NegativeMjark:
LGTM

NegativeMjark:
(Probably would have races if we moved the context API to another process. But I don't really mind here)

PR #342, opened 2017-02-28
Unconditionalise synapse components under dendron

leonerd:
Having each component be optional will make the `haproxy` config much harder to generate. Turns out we don't really make use of this conditional feature much, so make the code simpler by just starting all the components.

Also fixes the bug that the main synapse wasn't passed the `notify_appservices` argument before, so the appservice worker wasn't happy to start up.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -204,10 +204,10 @@ sub start
 
         # If we're using dendron-style split workers, we need to disable these
         # things in the main process
-        "start_pushers" => (not $self->{dendron}),
-        "notify_appservices" => (not $self->{dendron}),
-        "send_federation" => (not $self->{dendron}),
-        "notify_appservices" => (not $self->{dendron}),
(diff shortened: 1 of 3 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -202,11 +201,13 @@ sub start
...
 
-        "notify_appservices" => (not $self->{appservice}),
-
-        "send_federation" => (not $self->{federation_sender}),
+        # If we're using dendron-style split workers, we need to disable these
+        # things in the main process
+        "start_pushers"      => ( not $self->{dendron} ),
+        "notify_appservices" => ( not $self->{dendron} ),
+        "send_federation"    => ( not $self->{dendron} ),
+        "notify_appservices" => ( not $self->{dendron} ),
```
NegativeMjark: You seem to have `"notify_appservices" => ( not $self->{dendron} ),` in there twice.

leonerd: That's because it's twice as important... ;)

Fixed.

PR #344, opened 2017-02-28
Avoid some warnings in 21presence-events.pl

leonerd:
Some small fixes:

 * `next`ing out of a `sub` is generally bad; `return` instead
 * Avoid warning of undef in `eq` operator by explicitly setting empty string if undefined


erikjohnston:
lgtm

PR #345, opened 2017-03-06
Use a map file for path routing in haproxy

erikjohnston:
After discussing the various options with Mark, I decided that actually using a map file to specify the routing isn't the worst idea in the world as it makes it easier to share the config between production and sytest, as well as between different frontends.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -874,11 +874,11 @@ EOCONFIG
 
 sub generate_haproxy_map
 {
-    return <<"EOCONFIG";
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -896,6 +872,27 @@ backend client_reader
 EOCONFIG
 }
 
+sub generate_haproxy_map
+{
+    return <<"EOCONFIG";
```
leonerd: Since there's no perl-level variable interpolation in here, you can use `'EOCONFIG'` to be an q()'ed string so $ is not special there.

leonerd:
LGTM if Jenkins is happy

PR #347, opened 2017-03-14
Add a test to check for routing loops in the media repo

richvdh:


lib/SyTest/TCPProxy.pm
```diff
@@ -0,0 +1,91 @@
...
+               $outgoing->write( $$buffref );
+               $$buffref = "";
+               return 0;
+            },
+            on_closed => sub {
+               $outgoing->close();
+            },
+         );
+         
+         $self->loop->add( $incoming );
```
leonerd: Tiny stylistic point on IO::Async - generally notifiers get built into trees, rather than all flatly added to the toplevel loop instance. To do that use

    $self->add_child( $incoming );

instead, so the two Streams are children of the Listener, which has minor effects on error logging and debugging and so on

leonerd:
Semantically, the direction of this code looks fine, aside the minor comment about adding child notifiers.

Syntactically, don't forget to check `STYLE-GUIDE`; in particular here there's lots of parens lacking inner whitespace - `( $foo, $bar )` rather than `($foo, $bar)`. Also the fat-arrows used to make named arguments want lining up with whitespace before.

As a side note, I might sometime consider extracting bits of code example from here and elsewhere and make a proper `IO::AsyncX::TCPProxy` module on CPAN, but that's a separate step for me at some point

richvdh:
PTAL. I *hope* I have fixed the coding style issues, but I seem to be physically incapable of spotting some of them, so apologies for any I have missed.

tests/51media/01unicode.pl
```diff
@@ -44,11 +75,15 @@
 
 sub test_using_client
 {
-   my ( $client ) = @_;
+   my ( $client, $content ) = @_;
+
+   if( ! defined( $content )) {
```
leonerd: I'd be tempted to write

    unless( defined $content )

but this is fine too

leonerd:
LGTM

PR #349, opened 2017-03-24
Test guests can/can't join over federation

dbkr:
Test that they can join if guest access is allowed (expected fail) and cannot join when guest access is prohibited.

### DIFF
--- a/tests/30rooms/13guestaccess.pl
+++ b/tests/30rooms/13guestaccess.pl
@@ -485,6 +485,64 @@ test "GET /publicRooms includes avatar URLs",
       });
    };
 
+test "Guest users can accept invites to private rooms over federation",
+   requires => [ remote_user_fixture(), guest_user_fixture() ],
+
+   do => sub {
+      my ( $remote_user, $local_guest ) = @_;
+
+      my ( $room_id );
+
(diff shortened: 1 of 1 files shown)


tests/30rooms/13guestaccess.pl
```diff
@@ -485,6 +485,66 @@
...
+            content   => {
+               join_rule => "invite",
+            }
+         )
+      })->then( sub {
+         matrix_set_room_guest_access( $remote_user, $room_id, "forbidden" )
+      })->then( sub {
+         matrix_invite_user_to_room( $remote_user, $local_guest, $room_id )
+      })->then( sub {
+         matrix_join_room( $local_guest, $room_id )->main::expect_http_403
```
leonerd: On a purely stylistic note, I think it's becoming usual in tests to wrap the `->main::expect_http_403` onto a new line anyway, even though it does fit here, just to draw more attention to it. Its presence entirely inverts the expected logic here so it's quite important to see it.

    matrix_join_room( ... )
       ->main::expect_http_403

leonerd:
LGTM, aside from wanting one attention-grabbing linefeed.

PR #350, opened 2017-04-01
fix tests for https://github.com/matrix-org/synapse/pull/2090

ara4n:


leonerd:
LGTM

richvdh:
Again I've rewritten much of this so a re-review would be appreciated. It looks somewhat more complicated than it is. In a nutshell:

* factor out the common parts of some of the tests in 01unicode, 02nofilename, 03ascii into shared functions `upload_test_content`, `get_media`.
* replace two "Can download with ASCII file name..." tests with a loop which downloads the file from both places, and tests a bunch of other filenames besides.
* remove a couple of redundant calls to `test_using_client`.

tests/51media/03ascii.pl
```diff
@@ -2,80 +2,58 @@
...
-         $content_uri = $body->{content_uri};
+my %TEST_CASES = (
+   "ascii" => "inline; filename=ascii",
 
-         my $parsed_uri = URI->new( $body->{content_uri} );
-         my $server = $parsed_uri->authority;
-         my $path = $parsed_uri->path;
+   # synapse uses filename* encoding for this, though regular filename encoding
+   # with a quoted string would be valid too
+   "name with spaces" => "inline; filename*=utf-8''name%20with%20spaces",
```
erikjohnston: It would be nice if this was a map of file names to a list of valid content dispositions, so we don't force synapse's choice on other projects

richvdh: that's not quite as easy as it sounds. have made it parse the header.

PR #351, opened 2017-04-04
Use new replication tcp API

erikjohnston:


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -319,6 +319,15 @@ sub generate_listeners
       }
    }
 
+   if( my $replication_tcp_port = $self->{ports}{synapse_replication_tcp} ) {
+      push @listeners, {
+         type => "replication",
```
leonerd: Don't forget to line these up nicely

leonerd:
Semantically LGTM. Just needs one bit of whitespace fix.

leonerd:
LGTM

PR #352, opened 2017-04-20
Add a test for rejecting remote invites several times

richvdh:
Test for https://github.com/matrix-org/synapse/issues/1987. Passes under
synapse develop, fails on 0.19.5.

### DIFF
--- a/tests/30rooms/06invite.pl
+++ b/tests/30rooms/06invite.pl
@@ -120,6 +120,29 @@ test "Invited user can reject invite over federation",
    ],
    do => \&invited_user_can_reject_invite;
 
+test "Invited user can reject invite over federation several times",
+   # https://github.com/matrix-org/synapse/issues/1987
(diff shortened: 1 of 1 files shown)


tests/30rooms/06invite.pl
```diff
@@ -120,6 +120,29 @@ sub inviteonly_room_fixture
...
 
+test "Invited user can reject invite over federation several times",
+   # https://github.com/matrix-org/synapse/issues/1987
+   requires => [ remote_user_fixture(),
+      do {
+         my $creator = local_user_fixture();
+         $creator, inviteonly_room_fixture( creator => $creator );
+      }
+   ],
+  do => sub {
```
leonerd: Indenting seems amiss here.

leonerd:
Seems a reasonable idea since the test function already exists.

Semantically LGTM, needs indent adjustment

PR #354, opened 2017-05-03
Handle eventual consistency in public room tests

erikjohnston:
... I just wrapped the check part in a `try_repeat_until_success( ... )` loop...


leonerd:
LGTM; sorry for slowness.

erikjohnston:
ta, np

PR #355, opened 2017-05-23
Test that enabling disabling push works

erikjohnston:


leonerd:
It took me a little while to work out what the method of this test is; if I read it right, the idea is to set up push, send a message "1", then disable push, send "2" then enable and send "3", and assert that only messages 1 and 3 are received by push? If this is correct, can you put a little note or two of comment in there somewhere to explain this part?

If that is the case, this feels a somewhat racy and fragile way to perform the test. That said, offhand I can't think of a different way to assert that a push hasn't happened. :/

erikjohnston:
@leonerd I've added a comment that explains whats going on. It is icky, but I can't really think of a better way :(

leonerd:
Yeah that's fair enough. If that's how the test has to be done then it just has to be that way. The comment helps a lot to explain why.

LGTM

PR #357, opened 2017-06-07
Add sytest_template to speed up runs with PG

erikjohnston:


lib/SyTest/Homeserver.pm
```diff
@@ -82,14 +82,25 @@ sub clear_db_pg
...
+   # If there is a DB called sytest_template use that as the template for the
+   # sytest databases. Otherwise initialise the DB from scratch (which can take
+   # a fair few seconds)
+   my $dbh = DBI->connect( "dbi:Pg:dbname=sytest_template;host=$host", $args{user}, $args{password} );
+   if ( $dbh ) {
+      $dbh->do( "DROP DATABASE $args{database}" );  # we don't mind if this dies
+
+      $dbh->do( "CREATE DATABASE $args{database} WITH TEMPLATE sytest_template" ) or
+         die $dbh->errstr;
+   } else {
```
leonerd: Style guide - uncuddle the `}`:

    }
    else {

lib/SyTest/Homeserver.pm
```diff
@@ -82,14 +82,25 @@ sub clear_db_pg
...
+      my $dbh = DBI->connect( "dbi:Pg:dbname=$args{database};host=$host", $args{user}, $args{password} )
```
leonerd: You probably didn't want to make a new `$dbh` in the body of this `else` block, or else it won't be visible to code below. I suggest remove the `my`.

leonerd:
Will there be a sample schema file or instructions for people to create their own `sytest_template` database somehow?

erikjohnston:
> Will there be a sample schema file or instructions for people to create their own sytest_template database somehow?

Yeah, good idea. Where should I put it? README?

leonerd:
Yeah README should be sufficient. It'll be a one-time optional setup thing, it doesn't have to be too prominent.

leonerd:
LGTM

PR #364, opened 2017-06-29
Neater repeat loops that fail faster

leonerd:
* Neatens up the `try_repeat_until_success` code with delays into a self-contained `retry_until_success`, which applies the delay internally.

* Adds a new utility function, `repeat_until_true`, which repeats a Future<boolean>-returning block of code until its result is true. An outright exceptional failure is not caught and retried (as it would be with `retry_until_success`) but is instead immediately thrown upwards.

The distinct naming of these two is intentional - anything with the word `try` or `retry` in its name implies that it has a failure-catching semantic to it that will somehow handle failures itself; anything without does not and will propagate like normal code.

* Adds a new future following method, `main::check_http_code`, for evaluating a 3-way case split on HTTP results between OK, retry (via the above), or outright failure.

* Adjusts many callsites within the body of the test suite to use these new functions.

Overall this has the effect that temporary "It's not ready yet" failures in tests will be repeated, but permanent "500 the server died"-style failures will stop immediately with a more obvious message, rather than waiting for the 10 second generic timeout due to infinite retry.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -14,7 +14,7 @@ use SyTest::Assertions qw( :all );
 use SyTest::JSONSensible;
 
 use Future;
-use Future::Utils qw( try_repeat );
+use Future::Utils qw( try_repeat repeat );
 use IO::Async::Loop;
 
 use Data::Dump qw( pp );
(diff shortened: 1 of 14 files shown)


erikjohnston:
LGTM! Thanks :)

PR #369, opened 2017-08-22
Die more neatly if synapse fails to start

richvdh:
Try to avoid waiting for something that will never happen if synapse never
starts.


tests/05synapse.pl
```diff
@@ -28,11 +28,15 @@ sub extract_extra_args
 # see
 #   https://rt.perl.org/Public/Bug/Display.html?id=128774
 main::AT_END sub {
-   $OUTPUT->diag( "Killing synapse servers " ) if @synapses;
 
    ( fmap_void {
       my $synapse = $_;
 
+      # skip this if the process never got started.
+      return Future->done unless $synapse->{proc};
```
leonerd: That's a little bit of gut-wrenching into the innards of the `$synapse` object there. I'd use the `->pid` accessor method that you've already fixed to return zero if there isn't a process:

    return Future->done unless $synapse->pid;

richvdh: seems fair. done in dee61df

tests/05synapse.pl
```diff
@@ -28,11 +28,15 @@ sub extract_extra_args
 # see
 #   https://rt.perl.org/Public/Bug/Display.html?id=128774
 main::AT_END sub {
-   $OUTPUT->diag( "Killing synapse servers " ) if @synapses;
 
    ( fmap_void {
       my $synapse = $_;
 
+      # skip this if the process never got started.
+      return Future->done unless $synapse->{pid};
```
leonerd: Surely `$synapse->pid`. It's a method not an internal hash slot access.

richvdh: oh bugger

leonerd:
Much better. LGTM

PR #370, opened 2017-08-23
Install List::MoreUtils::XS

richvdh:


leonerd:
Ugh :/ More MoreUtils mess. I wonder what we're even using this for by now, and whether it can just be excised...

richvdh:
my investigation into the history of the project does not inspire confidence :/

richvdh:
Looks like it is used for exactly one function call. Will have a go at excising it.

leonerd:
https://github.com/matrix-org/sytest/pull/373

richvdh:
yay that looks better

PR #371, opened 2017-08-23
Set bcrypt_rounds = 4

richvdh:
to stop breakage occuring with modern python bcrypt


richvdh:
*FINALLY* a clean build. I actually only opened this PR to see if I could make it work. I'll make a new one with just bcrypt_rounds.

PR #372, opened 2017-08-23
Set bcrypt_rounds = 4

richvdh:
to stop breakage occuring with modern python bcrypt.

(We now use bcrypt 3 in synapse, as of
https://github.com/matrix-org/synapse/pull/2288)


richvdh:
(this will fail because of the List::MoreUtils mess)

leonerd:
Code itself LGTM; will let you deal with Jenkins / other complaints.

richvdh:
well, it didn't fail, because Jenkins isn't starting with a clean PERL5LIB.

PR #373, opened 2017-08-23
Use uniq() from List::Util 1.45 rather than List::MoreUtils

leonerd:
Because List::MoreUtils is turning out to be ever-harder to install.

See also https://github.com/matrix-org/sytest/pull/370 and https://rt.cpan.org/Ticket/Display.html?id=122875


leonerd:
Hum. Jenkins appears most unhappy but it doesn't look to be related to this particular change.

richvdh:
Jenkins is unhappy about the thing in #372.

richvdh:
@matrixbot retest this please

PR #374, opened 2017-08-25
Fixes to how we handle failures in test fixtures

richvdh:
A couple of fixes to make sure that things carry on when fixtures fail

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -364,6 +364,21 @@ sub repeat_until_true(&)
    }  until => sub { $_[0]->get };
 }
 
+# wrapper around Future::with_cancel which works around a bug whereby the
+# returned future does not keep a reference to the old future, which means it
+# may get garbage-collected.
+# (Ref: https://rt.cpan.org/Ticket/Display.html?id=122920)
+#
+# (also sets the label as a potential debugging aid)
+sub without_cancel
+{
+   my ( $future ) = @_;
+   my $new = $future->without_cancel;
+   $new->{oldref} = $future;
+   $new->set_label( "without_cancel(" . ( $future->label // $future ) . ")" );
+   return $new;
+}
+
 my @log_if_fail_lines;
 my $test_start_time;
 
@@ -429,7 +444,7 @@ sub fixture
 
       sub { $f_start->done( @_ ) unless $f_start->is_ready },
 
-      Future->needs_all( @req_futures )
+      Future->needs_all( map { without_cancel($_) } @req_futures )
          ->then( $setup )
          ->set_label( $name ),
 
@@ -541,7 +556,7 @@ sub _run_test
 
    my $success = eval {
       my @reqs;
-      my $f_setup = Future->needs_all( @req_futures )
+      my $f_setup = Future->needs_all( map { without_cancel($_) } @req_futures )
          ->on_done( sub { @reqs = @_ } )
          ->on_fail( sub { die "fixture failed - $_[0]\n" } );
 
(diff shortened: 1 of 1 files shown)


run-tests.pl
```diff
@@ -359,6 +359,20 @@ (&)
    }  until => sub { $_[0]->get };
 }
 
+# wrapper around Future::with_cancel which works around an apparent bug:
+#
+# the returned future does not keep a reference to the old future, which means
+# it may get garbage-collected.
```
leonerd: Hm. That sounds like a potential Future bug.

Reported https://rt.cpan.org/Ticket/Display.html?id=122920

run-tests.pl
```diff
@@ -359,6 +359,20 @@ (&)
...
+#
+# (also sets the label as a potential debugging aid)
+sub without_cancel {
+   my ($future) = @_;
+   my $new = $future->without_cancel;
+   $new->{oldref} = $future;
+   $new->set_label( "without_cancel(".( $future->label // $future ).")" );
+   return $new;
+}
```
leonerd: Semantic logic here all seems fine (give or take the somewhat gut-wrenchy poking of `{oldref}` but there's not a lot else that could be done here).

Syntax style doesn't match the style guide though. Opening `{` on its own line; whitespace inside `( $future )` and around the `.` operators.

run-tests.pl
```diff
@@ -527,9 +541,9 @@ sub _run_test
 
    my $success = eval {
       my @reqs;
-      my $f_setup = Future->needs_all( @req_futures )
+      my $f_setup = Future->needs_all( map { without_cancel($_) } @req_futures )
          ->on_done( sub { @reqs = @_ } )
-         ->on_fail( sub { die "fixture failed - $_[0]\n" } );
+         ->else( sub { die "fixture failed - $_[0]\n" } );
```
leonerd: This wants to stay as `on_fail` - it's not a piece of logic inserted *inline* into the failure-handling path, simply an incidental observer of the fact that failure happened.

richvdh: I'm not sure I follow. If it's an incidental observer, why do we want to `die` inside it? 

I thought the point of this code was that if any of the `@req_futures` fail, we want `$f_setup` to fail with a specific error message - so we *do* want to insert it into the failure-handling path.

Empirically, without this change, the test run just wedges (IIRC think it times out after 60s or something). The code in `on_fail` isn't `eval`ed, so the `die` ends up taking out lots of layers of the stack which seem to upset Future.pm. (It took me quite a while to track this down...)

run-tests.pl
richvdh: thanks. added to comment. 7997669

run-tests.pl
richvdh: Yup sorry. Fixed in 7997669

run-tests.pl
richvdh: Having poked at this a bit harder, I think I see what you mean: I think you are hoping that the `die` will cause us to leave the `eval` block at https://github.com/matrix-org/sytest/blob/6c464959c614ac92d3055fa70b385a15e19d45c6/run-tests.pl#L542.

Unfortunately, the `die` is happening within the context of the `on_fail` callback rather than directly in that eval block, which means it takes out lots of intermediate layers of stack which might still be important.

By turning this into an `else`, we instead make the `get` at https://github.com/matrix-org/sytest/blob/6c464959c614ac92d3055fa70b385a15e19d45c6/run-tests.pl#L586 throw, which has the desired effect.

richvdh:
@leonerd PTAL?

richvdh:
In the absence of further feedback from @leonerd, I'm going to take the executive decision to merge this and subsequent changes.

leonerd:
Retroactive LGTM

PR #375, opened 2017-08-25
Allow Fixtures to have a name, and use it for labelling futures

richvdh:
This is a useful debugging aid.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -371,11 +371,16 @@ sub log_if_fail
    push @log_if_fail_lines, split m/\n/, pp( $structure ) if @_ > 1;
 }
 
-struct Fixture => [qw( requires start result teardown )], predicate => "is_Fixture";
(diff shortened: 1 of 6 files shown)


run-tests.pl
```diff
@@ -371,12 +371,17 @@ sub log_if_fail
...
+struct Fixture => [qw( name requires start result teardown )], predicate => "is_Fixture";
 
+my $fixture_count = 0;
 sub fixture
 {
    my %args = @_;
 
+   # make up an id for later labelling etc
+   my $count = $fixture_count++;
+   my $name = $args{name} || "FIXTURE-$count";
```
leonerd: Generally best to use `//` which is the defined-or operator, in case someone did actually want to set something like an empty string or zero as a name

richvdh: indeed. 973c87c

leonerd:
LGTM

PR #376, opened 2017-08-29
Include unique ID in generated user IDs

richvdh:
This makes it easier to run repeated test runs against the same HS instance.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -25,6 +25,7 @@ use List::Util 1.33 qw( first all any maxstr max );
 use Struct::Dumb 0.04;
 use MIME::Base64 qw( decode_base64 );
 use Time::HiRes qw( time );
+use Date::Format qw( time2str );
 
 use Data::Dump::Filtered;
 Data::Dump::Filtered::add_dump_filter( sub {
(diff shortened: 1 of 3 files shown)


run-tests.pl
```diff
@@ -56,6 +57,10 @@
 
 our $BIND_HOST = "localhost";
 
+# a unique ID for this test run. It is used in some tests to create user IDs
+# and the like.
+our $TEST_RUN_ID = time2str( '%Y%m%d%H%M%S', time() );
```
leonerd: Since we don't directly depend on `Date::Format` and it isn't core, it might be better to stick to just the core `POSIX` instead.

    use POSIX qw( strftime );

    our $TEST_RUN_ID = strftime( '%Y%m%d%H%M%S', localtime() );

Also might be nice to give a nod to ISO8601 and put a `T` in there, so it's easier to visually inspect where the actual time part of that sits, since it would be more human-readable:

    '%Y%m%dT%H%M%S'

richvdh: both good points; done.

leonerd:
LGTM

PR #377, opened 2017-08-29
Fix 'Future is not yet complete' fail

richvdh:
If a Fixture is being torn down before it has run (for example because all other
fixtures in the same `requires` list have failed), it was possible for us to
try to get the result from an incomplete Future. This looks like a typo, so fix
it to wait for the right future.


leonerd:
Ahyes, good catch. Subtle logic there.

LGTM

PR #378, opened 2017-08-31
Provide a mechanism to let fixtures skip tests

richvdh:
In some situations, it turns out to make more sense to skip tests when certain fixtures fail than to fail them. Make it so that if a fixture fails with "SKIP: ...", then we skip any tests which require it rather than failing them.

(This PR includes 3 commits which it may be useful to review individually.)

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -205,7 +205,7 @@ require that fixture will be marked as failures. However, if the future fails
 with a reason beginning with the special string ``SKIP``, then any tests which
 require it are instead skipped. This can be used to differentiate between tests
 which we were unable to run due to constraints when setting up the test
-environment, as opposeed to those where setup steps that should have succeeded
+environment, as opposed to those where setup steps that should have succeeded
 did not.
 
 The intented use for fixtures is that test files will provide wrapper functions
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -200,6 +200,14 @@ main kinds of fixtures:
   over the lifetime of the test. These are fixtures that have a ``teardown``
   block.
 
+In general, if the future returned by ``setup`` fails, then any tests which
+require that fixture will be marked as failures. However, if the future fails
+with a reason beginning with the special string ``SKIP``, then any tests which
+require it are instead skipped. This can be used to differentiate between tests
+which we were unable to run due to constraints when setting up the test
+environment, as opposeed to those where setup steps that should have succeeded
+did not.
+
 The intented use for fixtures is that test files will provide wrapper functions
 that create a new fixture object to encapsulate some common setup pattern that
 later tests may require. Later tests can then simply invoke that function as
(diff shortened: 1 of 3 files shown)


run-tests.pl
```diff
@@ -525,9 +525,40 @@ sub _run_test
...
+   }
+   else {
+      $t->fail( $reason );
+   }
+}
+
+# Helper for _run_test. Waits for the req_futures to become ready, then runs
+# the test itself and waits for the test to complete.
+#
+# returns a pair [$res, $reason] where $res is one of:
```
leonerd: Best to document that with `(parens)` so the reader doesn't think it returns a two-element arrayref.

DEVELOP.rst
```diff
@@ -200,6 +200,14 @@ main kinds of fixtures:
   over the lifetime of the test. These are fixtures that have a ``teardown``
   block.
 
+In general, if the future returned by ``setup`` fails, then any tests which
+require that fixture will be marked as failures. However, if the future fails
+with a reason beginning with the special string ``SKIP``, then any tests which
+require it are instead skipped. This can be used to differentiate between tests
+which we were unable to run due to constraints when setting up the test
+environment, as opposeed to those where setup steps that should have succeeded
```
leonerd: opposed

run-tests.pl
```diff
@@ -549,17 +552,29 @@ sub _run_test
...
       my $f_setup = Future->needs_all( @$req_futures )
          ->on_done( sub { @reqs = @_ } )
-         ->on_fail( sub { die "fixture failed - $_[0]\n" } );
+         ->on_fail( sub {
+            my ( $reason ) = @_;
+
+            # if any of the fixtures failed with a special 'SKIP' result, then skip
+            # the test rather than failing it.
+            if ( $reason =~ /^SKIP(: *(.*))?$/ ) {
+               my $r = ": $2" // '';
```
leonerd: The string `": $2"` is never going to be undefined here even if `$2` was. You probably want to use ternary instead

    my $r = defined $2 ? ": $2" : "";

run-tests.pl
```diff
@@ -549,17 +552,29 @@ sub _run_test
...
+               $skip_reason = "failing fixture$r";
```
leonerd: Or maybe:

    $skip_reason = "failing fixture";
    $skip_reason .= ": $2" if defined $2;

run-tests.pl
richvdh: done in b8a39cb

DEVELOP.rst
richvdh: done in b8a39cb

run-tests.pl
richvdh: done in b8a39cb

richvdh:
@leonerd: PTAL

richvdh:
as per https://github.com/matrix-org/sytest/pull/374#issuecomment-330774846, merging this. The test failure seems unrelated.

leonerd:
Retroactive LGTM

PR #380, opened 2017-08-31
Factor out '_get_dbconfig' to Homeserver.pm

richvdh:
Extract a function which loads the database config, and clears the test database. This is going to be useful for Dendrite too.

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -99,13 +99,16 @@ sub _get_dbconfig
       if( defined $db_name ) {
          if( $db_name eq 'psycopg2' ) {
             $db_config{type} = 'pg';
-         } elsif( $db_name eq 'sqlite3' ) {
+         }
+         elsif( $db_name eq 'sqlite3' ) {
             $db_config{type} = 'sqlite';
-         } else {
+         }
+         else {
             die "Unrecognised DB name '$db_name' in $db_config_abs_path";
          }
       }
-   } else {
+   }
+   else {
       YAML::DumpFile( $db_config_abs_path, \%defaults );
       %db_config = %defaults;
    }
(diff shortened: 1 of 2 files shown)


lib/SyTest/Homeserver.pm
```diff
@@ -59,19 +63,101 @@ sub write_json_file
...
+   my ( %db_config );
+   if( -f $db_config_abs_path ) {
+      %db_config = %{ YAML::LoadFile( $db_config_abs_path ) };
+
+      # backwards-compatibility hacks
+      my $db_name = delete $db_config{name};
+      if( defined $db_name ) {
+         if( $db_name eq 'psycopg2' ) {
+            $db_config{type} = 'pg';
+         } elsif( $db_name eq 'sqlite3' ) {
```
leonerd: Small style - uncuddle the `elsif` onto its own line

    }
    elsif( COND ) {

lib/SyTest/Homeserver.pm
```diff
@@ -59,19 +63,101 @@ sub write_json_file
...
+   my ( %db_config ) = @_;
+
+   my $db_type = $db_config{type};
+   if( $db_type eq 'pg' ) {
+      foreach (qw( database host user password )) {
+         if( !$db_config{args}->{$_} ) {
+            die "Missing required database argument $_";
+         }
+      }
+   } elsif( $db_type eq 'sqlite' ) {
```
leonerd: Here too

lib/SyTest/Homeserver.pm
```diff
@@ -59,19 +63,101 @@ sub write_json_file
...
+   return %db_config;
+}
+
+sub _check_db_config
+{
+   my $self = shift;
+   my ( %db_config ) = @_;
+
+   my $db_type = $db_config{type};
+   if( $db_type eq 'pg' ) {
```
leonerd: This is probably fine for now, but if too many more database types start turning up, it might make sense to either split the check methods out similar to the clear ones, or to have an entire set of SyTest::Homeserver::Database subclasses to store the check and clear methods in.

richvdh: seems unlikely, but yes, noted.

lib/SyTest/Homeserver.pm
richvdh: fixed in f9c4876

lib/SyTest/Homeserver.pm
richvdh: fixed in f9c4876

leonerd:
LGTM

PR #381, opened 2017-08-31
Add --server-implementation option

richvdh:
Begin the process of making the HS impl more pluggable. Add a
--server-implementation option, which is then used to find a factory which
will instantiate the homeserver class.

### DIFF
--- a/lib/SyTest/HomeserverFactory.pm
+++ b/lib/SyTest/HomeserverFactory.pm
@@ -1,0 +1,39 @@
+# Copyright 2017 New Vector Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package SyTest::HomeserverFactory;
+
+# get the name of this implementation, by which it can be referenced with -I
+sub name
+{
+   my $cls = "" . shift;
+   $cls =~ s/^SyTest::HomeserverFactory:://;
+   return $cls;
+}
+
+sub new
+{
+   my $class = shift;
+   my %params = @_;
+
+   my $self = bless {}, $class;
+
+   $self->_init( \%params );
+
+   return $self;
+}
+
+sub _init {}
+
+1;
--- a/lib/SyTest/HomeserverFactory.pm
+++ b/lib/SyTest/HomeserverFactory.pm
@@ -14,10 +14,12 @@
 
 package SyTest::HomeserverFactory;
 
-# get the name of this implementation, by which it can be referenced with -I
+# get the name of this implementation, by which it can be referenced with -I.
+#
+# Intented to be called as a class method: $FACTORY_CLASS->name
 sub name
 {
-   my $cls = "" . shift;
+   my $cls = shift;
    $cls =~ s/^SyTest::HomeserverFactory:://;
    return $cls;
 }
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -38,9 +38,14 @@ use Module::Pluggable
    search_path => [ "SyTest::Output" ],
    require     => 1;
 
+use Module::Pluggable
+   sub_name    => "homeserver_factories",
+   search_path => [ "SyTest::HomeserverFactory" ],
+   require     => 1;
+
 # A number of commandline arguments exist simply for passing values through to
-# the way that synapse is started by tests/05synapse.pl. We'll collect them
-# all in one place for neatness
+# the way that the server is started by tests/05homeserver.pl. We'll collect
+# them all in one place for neatness
 our %SYNAPSE_ARGS = (
    directory  => "../synapse",
    python     => "python",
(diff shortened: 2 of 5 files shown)


lib/SyTest/HomeserverFactory.pm
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package SyTest::HomeserverFactory;
+
+# get the name of this implementation, by which it can be referenced with -I
+sub name
+{
+   my $cls = "" . shift;
```
leonerd: Stringifying the `$self` ref is probably not the best idea here. You should do that with

    my $class = ref( shift );

leonerd:
Overall there's no real point to having an entire "factory" set of shadow classes here. That's a common design pattern in some languages like Java, because classes and `new` are a bit weird and can't be passed around as true first-class values. In Perl you don't need a factory like this, because you can just call the `new` method on any class name you have sitting around as a string in a scalar.

    my $IMPL = "Synapse";  # or whatever, from commandline arg
    my $hs_class = "SyTest::Homeserver::$IMPL";

    my $hs = $hs_class->new( %args );

If you want a bit of a nicer error message for unrecognised impl names you could insert

    $hs_class->can( "new" ) or die "Unrecognised homeserver implemenation type $IMPL\n";

The `Module::Pluggable` loop is still useful for printing the possible choices in the help text though.

run-tests.pl
```diff
@@ -38,9 +38,14 @@
    search_path => [ "SyTest::Output" ],
    require     => 1;
 
+use Module::Pluggable
+   sub_name    => "homeserver_factories",
+   search_path => [ "SyTest::HomeserverFactory" ],
```
leonerd: I'd change this to

    sub_name  => "homeserver_classes",
    search_path => [ "SyTest::Homeserver" ],

run-tests.pl
```diff
@@ -128,10 +141,22 @@ sub usage
 {
    my ( $exitcode ) = @_;
 
-   print STDERR <<'EOF';
+   my @output_formats =
+      map { $_->FORMAT }
+      grep { $_->can( "FORMAT" ) } output_formats();
+
+   my @homeserver_implementations =
+      map { $_->name() } homeserver_factories();
```
leonerd: Perhaps rather than a `sub name` that generates it from the class name, each HS implementation class could be given a

    use constant NAME => "Synapse";

if the HS impl class names don't directly correspond.

run-tests.pl
```diff
@@ -170,14 +198,20 @@ sub usage
...
    exit $exitcode;
 }
 
 my $OUTPUT = first { $_->can( "FORMAT") and $_->FORMAT eq $OUTPUT_FORMAT } output_formats()
    or die "Unrecognised output format $OUTPUT_FORMAT\n";
 
+$SERVER_IMPL = 'Synapse' unless $SERVER_IMPL;
+my $hs_factory_class = first { $_->name() eq $SERVER_IMPL } homeserver_factories()
+   or die "Unrecognised server implementation $SERVER_IMPL\n";
+our $HS_FACTORY = $hs_factory_class -> new();
```
leonerd: Then down here you could simpler

    our $HS_CLASS = first { $_->NAME eq $SERVER_IMPL } homeserver_classes() or
       or die "Unrecognised server implementation $SERVER_IMPL\n";

richvdh:
> In Perl you don't need a factory like this, because you can just call the new method on any class name you have sitting around as a string in a scalar.

The main reason I introduced the factories was to give a place to store the parameters for the Homeserver instances before they are created. Apologies, I now realise how arbitrary it must look in this PR without the context of what I was trying to achieve.

6426c75 (which I haven't PRed yet) introduces a mechanism by which each HomeserverFactory can add a bunch of command-line arguments, and stores the values of those arguments in the HomeserverFactory instance. In particular note that I instantiate the factory during commandline-argument parsing.

I guess I *could* achieve the same effect with class methods and package-scope variables, but it felt kludgy.

WDYT?

leonerd:
Ahh, yes. I admit I've only given a brief skim over 6426c75, but that looks like indeed a nice way to implement those options and so on. In that case, yes you could put the options in the same homeserver classes, but it starts to lump together two separate concerns there, so having two classes keeps them split.

lib/SyTest/HomeserverFactory.pm
richvdh: it's being called as a class method, so `shift` actually gives the class name. So the forced-stringification is redundant. Removed in 3b6aecc.

run-tests.pl
```diff
@@ -39,9 +39,14 @@
...
+   sub_name    => "homeserver_factories",
+   search_path => [ "SyTest::HomeserverFactory" ],
+   require     => 1;
+
 # A number of commandline arguments exist simply for passing values through to
-# the way that synapse is started by tests/05synapse.pl. We'll collect them
-# all in one place for neatness
+# the way that the server is started by tests/05homeserver.pl. We'll collect
+# them all in one place for neatness
 our %SYNAPSE_ARGS = (
```
richvdh: this actually ends up going away altogether, so I haven't bothered renaming it.

richvdh:
@leonerd PTAL

run-tests.pl
```diff
@@ -175,14 +203,20 @@ sub usage
...
    exit $exitcode;
 }
 
 my $OUTPUT = first { $_->can( "FORMAT") and $_->FORMAT eq $OUTPUT_FORMAT } output_formats()
    or die "Unrecognised output format $OUTPUT_FORMAT\n";
 
+$SERVER_IMPL = 'Synapse' unless $SERVER_IMPL;
+my $hs_factory_class = first { $_->name() eq $SERVER_IMPL } homeserver_factories()
+   or die "Unrecognised server implementation $SERVER_IMPL\n";
+our $HS_FACTORY = $hs_factory_class -> new();
```
leonerd: Style issue: `->` doesn't want whitespace either side.

PR #382, opened 2017-08-31
Add diagnostics to `await_connectable`

richvdh:
- they are generally useful, so don't need to be limited to Synapse

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -138,7 +138,7 @@ sub await_connectable
       })
    } while => sub { !$_[0]->failure and !$_[0]->get };
 
-   $fut -> on_done( sub {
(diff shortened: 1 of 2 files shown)


lib/SyTest/Homeserver.pm
```diff
@@ -132,7 +136,13 @@ sub await_connectable
          $loop->delay_future( after => $delay )
               ->then_done(0);
       })
-   } while => sub { !$_[0]->failure and !$_[0]->get }
+   } while => sub { !$_[0]->failure and !$_[0]->get };
+
+   $fut -> on_done( sub {
```
leonerd: Unnecessary whitespace around `->` arrow operator.

leonerd:
LGTM besides style

PR #383, opened 2017-08-31
move port allocation into Synapse.pm

richvdh:
(Most of) this is fairly specific to synapse, so do it inside the
Synapse-specific module rather than the test code.


leonerd:
Retroactive LGTM

PR #385, opened 2017-09-18
Check that current state events are in timeline

erikjohnston:


tests/31sync/06state.pl
```diff
@@ -745,3 +745,158 @@ sub wait_for_event_in_room {
          Future->done(1);
       })
    };
+
+# Test to check that current state events appear in the timeline,
+# even if they were set during a period the user couldn't see
+test "Current state appears in timeline in private history",
```
richvdh: please could you add a reference to a github bug (preferably a synapse one, but an ios one would do)?

tests/31sync/06state.pl
```diff
@@ -745,3 +745,158 @@ sub wait_for_event_in_room {
...
+         my ( $body ) = @_;
+
+         my $room = $body->{rooms}{join}{$room_id};
+
+         # Check that we see invitee join event
+         any {
+            $_->{type} eq "m.room.member"
+            && $_->{state_key} eq $invitee->user_id
+            && $_->{content}{membership} eq "join"
+         } @{ $room->{timeline}{events} }
```
richvdh: not sure I'd expect this to appear in the timeline, but ymmv

tests/31sync/06state.pl
```diff
@@ -745,3 +745,158 @@ sub wait_for_event_in_room {
...
+            or die "No join for joined user";
+
+         Future->done( 1 );
+      })
+   };
```
richvdh: NL ?

tests/31sync/06state.pl
erikjohnston: It can't really be put in the "state" part, as that specifically is for state at the start of the timeline, which would rightly have the "invite" in. It does feel quite icky though :/

tests/31sync/06state.pl
erikjohnston: Sigh. This is what I get for switching editor from one that auto inserts new line at the end.

PR #386, opened 2017-09-20
Avoid setting --python from install_and_run.sh

richvdh:
- so that it can also be used for dendrite.

### DIFF
--- a/jenkins/install_and_run.sh
+++ b/jenkins/install_and_run.sh
@@ -27,9 +27,7 @@ export BIND_HOST
 # If running dendron then give it somewhere to write log files to
 mkdir -p var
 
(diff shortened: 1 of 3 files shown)


richvdh:
(There may be some old synapse branches which rely on this, but it should basically be ok since https://github.com/matrix-org/synapse/commit/6e67aaa7f249b196aa0288d713c8265c957cfbd5)

leonerd:
Probably LGTM, though I'd worry eversoslightly about that `export TOX_BIN` there. Does anything that `tox` potentially invokes rely on having that set?

richvdh:
Possibly, but we aren't running tox here...

PR #387, opened 2017-09-20
Fix some races in the device-lists tests

richvdh:
both matrix_put_e2e_keys and matrix_set_device_display_name cause a device
"changed" notification, so it was impossible to tell if we'd received both
notifications - which meant that sometimes we would carry on with the test and
the notification would arrive late.

Fix this either by waiting for the changed notification after the first call,
or by dropping the call to matrix_set_device_display_name.

Also factor out the 'sync until the notification appears' code.

### DIFF
--- a/tests/41end-to-end-keys/06-device-lists.pl
+++ b/tests/41end-to-end-keys/06-device-lists.pl
@@ -9,13 +9,34 @@ sub is_user_in_changed_list
           any { $_ eq $user->user_id } @{ $body->{device_lists}{changed} };
 }
 
-sub is_user_in_left_list
-{
-   my ( $user, $body ) = @_;
 
-   return $body->{device_lists} &&
-          $body->{device_lists}{left} &&
-          any { $_ eq $user->user_id } @{ $body->{device_lists}{left} };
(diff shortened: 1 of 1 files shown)


tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -9,13 +9,34 @@ sub is_user_in_changed_list
...
+      ->then( sub {
+         my ( $body ) = @_;
+
+         log_if_fail "$msg: body", $body;
+
+         my $res = $body->{device_lists} &&
+            $body->{device_lists}{$device_list} &&
+            any { $_ eq $wait_for_id } @{ $body->{device_lists}{$device_list} };
+
+         Future->done( $res && $body );
```
erikjohnston: Why the ` && $body`?

richvdh: to make it return the body, as per https://github.com/matrix-org/sytest/pull/387/files#diff-f4f7b7565a33f2c8f4cc0f2c24bb75f6R13

erikjohnston: oh yes thats how this language works

PR #391, opened 2017-09-22
Test publicity down the summary API

dbkr:
This adds to the publicity get/set test to assert that the summary
API publishes publicity the way it looks like it does.

This test currently fails.


erikjohnston:
What's the status of this?

@matrixbot  restest this please?

dbkr:
I think this was just made as a demonstration of something that looked incorrect in synapse's impl of the groups API. Given it's gone live now, we may not care about this anymore.

richvdh:
I guess we dont' care.

PR #393, opened 2017-10-04
Start migrating away from /events to /sync

erikjohnston:


tests/10apidoc/33room-members.pl
```diff
@@ -554,7 +523,7 @@ sub matrix_join_room_synced
       do => sub {
          matrix_join_room( $user, $room_id, %params );
       },
-      check => sub { exists $_[0]->{rooms}{join}{$room_id} },
+      check => sub { exists $_[0]->{rooms}{join}{$_[1]} },
```
richvdh: I'm *assuming* we're doing this so that we can join by alias. If so, could you s/room_id/room_id_or_alias/ elsewhere in this function?

tests/10apidoc/33room-members.pl
```diff
@@ -521,7 +490,7 @@ sub local_user_and_room_fixtures
 {
    my %args = @_;
 
-   my $user_fixture = local_user_fixture();
+   my $user_fixture = local_user_fixture( %args );
```
richvdh: why are we doing this?

richvdh: oh, because the next commit wants to pass things down to it

richvdh:
This seems to start well but get messy somewhere along the lines.

123bff3 is great, though it includes would have been nice to remove the spurious EXPORTs from `09synced` in the same commit. It would also be nice to give `sync_room_contains` some POD documenting what it does.

279a2b5 seems like a reasonable idea, but also feels like it's going to break *everything*.

c4c190d is nice

27bd5d6 seems generally good.

60afcf9 is WAT. It doesn't seem to correlate to its commit comment in any way. 

Could I suggest we pull out 123bff3, c4c190d and 27bd5d6 into a separate PR and see where we stand?

erikjohnston:
I've pulled them into a separate PR

erikjohnston:
Ha, I seem to have messed up my commits in 60afcf9. Sigh.

PR #394, opened 2017-10-06
Start migrating away from /events to /sync, take 2.

erikjohnston:
Supersedes: https://github.com/matrix-org/sytest/pull/393


tests/10apidoc/33room-members.pl
```diff
@@ -554,7 +523,7 @@ sub matrix_join_room_synced
       do => sub {
          matrix_join_room( $user, $room_id, %params );
```
richvdh: again, can we s/room_id/room_id_or_alias/ in here

PR #397, opened 2017-10-12
Add diagnostics to test runner

richvdh:
If we're called with -vv, print some diagnostics about running tests and
fixtures, to help with tracking down mysterious fails.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -448,21 +448,37 @@ sub fixture
    # actually invoke $setup
    @req_futures or push @req_futures, $f_start;
 
+   my $start = sub {
+      return if $f_start->is_ready;   # already started
+      $OUTPUT->diag( "Starting fixture $name" ) if ( $VERBOSE >= 2 );
+      $f_start->done( @_ );
+   };
+
    return Fixture(
       $name,
 
       \@requires,
 
-      sub { $f_start->done( @_ ) unless $f_start->is_ready },
+      $start,
 
       Future->needs_all( map { without_cancel($_) } @req_futures )
-         ->then( $setup )
-         ->set_label( $name ),
+         ->on_fail( sub {
+            my ( $reason ) = @_;
+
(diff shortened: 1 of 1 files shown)


run-tests.pl
```diff
@@ -448,21 +448,37 @@ sub fixture
...
+      $start,
 
       Future->needs_all( map { without_cancel($_) } @req_futures )
-         ->then( $setup )
-         ->set_label( $name ),
+         ->on_fail( sub {
+            my ( $reason ) = @_;
+
+            warn( "$name: input fixture failed with $reason" )
+               if ( $VERBOSE >= 3 );
```
erikjohnston: I'm surprised we seem to be warning that things failed at a higher verbosity level than when it successfully finishes.

run-tests.pl
```diff
@@ -448,21 +448,37 @@ sub fixture
...
+         })->then( $setup )->on_done( sub {
```
erikjohnston: Nit: I think we prefer to have one of these per line, to make it easier to follow the chain of whats going on

run-tests.pl
erikjohnston: Also, would it be better to have it after `->then( $setup )` in case that fails? Or are we also logging on failure further down?

richvdh: yeah, the reason for both (the fact it's only logged at high verbosity, and that `$setup` failures aren't logged here) is that both situations will make this fixture fail, which will be logged at https://github.com/matrix-org/sytest/pull/397/files/88138b6705dcbf70a484ce8b3d6c5886e6535280#diff-2bd42e0e657c22140d1cc8bf57bbad95R628.

run-tests.pl
richvdh: done, though unconvinced it's much clearer

richvdh:
@erikjohnston PTAL

PR #405, opened 2017-10-20
Use URI->path_segments instead of ->path

leonerd:
The

    $uri->path

method returns the raw string from the incoming HTTP request, complete with `%xx` encoding if present.

We should use the convenience utility

    $uri->path_components

Which splits the path on literal `/` bytes, and then `%xx`-decodes each of the components individually.


erikjohnston:
Ta!

PR #411, opened 2017-10-26
Test that ASes can get and set profiles for users they control

richvdh:


tests/60app-services/02ghost.pl
```diff
@@ -189,3 +189,106 @@
...
+      do_request_json_for(
+         $as_user,
+         method => "PUT",
+         uri    => "/r0/profile/$user_id/avatar_url",
+         params => {
+            user_id => $user_id,
+         },
+         content => { avatar_url => $avatar_url },
+      );
+   };
```
erikjohnston: TBH I prefer to merge check and do, and its what we do elsewhere, but *shrugs*

richvdh: It's certainly not what we do wherever I cargo-culted this from. The nice thing about the separate functions is that it checks that you're testing what you think you're testing by making sure that the check fails before the do, and passes after the do.

erikjohnston: Ah yes, I guess I meant "what we've done recently since we last changed our minds". I don't really have much against it when it makes sense

PR #414, opened 2017-11-01
Autoset default displaynames

ara4n:


ara4n:
for https://github.com/matrix-org/synapse/pull/2617

PR #418, opened 2017-11-08
Disable event stream by default

erikjohnston:


richvdh:
retest this please

tests/10apidoc/01register.pl
```diff
@@ -160,7 +160,7 @@ sub matrix_register_user
 
       my $f = Future->done;
 
-      if( $opts{with_events} // 1 ) {
+      if( $opts{with_events} // 0 ) {
```
richvdh: think `// 0` is redundant fwiw

richvdh:
why are we doing this?

erikjohnston:
- So that we don't need to implement a stub API on /dendrite
- Generally tries to avoid having magic happen behind the scenes
- Helps reduce where we use /events instead of /sync

tests/30rooms/04messages.pl
```diff
@@ -22,27 +22,33 @@
...
-               return unless $event->{room_id} eq $room_id;
-
                $content->{msgtype} eq $msgtype or
                   die "Expected msgtype as $msgtype";
                $content->{body} eq $msgbody or
                   die "Expected body as '$msgbody'";
-               $event->{user_id} eq $senduser->user_id or
-                  die "Expected sender user_id as ${\$senduser->user_id}\n";
+               $event->{sender} eq $senduser->user_id or
+                  die "Expected sender user_id as ${\$senduser->sender}\n";
```
richvdh: this should still be user_id

tests/30rooms/04messages.pl
```diff
@@ -22,27 +22,33 @@
    do => sub {
       my ( $senduser, $local_user, $room_id ) = @_;
 
-      matrix_send_room_message( $senduser, $room_id,
-         content => { msgtype => $msgtype, body => $msgbody },
+      Future->needs_all(
+         matrix_sync( $senduser ),
```
richvdh: why do we have these extra syncs now?

erikjohnston:
Yeah, its quite invasive, but I think worth it in the end :/

tests/30rooms/04messages.pl
erikjohnston: Turns out that this was because the test after it was relying on there being a sync token. That test was buggered anyway so I've changed it to not rely on that.

erikjohnston:
Right, the error looks to be a flakey test, I think https://github.com/matrix-org/sytest/commit/3885ee43d5c99817176ac00c1138e8b54a65f7d1 fixes it. Do you want that as a separate PR?

tests/30rooms/04messages.pl
```diff
@@ -52,30 +52,45 @@
...
-               next unless $event->{type} eq "m.room.message";
-               my $content = $event->{content};
+            # We expect the event to be in the initial sync
+            sync_timeline_contains( $body, $room_id, sub {
+               my ( $event ) = @_;
+
+               return unless $event->{type} eq "m.room.message";
+
+               return 1;
+            }) or die "Excpected event not in initial sync";
```
richvdh: expected

erikjohnston:
Thanks! Sowwy for the sheer invasiveness of this

PR #420, opened 2017-11-10
Update registration username tests

richvdh:
/register now allows uppercase usernames, and downcases them.

It should still reject special characters though.


richvdh:
(cf matrix-org/synapse#2662)

richvdh:
ptal at this too.

PR #421, opened 2017-11-15
Fix flakey sytest 80jira/syn606

erikjohnston:


tests/30rooms/30history-visibility.pl
```diff
@@ -576,8 +576,9 @@ sub ignore_presence_for
 
 sub await_event_not_history_visibility_or_presence_for
 {
-   my ( $user, $room_id, $allowed_users ) = @_;
+   my ( $user, $room_id, $allowed_users, %params ) = @_;
    await_event_for( $user,
+      %params,
```
richvdh: you've moved `%params` from the start of the arg list to the end in `GET_new_events_for`: let's be consistent and put it at the end.

tests/90jira/SYN-606.pl
```diff
@@ -27,20 +26,28 @@
...
          })->then( sub {
+            my ( $body ) = @_;
+
+            my $from_token = $body->{messages}{end};
+
             Future->needs_all(
                matrix_send_room_text_message( $user, $room_id1, body => "moose" ),
-               await_event_not_history_visibility_or_presence_for( $nonjoined_user, $room_id1, [] ),
+               await_event_not_history_visibility_or_presence_for( $nonjoined_user, $room_id1, [],
+                  from => $from_token,
```
richvdh: there's no explanation anywhere (PR, commit comment, a comment in code) as to what is going on here or why it is a good thing. Am I supposed to read your mind?

generally passing an explicit `from` token seems a bit odd here. I thought the whole point of the `await_event` functions was that they took this stuff away from you (and if you wanted to manage it yourself, you ought to be using lower-level methods).

If this is the right thing to e doing for some reason, please could you add a comment explaining why.

erikjohnston: Sorry, I've updated with a comment.

tests/90jira/SYN-606.pl
```diff
@@ -27,20 +26,37 @@
...
+
+            # We need to manually handle the from tokens here as the await_event*
+            # methods may otherwise reuse results from an /events call that did
+            # not include the specified room (due to the user not being joined to
+            # it). This could cause the event to not be found.
+            # I.e. there is a race where we send a message, the background /events
+            # stream streams past the message, and then /events stream triggered by
+            # await_event_* (which *does* include the room_id) starts streaming
+            # from *after* the message. Hence the event is neither in the cache
+            # nor in the liv event stream.
```
richvdh: liv

PR #423, opened 2017-11-22
Sign join event when sending over federation

erikjohnston:
We also remove some of the `prev_state` checks, as that field isn't used anymore.


richvdh:
you broke the tests

erikjohnston:
> We also remove some of the prev_state checks, as that field isn't used anymore.

Apparently synapse is unhappy if you don't supply it at least an empty list.

erikjohnston:
@matrixbot retest this please

PR #428, opened 2018-01-29
Drop sequences from postgres database

richvdh:
When we clear out a postgres database, we should make sure to drop any defined
sequences.

(This is needed to make the tests pass against
https://github.com/matrix-org/synapse/commit/df8246d, which introduces a
standalone SEQUENCE object).

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -341,6 +341,15 @@ sub _clear_db_pg
          $dbh->do( "DROP TABLE $tablename CASCADE" ) or
             die $dbh->errstr;
       }
+
+      foreach my $row ( @{ $dbh->selectall_arrayref(
+         "SELECT c.relname FROM pg_class c LEFT JOIN pg_namespace n ON n.oid = c.relnamespace WHERE relkind='S' AND n.nspname='public'"
+      ) } ) {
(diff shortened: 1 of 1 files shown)


leonerd:
LGTM from a perlish perspective, if the Postgres part is fine.

lib/SyTest/Homeserver.pm
```diff
@@ -341,6 +341,15 @@ sub _clear_db_pg
          $dbh->do( "DROP TABLE $tablename CASCADE" ) or
             die $dbh->errstr;
       }
+
+      foreach my $row ( @{ $dbh->selectall_arrayref(
+         "SELECT c.relname FROM pg_class c LEFT JOIN pg_namespace n ON n.oid = c.relnamespace WHERE relkind='S' AND n.nspname='public'"
+      ) } ) {
```
erikjohnston: If we're assuming we're in the public namespace then i wonder if we could just get away with doing a `DROP SCHEMA public CASCADE; CREATE SCHEMA public` or whatever.

But if this works I'm happy for us to do it like this for now.

richvdh: possibly, although I read that you could run into problems with which user ends up owning the schema if you do it that way.

PR #434, opened 2018-03-08
Tests for history purge transactional API

richvdh:
purge_history now has a transactional API, which we should use and test.

(cf https://github.com/matrix-org/synapse/pull/2962)

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -5,11 +5,20 @@ use Future::Utils qw( repeat );
 sub await_purge_complete {
    my ( $admin_user, $purge_id ) = @_;
 
+   my $delay = 0.1;
+
    return repeat( sub {
-      do_request_json_for( $admin_user,
-         method  => "GET",
-         uri     => "/r0/admin/purge_history_status/$purge_id",
+      my ( $prev_trial ) = @_;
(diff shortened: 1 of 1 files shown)


richvdh:
retest this please

tests/48admin.pl
```diff
@@ -1,5 +1,22 @@
...
+   return repeat( sub {
+      do_request_json_for( $admin_user,
+         method  => "GET",
+         uri     => "/r0/admin/purge_history_status/$purge_id",
+      )->then( sub {
+         my ($body) = @_;
+         assert_json_keys( $body, "status" );
+         Future->done( $body->{status} );
+      })
+   }, while => sub { $_[0]->get eq 'active' });
```
erikjohnston: I'd use either `repeat_until_true` or `retry_until_success` helper functions, as they have a small backoff

richvdh: well, I'm deliberately doing my own repeat so that I can distinguish between failure (request fails) and not ready yet (status == 'active').

I've added a backoff.

erikjohnston: Ah ok, I just assumed one of those could be used. Ah well

PR #438, opened 2018-04-26
check that module installation has succeeded

richvdh:


richvdh:
paging @leonerd in case this is of interest to you... IO::Async::SSL was failing its tests, but the failure was getting lost in a bunch of other spew.

richvdh:
well, merging this so that it doesn't get lost.

PR #439, opened 2018-04-27
Force installation of Alien::Base::ModuleBuild

richvdh:
Apparently 'Alien::Sodium' (which is a dependency of 'Crypt::NaCl::Sodium') won't install unless Alien::Base::ModuleBuild is already installed.

I don't know how this ever worked :/


richvdh:
another one cc @leonerd in case you have any better ideas here

PR #445, opened 2018-05-29
Full test for lazy loading members

ara4n:
Paired by with https://github.com/matrix-org/synapse/pull/3331


ara4n:
i'm going to assume this has been lgtm'd given #2970 was.

ara4n:
actually, i guess there's no harm in getting an explicit lgtm from @richvdh

PR #451, opened 2018-06-15
Move critical test flag to homeserver configuration file

anoadragon453:
The critical flag for a test is no longer set in the test itself, but in `lib/HomeserverFactory/<name>.pl`. This allows us to run the full testsuite against homeservers which are still ongoing in development, and makes viewing which tests are critical for each homeserver fairly convenient.

To manage this, an `id` was assigned to each test which could be referenced when specifying which tests were critical. The ids were generated by taking the test name, lowercasing, changing spaces to underscores, and removing the `$` from variable names. Test ids should be updated if the test name ever changes, and vice versa.

Also changed some documentation and added some preliminary information about homeserver-specific configuration file. Some further documentation on what is needed to add sytest compatibility for a completely new homeserver should also be added at some point.

### DIFF
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -102,6 +102,15 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
+sub get_critical_tests
+{
+   return (
+       'POST [^ ]+ register can create a user',
+       'POST /createRoom makes a public room',
+       'POST /rooms/:room_id/join can join a room',
+       'POST /rooms/:room_id/leave can leave a room'
+   );
+}
 
 package SyTest::HomeserverFactory::Synapse::ViaDendron;
 use base qw( SyTest::HomeserverFactory::Synapse );
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -105,10 +105,10 @@ sub create_server
 sub get_critical_tests
 {
    return (
-       'POST [^ ]+ register can create a user',
+       'POST \S+ register can create a user',
        'POST /createRoom makes a public room',
        'POST /rooms/:room_id/join can join a room',
-       'POST /rooms/:room_id/leave can leave a room'
+       'POST /rooms/:room_id/leave can leave a room',
    );
 }
 
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -50,15 +50,25 @@ is described in more detail in the following sections.
     Provides an ``ARRAY`` reference giving a list of named requirements and
     fixture objects.
 
-``critical``
-    If true and the test fails, the entire test run will bail out at this
-    point; no further tests will be attempted at all.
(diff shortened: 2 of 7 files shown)


lib/SyTest/HomeserverFactory/Synapse.pm
```diff
@@ -102,6 +102,15 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
+sub get_critical_tests
+{
+   return (
+       'POST [^ ]+ register can create a user',
```
ara4n: it's still more idiomatic to do \S (meaning "not whitespace"), but not a big deal.

lib/SyTest/HomeserverFactory/Synapse.pm
```diff
@@ -102,6 +102,15 @@ sub create_server
...
+       'POST /createRoom makes a public room',
+       'POST /rooms/:room_id/join can join a room',
+       'POST /rooms/:room_id/leave can leave a room'
```
ara4n: prolly nice to have trailing comma here

ara4n:
that feels way saner to me - thanks; lgtm :)

lib/SyTest/HomeserverFactory/Synapse.pm
anoadragon453: Oh, another thing I didn't know perl could do!

anoadragon453:
Fixed :)

anoadragon453:
@erikjohnston @ara4n thoughts on the point of a critical test?

anoadragon453:
Actually, I want to nerf the last comment and simply state that this is an update to the existing system, that was at one point agreed upon as being useful.

While we can discuss about whether it should be kept in the first place, that should be done in a separate discussion as it is out of scope for this PR.

Ideally I'd just like feedback on the changes presented within this PR, which @richvdh has already given a LGTM.

richvdh:
As I see it we have three choices here:

1: land this PR as is, so that different impls can have different ideas of what is "critical"; possibly, consider removing the "critical" feature later.
2: remove the critical feature now.
3: keep the critical feature as-is and instead tune the list of tests which are considered critical.

my problem with 1 is that the second part may never happen, and in the meantime we've added extra code to maintain and got this system which doesn't make a lot of sense in my eyes. I still don't understand why it is meaningful for a test to be considered critical when running against one HS and not when running against another.

If dendrite currently fails all of the critical tests, then I guess option 3 isn't very attractive either, though I wouldn't object massively to just disabling the critical flag on all the tests for now and reconsidering later.

Personally, removing the feature entirely still gets my vote.

anoadragon453:
The issue that prompted this whole thing is that we couldn't run a full test of dendrite with the critical tests enabled, as it would fail some stuff early on that other tests didn't actually rely on, so presumably they weren't really that critical in the first place.

In that case then, 2 would probably be my vote as well, as we already have test skipping based on failures covered.

babolivier:
Jumping on Dendrite-related sytest work here, @anoadragon453 @richvdh has there been further discussions on this issue other than the comments in this PR?
If not, given that I'd also vote for 2, what do you think about me closing this PR and opening a new one for getting rid of the critical feature?

PR #452, opened 2018-06-18
Allow setting a white/blacklist for test names

anoadragon453:
This allows a list of regexes for test names to be defined, where only test names matching that list will be run.

This might break tests that rely on other tests? Not quite sure yet.

The other concern is whether checking a list of regexes per test is too inefficient. It only will be if we end up with many, many regexes for a single homeserver, which may be the case if we want to specifically test lots of little things. I thought about having the regexes be checked against a folder or filename, instead of test name, but that may not be specific enough. If you think there's a better way to do it let me know.

Depends on #451

### DIFF
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -102,14 +102,6 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
-# Tests to run for Synapse defined by array of regexes
-sub get_tests
-{
-   return (
-       '.*',
-   );
-}
-
 package SyTest::HomeserverFactory::Synapse::ViaDendron;
 use base qw( SyTest::HomeserverFactory::Synapse );
 
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -102,6 +102,15 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
+# Tests to run for Synapse defined by array of regexes
+sub get_tests
+{
+   return (
+       '.*',
+   );
+}
+
+# Critical tests for Synapse defined by array of regexes
 sub get_critical_tests
 {
    return (
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -110,17 +110,6 @@ sub get_tests
    );
 }
 
-# Critical tests for Synapse defined by array of regexes
-sub get_critical_tests
-{
-   return (
-       'POST \S+ register can create a user',
-       'POST /createRoom makes a public room',
-       'POST /rooms/:room_id/join can join a room',
-       'POST /rooms/:room_id/leave can leave a room',
-   );
-}
-
 package SyTest::HomeserverFactory::Synapse::ViaDendron;
 use base qw( SyTest::HomeserverFactory::Synapse );
 
--- a/lib/SyTest/HomeserverFactory/Synapse.pm
+++ b/lib/SyTest/HomeserverFactory/Synapse.pm
@@ -102,6 +102,15 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
+sub get_critical_tests
+{
+   return (
+       'POST [^ ]+ register can create a user',
+       'POST /createRoom makes a public room',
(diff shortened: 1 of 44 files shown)


richvdh:
looks plausible to me, modulo resolution on #451

anoadragon453:
As we're starting to set up sytest in Dendrite's CI, we wanted to have a simple method of setting which tests should and shouldn't be run. Previously this PR mandated hard-coded lists of regexes on test names per homeserver which would tell sytest which tests to run. Other than regexes easily capturing unintended tests by mistake, this required making a PR to sytest every time Dendrite got a little bit better, as well as this implementation previously not supporting tests skipped because an earlier one was skipped.

This has all been addressed now. We now provide two new cli flags (`--test-blacklist-file` and `--test-whitelist-file`) that lets you set either a black or white list using a file that is simply a list of test names, one per line. The idea is that a test whitelist file can be put in Dendrite's repo, and new PRs can edit this file to add sytest tests that they believe their PR should allow Dendrite to pass. Travis will then run this using the flag and report the results.

run-tests.pl
```diff
@@ -125,6 +133,14 @@ sub usage
                                    ~~ @<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
                                       shift( @homeserver_implementations ) || ''
 
+   -W, --test-whitelist-file    - whitelist file containing test names to run
+                                  One per line. Cannot be used in conjunction
+                                  with --test-blacklist-file
+
+   -B, --test-blacklist-file    - blacklist file containing test names to run
```
richvdh: to *not* run, one assumes

run-tests.pl
```diff
@@ -185,6 +201,36 @@ sub usage
 Getopt::Long::Configure("no_passthrough");
 GetOptions($HS_FACTORY->get_options()) or usage(1);
 
+# Check if both options have been set
+if (length($BLACKLIST_FILE) and length($WHITELIST_FILE)) {
```
richvdh: you can just use truthiness here:

```suggestion
if( $BLACKLIST_FILE and $WHITELIST_FILE ) {
```

likewise below

run-tests.pl
```diff
@@ -185,6 +201,36 @@ sub usage
...
+   print STDERR "Not allowed to set both whitelist and blacklist options.\n";
```
richvdh: the perly way to do this is:

```suggestion
   die "Not allowed to set both whitelist and blacklist options.\n";
```

(and the exit is then redundant)

run-tests.pl
```diff
@@ -185,6 +201,36 @@ sub usage
...
+   exit 1
+}
+
+# Read in test blacklist rules if set
+my @TEST_BLACKLIST = ();
+if (length($BLACKLIST_FILE)) {
+   open(my $blacklist_data, "<", $BLACKLIST_FILE) or die "Couldn't open blacklist file for writing: $!\n";
```
richvdh: er, for reading

run-tests.pl
```diff
@@ -185,6 +201,36 @@ sub usage
...
+      chomp $test_name;
+      push @TEST_BLACKLIST, $test_name;
+   }
+   close $blacklist_data;
+}
+
+# Read in test whitelist rules if set
+my @TEST_WHITELIST = ();
+if (length($WHITELIST_FILE)) {
+   open(my $whitelist_data, "<", $WHITELIST_FILE) or die "Couldn't open whitelist file for writing: $!\n";
```
richvdh: reading

lib/SyTest/HomeserverFactory/Synapse.pm
```diff
@@ -102,7 +102,6 @@ sub create_server
    return $self->{impl}->new( %params );
 }
 
-
```
richvdh: ftr I think that the extra line between class definitions isn't a bad thing, but :man_shrugging:

run-tests.pl
```diff
@@ -794,6 +840,16 @@ sub list_symbols
 
    my $m = $test->multi ? "enter_multi_test" : "enter_test";
 
+   # Check if this test has been blocked by the blacklist. If so, mark as expected fail
+   if ( scalar( @TEST_BLACKLIST ) and grep { $test->name =~ /\Q$_\E\z/ } @TEST_BLACKLIST ) {
```
richvdh:
```suggestion
   if ( scalar( @TEST_BLACKLIST ) and grep $test->name, @TEST_BLACKLIST ) {
```

likewise below

PR #454, opened 2018-06-21
Add a note about running the perl scripts in Docker on windows

turt2live:
The pesky \r\n makes the shebang unhappy.


turt2live:
The instructions don't apply to the docker image provided anymore. This is superseded by https://github.com/matrix-org/sytest/pull/511

PR #455, opened 2018-06-21
Add tests for creating rooms with names, topics, and avatars

turt2live:
Creating rooms with avatars is under a proposal at the time of writing.

Spec: https://github.com/matrix-org/matrix-doc/pull/1326
Synapse: https://github.com/matrix-org/synapse/pull/3417

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -69,7 +69,6 @@ test "POST /createRoom makes a private room with invites",
 
          content => {
             visibility => "private",
-            # TODO: This doesn't actually appear in the API docs yet
             invite     => [ $invitee->user_id ],
          },
       )->then( sub {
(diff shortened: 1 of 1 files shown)


turt2live:
I've updated this to be just for creating rooms with names and topics. Although the proposal is to deprecate these parameters, it seems useful to test these functions while the fields still exist. A later PR should remove them in the event the spec completes the deprecation process.

richvdh:
As per https://github.com/matrix-org/synapse/pull/3421#pullrequestreview-142897713, I think the tests of guest_can_join are incorrect.

turt2live:
Have removed the tests - please take a look

tests/10apidoc/30room-create.pl
```diff
@@ -82,6 +81,88 @@
...
+         },
+      )->then( sub {
+         my ( $body ) = @_;
+
+         assert_json_keys( $body, qw( room_id ));
+         assert_json_nonempty_string( $body->{room_id} );
+
+         my ( $room_id ) = $body->{room_id};
+
+         do_request_json_for( $user,
```
richvdh: ok so the problem with this is that it will be flaky when run against a synapse with a worker config, because the client_reader may not have got the memo about the new room when we get here.

A more reliable way is to call `matrix_create_room_synced` and see what you get back in the /sync response. #481 will probably be helpful here.

turt2live:
I think this now does the right thing - PTAL. I have heavily abused copy/paste for this.

PR #456, opened 2018-06-22
Add tests for /_matrix/media/r0/config

Half-Shot:
Not sure how I will go about testing the value, or if it's necessary.

Tests matrix-org/synapse#3184

### DIFF
--- a/tests/51media/30config.pl
+++ b/tests/51media/30config.pl
@@ -1,0 +1,17 @@
+test "Can read configuration endpoint",
+   requires => [ $main::API_CLIENTS[0] ],
+
+   check => sub {
+        my ( $http ) = @_;
+        $http->do_request(
+            method   => "GET",
+            full_uri => "/_matrix/media/r0/config",
+        )->then( sub {
+            my ( $body, $response ) = @_;
+                
+            assert_json_keys( $body, qw( m.upload.size ) );
+            #TODO: We should probably check the size is correct.
+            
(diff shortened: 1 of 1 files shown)


richvdh:
if you made your branches have the same name, the sytests would run against the right branch of synapse...

tests/51media/30config.pl
```diff
@@ -0,0 +1,17 @@
...
+
+   check => sub {
+        my ( $http ) = @_;
+        $http->do_request(
+            method   => "GET",
+            full_uri => "/_matrix/media/r0/config",
+        )->then( sub {
+            my ( $body, $response ) = @_;
+                
+            assert_json_keys( $body, qw( m.upload.size ) );
```
richvdh: given that this is optional, I'm not sure it is correct to assert it here.

You could assert that, if it is present, it is an int of a plausible size?

Half-Shot:
>if you made your branches have the same name, the sytests would run against the right branch of synapse...

This really aught to be mentioned somewhere on the matrix-org/synapse README.

richvdh:
> aught

You keep using that word. I do not think it means what you think it means.

Anyway, if anywhere, it should be in CONTRIBUTING, but it's a hack that only works for people in the matrix-org github org, so I'm unsure it should be there.

Half-Shot:
Closing and reopening so sytest will apparently test against the correct thing.

PR #458, opened 2018-06-29
Update 'user' field in application service tests to 'username'

anoadragon453:
The spec changed a while ago to have `/r0/register` accept `username` as a field instead of `user`. Most of SyTest's tests got updated, but not the application service ones.

### DIFF
--- a/tests/60app-services/01as-create.pl
+++ b/tests/60app-services/01as-create.pl
@@ -13,7 +13,7 @@ test "AS can create a user",
          uri    => "/r0/register",
 
          content => {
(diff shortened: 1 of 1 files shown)


erikjohnston:
Can we add a test that checks that `user` still works please (with comments about the fact its deprecated)? We probably don't want to regress and break `user` from working for now, we should only remove that test once we remove the code in e.g. synapse. Things like dendrite can make the test optional.

richvdh:
looks like this has bitrotted somewhat. closing, since it's nearly two years old.

PR #460, opened 2018-07-02
Remove test for unused /createUser endpoint

anoadragon453:
A very, very long time ago there was an specific endpoint, `/createUser`, that application services could use to create users.

This has been superseded by AS's creating their users through the C-S API, and thus as far as @Half-Shot and I can tell, no AS's use the old endpoint any longer.

There is presumably no point in supporting this endpoint in Dendrite, and we'd at the same time like to have all AS sytest's passing. Thus the proposal of removing this test.

If removal is considered detrimental for any reason, we can propose to move this test to a "deprecated" tests file which we can simply ignore the results of for Dendrite's CI pipeline.

### DIFF
--- a/tests/60app-services/01as-create.pl
+++ b/tests/60app-services/01as-create.pl
@@ -79,46 +79,6 @@ test "AS can create a user via the legacy /v1 endpoint",
       });
    };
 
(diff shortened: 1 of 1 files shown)


richvdh:
[the jenkins failures look unrelated]

PR #461, opened 2018-07-02
[WIP] Dendrite: Print homeserver output by default

anoadragon453:
Print out the server logs during tests by default.

I don't know if this change is Dendrite-specific, as I don't have a Synapse sytest environment set up. If anyone could tell me if this change override's the `-S` option for Synapse, do let me know :)


anoadragon453:
Why did I want this.

PR #462, opened 2018-07-10
add test to ensure you can query devicelist of invited users.

ara4n:
also add known broken test (commented out to stop it running and slowing things down
using the new broken_test directive) to track matrix-org/synapse#3503 - that we
don't get device_list updates from invited users.


run-tests.pl
```diff
@@ -737,8 +737,9 @@ sub list_symbols
 
          no warnings 'once';
 
-         local *test       = sub { _push_test( $filename, 0, @_ ); };
-         local *multi_test = sub { _push_test( $filename, 1, @_ ); };
+         local *test        = sub { _push_test( $filename, 0, @_ ); };
+         local *multi_test  = sub { _push_test( $filename, 1, @_ ); };
+         local *broken_test = sub {};
```
richvdh: well, if we're going to do this, then let's document what it's for, both here and in DEVELOP.rst. But I'm a bit unconvinced - the `bug` param means we'll expect it to fail, so why isn't that good enough?

tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -233,6 +233,163 @@ sub sync_until_user_in_device_list
...
+
+      matrix_create_room( $user1 )->then( sub {
+         ( $room_id ) = @_;
+         matrix_put_e2e_keys( $user2 );
+      })->then( sub {
+         matrix_sync( $user1 );
+      })->then( sub {
+         matrix_invite_user_to_room( $user1, $user2, $room_id )
+      # })->then( sub {
+      #    # KNOWN BROKEN: (https://github.com/matrix-org/synapse/issues/3503)
```
richvdh: I don't really understand what you're doing here. If you just comment it out and run the query anyway, surely it is going to fail, because the update might not yet have propagated over federation?

run-tests.pl
richvdh: stick a timeout on sync_until_user_in_device_list if necessary.

tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -233,6 +233,163 @@ sub sync_until_user_in_device_list
...
+      #    # We'd expect to get told via the devicelist at this point that the
+      #    # invited user's devices have been added.
+      #    sync_until_user_in_device_list( $user1, $user2 );
+      })->then( sub {
+         # check whether device changes get pushed for invited users before they join
+         matrix_set_device_display_name( $user2, $user2->device_id, "test display name" ),
+      # })->then( sub {
+      #    # KNOWN BROKEN: we should get an update from the invited user when they
+      #    # update their devices, but we don't because of
+      #    # https://github.com/matrix-org/synapse/issues/3504
```
richvdh: I can't see 3504 getting fixed for quite a while, since it will require some significant changes to the way invites work. Building even a failing test around assuming it will work seems a bit pointless.

richvdh:
Revisiting this 10 months later, I'm even more confused about what it is doing.

AFAICT it is meant to be a repro test for https://github.com/matrix-org/synapse/issues/3503, however without a fix for that on the horizon, I don't think the open PR is adding much value. If, at some point in the future, we get around to addressing https://github.com/matrix-org/synapse/issues/3503, we can always resurrect this PR.

PR #464, opened 2018-07-16
Tests for erasure

richvdh:


richvdh:
the failure appears to be unrelated (matrix-org/synapse#3548)

PR #466, opened 2018-07-18
Test hs/upload limits

Half-Shot:
Changed branch name to apparently make sytest happy which means this follows on from #456.


Half-Shot:
Well, I must be doing something wrong because this is now the correct branch name, the ``full_uri`` is correct and I'm fairly certain that path comes back with a config.

richvdh:
@matrixbot retest this please

richvdh:
[this is supposed to be testing matrix-org/synapse#3184]

richvdh:
> Well, I must be doing something wrong because this is now the correct branch name, the full_uri is correct and I'm fairly certain that path comes back with a config.

the reason it didn't pick up your synapse branch is that the synapse branch is on a fork of the repo rather than the matrix.org one.

PR #468, opened 2018-07-20
add test for /context/ with lazy_load_members filter

ara4n:
tests https://github.com/matrix-org/synapse/pull/3567


tests/10apidoc/00prepare.pl
```diff
@@ -42,3 +42,56 @@ sub new_User
 
    return $user;
 }
+
+
+# assert that the given members are in the body of a sync response
```
richvdh: this feels out of place here. did it need to move out of 15lazy-members.pl?

tests/10apidoc/00prepare.pl
```diff
@@ -42,3 +42,56 @@ sub new_User
...
+   log_if_fail "Room", $room;
+
+   assert_json_keys( $room, qw( timeline state ephemeral ));
+
+   return assert_room_members_in_state( $room->{state}{events}, $member_ids );
+}
+
+
+# assert that the given members are present in a block of state events
+sub assert_room_members_in_state {
```
richvdh: somehow this name makes me think we're checking the state contains *at least* these members. Maybe `assert_state_room_members_matches` ?

richvdh: this also feels out of place. I'm not really keen on this file being used as a dumping ground for random assertion functions.

Maybe `10apidoc/09synced.pl` is a more appropriate home?

tests/10apidoc/00prepare.pl
```diff
@@ -42,3 +42,56 @@ sub new_User
...
+   my ( $events, $member_ids ) = @_;
+
+   log_if_fail "members:", $member_ids;
```
richvdh: s/"members:"/"expected members:"/?

richvdh:
\o/

PR #478, opened 2018-08-02
Check that federation endpoints check room id

richvdh:
Regression tests for https://github.com/matrix-org/synapse/pull/3641


richvdh:
(the merged tests have failed because the synapse branch is against 0.33, which lacks support for lazyloading etc)

PR #480, opened 2018-08-02
Fixes for Python 3

hawkowl:


tests/60app-services/01as-create.pl
```diff
@@ -26,7 +26,7 @@
       });
    };
 
-test "AS can create a user with inhibit_login",
+test "AS can create a user with an underscore",
```
richvdh: it feels odd to have this test here but leave the counter-case in SYN-738.pl. No objection to moving them, but let's move both.

PR #484, opened 2018-08-08
Dockerised SyTest

hawkowl:


docker/Dockerfile
```diff
@@ -1,40 +1,26 @@
...
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
 
-WORKDIR /root/sytest
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+
+# The dockerfile context, when ran by the buildscript, will actually be the
```
richvdh: "when run by"

docker/synapse_sytest.sh
```diff
@@ -0,0 +1,59 @@
...
+
+if [ -e "/test/run-tests.pl" ]
+then
+    echo "Using local sytests..."
+else
+    branch_name="$(git --git-dir=/src/.git symbolic-ref HEAD 2>/dev/null)" || branch_name="(unnamed branch)"
+
+    echo "Trying to get same-named sytest..."
+    wget -q https://github.com/matrix-org/sytest/archive/$branch_name.tar.gz -O sytest.tar.gz
+4
```
richvdh: 4? I raise you 5.

docker/README.md
```diff
@@ -1,40 +1,33 @@
...
+- matrixdotorg/sytest, a base container with SyTest dependencies installed
+- matrixdotorg/sytest-synapsepy2, a container which will run SyTest against Synapse on Python 2.7
+- matrixdotorg/sytest-synapsepy3, a container which will run SyTest against Synapse on Python 3.5
 
-Or other commands as per [the main SyTest
-documentation](https://github.com/matrix-org/sytest#running).
+## Using the Synapse containers
 
-Alternatively:
+Once pulled from Docker Hub, the container can be ran on a Synapse checkout:
```
richvdh: "can be run on"

docker/README.md
```diff
@@ -1,40 +1,33 @@
...
-docker run --rm -it sytest bash
-```
+Included currently is:
 
-And then at the shell prompt:
-
-```
-./run-tests.pl
-```
+- matrixdotorg/sytest, a base container with SyTest dependencies installed
```
richvdh: one thing that I didn't quite grok at first was that the docker images don't in fact include Sytest itself, which will be pulled from git. I realise it's alluded to below, but it might be worth making explicit?

docker/README.md
```diff
@@ -1,40 +1,33 @@
...
 
 ```
-docker run --rm sytest <command>
+$ docker run --rm -it -v /path/to/synapse\:/src -v /path/to/where/you/want/logs\:/logs matrixdotorg/sytest-synapsepy2
 ```
 
-Where `<command>` is `./run-tests.pl` or similar.
+This will run on the same branch in SyTest as the Synapse checkout, if possible, but will fall back to using develop.
+
+If you want to use a checkout of SyTest, mount it to `/test` inside the container by adding `-v /path/to/sytest\:/test` to the docker command.
```
richvdh: s/a checkout/an existing checkout/ ?

docker/synapse_sytest.sh
```diff
@@ -0,0 +1,59 @@
+#! /usr/bin/env bash
+
+set -x
```
richvdh: looks like we're going to need `set -e` here too.

docker/Dockerfile
```diff
@@ -1,40 +1,26 @@
...
+# repo root, not the docker folder
+ADD install-deps.pl ./install-deps.pl
```
richvdh: having added these to the docker image, are they ever actually used?

docker/synapse_sytest.sh
```diff
@@ -0,0 +1,59 @@
...
+
+if [ -e "/test/run-tests.pl" ]
```
richvdh: why is this this `/test/run-tests.pl` rather than `./run-tests.pl` ? The inconsistency makes me wonder if I'm misunderstanding what is going on. Which is entirely plausible, given my level of docker fu.

docker/synapse_sytest.sh
richvdh: (with some faffery around the wget)

.gitignore
```diff
@@ -5,3 +5,7 @@
 /.synapse-base
 /synapse
 /var
+*~
+\#*
+logs/
```
richvdh: can we make these `/logs` and `/results.tap` rather than matching those filenames anywhere?

.gitignore
```diff
@@ -5,3 +5,7 @@
 /.synapse-base
 /synapse
 /var
+*~
```
richvdh: there's some attempt to sort this file; this and the below probably want to go alongside `.*.swp`

PR #485, opened 2018-08-16
Generate JUnit results in the Docker image

hawkowl:


docker/synapse_sytest.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 # Attempt to find a sytest to use.
 # If /test/run-tests.pl exists, it means that a SyTest checkout has been mounted into the Docker image.
-if [ -e "/test/run-tests.pl" ]
+if [ -e "./run-tests.pl" ]
```
richvdh: thank you :smile:

docker/synapse_sytest.sh
```diff
@@ -70,4 +70,8 @@ cp results.tap /logs/results.tap
 cp server-0/homeserver.log /logs/homeserver-0.log
 cp server-1/homeserver.log /logs/homeserver-1.log
 
+# Write out JUnit for CircleCI
+mkdir -p /logs/sytest
+perl /tap-to-junit-xml.pl --puretap --input=/logs/results.tap --output=/logs/sytest/results.xml "SyTest"
```
richvdh: I'm not entirely sure this is the optimal approach, vs (for instance) getting Sytest to spit out JUnit directly, or using a TAP-to-cpan thing from CPAN rather than dumping a script into our repo, but in the interests of not spending days hacking on sytest, let's go with this.

richvdh: One question though: why are the TAP results in `/logs` but the junit results in `/logs/sytest` ? shouldn't it be `/logs/junit-results.xml` or `/logs/junit-results/results.xml` or something?

hawkowl: @richvdh CircleCI accepts the different suites under subdirectories named after the suite, in case you have multiple test suites -- I don't know if it's a convention or not, but it makes CircleCI happier.

hawkowl: unfortunately the script isn't distributed on cpan as far as I can tell (and I needed to make a very minor change to make it spit out test names, not numbers)

PR #486, opened 2018-08-17
[WIP] initial tests for key backup

uhoreg:
tests for https://github.com/matrix-org/synapse/pull/2731 (or whatever its successor will end up being)

### DIFF
--- a/tests/41end-to-end-keys/07-backup.pl
+++ b/tests/41end-to-end-keys/07-backup.pl
@@ -1,0 +1,271 @@
+my $fixture = local_user_fixture();
+
+my $current_version; # FIXME: is there a better way of passing the backup version between tests?
+
+test "Can create backup version",
+   requires => [ $fixture ],
+
+   proves => [qw( can_create_backup_version )],
+
+   do => sub {
+      my ( $user ) = @_;
+
+      do_request_json_for( $user,
+         method  => "POST",
+         uri     => "/unstable/room_keys/version",
+         content => {
+            algorithm => "m.megolm_backup.v1",
+            auth_data => "anopaquestring",
+         }
+      )->then( sub {
+         my ( $content ) = @_;
+         log_if_fail "Content", $content;
+
+         assert_json_keys( $content, "version" );
+
+         $current_version = $content->{version};
+
+         do_request_json_for( $user,
+            method  => "GET",
+            uri     => "/unstable/room_keys/version",
+         );
+      })->then( sub {
+         my ( $content ) = @_;
+         log_if_fail "Content", $content;
+
+         assert_json_keys( $content, "algorithm" );
+
+         assert_json_keys( $content, "auth_data" );
+
+         $content->{algorithm} eq "m.megolm_backup.v1" or
+            die "Expected algorithm to match submitted data";
+
+         $content->{auth_data} eq "anopaquestring" or
+            die "Expected auth_data to match submitted data";
+
+         # FIXME: check that version matches the version returned above
+
+         Future->done(1);
+      });
+   };
+
+test "Can backup keys",
+   requires => [ $fixture, qw( can_create_backup_version ) ],
+
+   proves => [qw( can_backup_e2e_keys )],
+
+   do => sub {
+      my ( $user ) = @_;
+
+      do_request_json_for( $user,
+         method  => "PUT",
+         uri     => "/unstable/room_keys/keys/!abcd/1234",
+         params  => {
+            version => $current_version,
+         },
+         content => {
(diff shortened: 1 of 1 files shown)


tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,271 @@
+my $fixture = local_user_fixture();
+
+my $current_version; # FIXME: is there a better way of passing the backup version between tests?
```
ara4n: this is fine.

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,271 @@
...
+   proves => [qw( can_create_backup_version )],
+
+   do => sub {
+      my ( $user ) = @_;
+
+      do_request_json_for( $user,
+         method  => "POST",
+         uri     => "/unstable/room_keys/version",
+         content => {
+            algorithm => "m.megolm_backup.v1",
```
ara4n: hm, shouldn't we use a real algorithm for this? (or is this a real one? it surprises me it doesn't include info about the ciphers being used, like the m.room.encryption algorithm does)

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,271 @@
...
+
+         assert_json_keys( $content, "auth_data" );
+
+         $content->{algorithm} eq "m.megolm_backup.v1" or
+            die "Expected algorithm to match submitted data";
+
+         $content->{auth_data} eq "anopaquestring" or
+            die "Expected auth_data to match submitted data";
+
+         # FIXME: check that version matches the version returned above
```
ara4n: hm, isn't this just comparing $current_version with $content->{version} or something?

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,271 @@
...
+         );
+      })->then( sub {
+         my ( $content ) = @_;
+         log_if_fail "Content", $content;
+
+         assert_json_keys( $content, "algorithm" );
+
+         assert_json_keys( $content, "auth_data" );
+
+         $content->{algorithm} eq "m.megolm_backup.v1" or
```
ara4n: nitpicking, but i'd probably factor out these constants in the name of DRY

ara4n: actually, ignore this, it's more legible at it is.

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,271 @@
...
+         assert_json_keys( $content, "is_verified" );
+
+         assert_json_keys( $content, "session_data" );
+
+         # The data should not be overwritten, so should be the same as what
+         # was set by the previous test.
+         $content->{first_message_index} == 1 or
+            die "Expected first message index to match submitted data";
+
+         $content->{forwarded_count} == 0 or
```
ara4n: we should probably exercise `forwarded_count` here too - perhaps just put another update attempt in here to check that forwarded keys don't trump non-forwarded ones? (iirc that's how the logic went)

ara4n:
lgtm, other than minor feedback!

uhoreg:
Reminder to myself before I forget: still need to check the bulk uploading/downloading endpoints, as well as the delete endpoints. Also the the get endpoints without a version specified

tests/41end-to-end-keys/07-backup.pl
uhoreg: Yup, except that I don't think the version is returned yet, so I need to fix synapse to do that first. ;)

tests/41end-to-end-keys/07-backup.pl
uhoreg: That's the algorithm given in the proposal, but I agree, it should probably include some algorithm info

tests/41end-to-end-keys/07-backup.pl
uhoreg: So it turns out that Synapse *does* return the version, even though it wasn't in the proposal. :man_shrugging:

ara4n: oops. i can imagine that i fixed up some of the thinkos in the proposal whilst doing the implementation, but then given the implementation never got fully finished, the proposal wasn't updated - sorry.

dbkr:
I've added more tests in https://github.com/matrix-org/sytest/pull/486

dbkr:
Superseded by https://github.com/matrix-org/sytest/pull/503

PR #487, opened 2018-08-25
test to ensure spurious state doesn't leak into incr sync responses when LLing

ara4n:
tests https://github.com/matrix-org/synapse/pull/3760
regression test for https://github.com/vector-im/riot-web/issues/7223


ara4n:
@matrixbot retest this please

tests/10apidoc/09synced.pl
```diff
@@ -205,7 +205,34 @@ sub await_sync_presence_contains {
 }
 
 
-push @EXPORT, qw( assert_room_members assert_state_room_members_matches );
+push @EXPORT, qw( assert_room_members assert_state_room_members_matches assert_room_state);
```
richvdh: nice to have: generally it seems helpful to have a `push @EXPORT` next to each func being exported, rather than one big one at the top

tests/10apidoc/09synced.pl
```diff
@@ -205,7 +205,34 @@ sub await_sync_presence_contains {
...
+
+# assert that the state body of a sync response is made up of the given state types.
```
richvdh: nice to have: generally we seem to be using POD for this sort of thing. It'd be nice to be consistent, even if it's slightly more verbose.

tests/10apidoc/09synced.pl
```diff
@@ -205,7 +205,34 @@ sub await_sync_presence_contains {
...
+sub assert_room_state {
+   my ( $body, $room_id, $state_types ) = @_;
```
richvdh: some docs on what state_types should look like would be helpful, to save people reverse-engineering it.

tests/10apidoc/09synced.pl
```diff
@@ -205,7 +205,34 @@ sub await_sync_presence_contains {
 }
 
 
-push @EXPORT, qw( assert_room_members assert_state_room_members_matches );
+push @EXPORT, qw( assert_room_members assert_state_room_members_matches assert_room_state);
+
+# assert that the state body of a sync response is made up of the given state types.
+sub assert_room_state {
```
richvdh: this should probably be `assert_room_state_types_match` or something, because (a) it only checks for the types, not the whole state, and (b) it checks that they match exactly (as opposed to, for example, being a superset of the desired list)

richvdh: Also, should it be `assert_state_types_match` and take a state block, ie pull lines 214-215 up to the caller, making this equivalent to `assert_state_room_members_matches` rather than `assert_room_members`? That feels like it would be more intuitive to me.

tests/31sync/15lazy-members.pl
```diff
@@ -258,7 +267,6 @@
       });
    };
 
-
```
richvdh: nit: this is inconsistent...

tests/10apidoc/09synced.pl
ara4n: the codebase is very inconsistent on this, but have moved them.

tests/10apidoc/09synced.pl
```diff
@@ -205,16 +205,20 @@ sub await_sync_presence_contains {
...
-sub assert_room_state {
-   my ( $body, $room_id, $state_types ) = @_;
+Assert that the state body of a sync response is made up of the given state types.
 
-   my $room = $body->{rooms}{join}{$room_id};
-   my $state = $room->{state}{events};
+$state is an arrayref of state events.
 
-   log_if_fail "Room", $room;
+$state_types is an arrayref of arrayrefs, each a tuple of type & state_key.
```
richvdh: an example might have been helpful.

ara4n:
retest this please

PR #498, opened 2018-09-27
Tests for missing prevs in incoming transactions

richvdh:
This tries to test the things that are fixed in https://github.com/matrix-org/synapse/pull/3968.

(The first commit is a bit of a refactor; the second updates a currently-disabled test to do something useful.)

### DIFF
--- a/tests/50federation/36-state.pl
+++ b/tests/50federation/36-state.pl
@@ -118,7 +118,7 @@ test "Outbound federation requests /state_ids and correctly handles 404",
          ( $room ) = @_;
 
          # Generate but don't send an event
-         my $missing_event = $room->create_event(
+         my $missing_event = $room->create_and_insert_event(
             type => "m.room.message",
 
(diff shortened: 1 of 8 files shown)


tests/50federation/36-state.pl
```diff
@@ -201,118 +237,208 @@
...
+               }
+
+               my $resp = {
+                  pdu_ids => [
+                     map { $_->{event_id} } values( %state ),
+                  ],
+                  auth_chain_ids => [
+                     # XXX I'm not really sure why we have to return our
+                     # auth_events here, when they are already in the event
+                     map { $_->[0] } @{ $missing_event_y->{auth_events} },
```
erikjohnston: This is meant to be the full auth chain, rather than just the auth events of the state events.

richvdh: k thanks, will update the comment slightly

PR #499, opened 2018-10-01
Test missing auth_event edge case

richvdh:
This is kindof a test for matrix-org/synapse#3960, though to be honest it's just testing one thing that happens to be affected by that (ie, the thing that tripped me up when I was trying to get something else to work, so I happened to have similar tests I could modify relatively easily).

### DIFF
--- a/tests/50federation/36-state.pl
+++ b/tests/50federation/36-state.pl
@@ -448,6 +448,161 @@ test "Outbound federation requests missing prev_events and then asks for /state_
       });
    };
 
+
+test "Federation handles empty auth_events in state_ids sanely",
+   requires => [ $main::OUTBOUND_CLIENT, $main::INBOUND_SERVER, $main::HOMESERVER_INFO[0],
+                 local_user_and_room_fixtures( user_opts => { with_events => 1 } ),
+                 federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $outbound_client, $inbound_server, $info, $creator, $room_id, $user_id ) = @_;
+      my $first_home_server = $info->server_name;
+
+      # in this test, we're going to create a DAG like this:
+      #
+      #   A
+      #   |
+      #   B
+      #   .
+      #   .
+      #   .
+      #  (Y)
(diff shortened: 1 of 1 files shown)


erikjohnston:
I'm hesitant to have a sytest that tests that synapse succeeds if the server lies about the auth chain in /state (in particular, its quite tempting to assert that it does return the full auth chain).

Can we not test this by lying in `pdu_ids` instead?

richvdh:
> Can we not test this by lying in pdu_ids instead?

I don't think so. The thing we are trying to test is crappy behaviour when things are missing from the auth chain, not when they are missing from the state.

richvdh:
[I suppose that a python test might be a better way to test this. But that sounds like work.]

erikjohnston:
I'm surprised that we do auth checks against auth_ids tbh. Relatedly, shouldn't this sytest have failed since the fix hadn't been merged when the tests ran?

richvdh:
> I'm surprised that we do auth checks against auth_ids tbh

well, empirically...

> Relatedly, shouldn't this sytest have failed since the fix hadn't been merged when the tests ran?

no, because the branches are named the same, so the tests ran against the relevant branch of synapse.

erikjohnston:
> no, because the branches are named the same, so the tests ran against the relevant branch of synapse.

Oh, I must be thinking of synapse merged CI that doesn't test the right branch?

erikjohnston:
The only alternatives I can think of is to only assert the transaction request succeeds (assuming it got killed in that bug), or sending another event in the same transaction and asserting that gets persisted correctly.

richvdh:
> only assert the transaction request succeeds (assuming it got killed in that bug)

Not sure what you mean: I think that is what we are asserting.

erikjohnston:
We're also asserting that the events get correctly passed through to the client. If a server responds incorrectly to a `/state/` its not unreasonable for the receiving server to stop processing those events imo

richvdh:
ah right. Possibly.

PR #500, opened 2018-10-02
Use the correct variable for frozen dict usage

hawkowl:


richvdh:
@hawkowl we should not forget about fixing this...

richvdh:
the fail is #4026

anoadragon453:
It looks like we forgot to fix this.

richvdh:
every so often I remember about it, have a little cry, and then carry on with my life.

richvdh:
superceded by #778

PR #501, opened 2018-10-03
Worker-based Docker images

hawkowl:


pydron.py
```diff
@@ -0,0 +1,156 @@
...
+        tries -= 1
+
+        try:
+            r = requests.get("http://" + url, timeout=5)
+            if r.status_code:
+                # We don't care what the status code is, just that it's
+                # responding.
+                return
+        except Exception as e:
+            time.sleep(2)
```
erikjohnston: I think we should poll more frequently than every 2s TBH, otherwise its going to take an age to start up. Something like one every 200ms is probably fine.

pydron.py
```diff
@@ -0,0 +1,156 @@
...
+
+_print("Starting pydron...")
+
+# Sort the args into key:value pairs in a dict
+input_args = sys.argv[1:]
+args = {}
+
+while input_args:
+    key = input_args.pop(0)
+    args[key] = input_args.pop(0)
```
erikjohnston: Can we please use an argument parsing library?

pydron.py
```diff
@@ -0,0 +1,156 @@
...
+            + " --config-path="
+            + args["--synapse-config"]
+        )
+        exc = subprocess.Popen(
+            shlex.split(base), stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        running[i] = exc
+
+    # Give them some time to start up...
+    time.sleep(3)
```
erikjohnston: Why? We're going to explicitly wait for them to start up in a moment

pydron.py
```diff
@@ -0,0 +1,156 @@
...
+    # If we get told to shut down, log it
+    _print("Told to quit because %s" % (repr(e)))
+
+    # Terminate all the workers as cleanly as possible.
+    for x in running.keys():
+        if not running[x].returncode:
+            _print("Terminating %s" % (x,))
+            running[x].terminate()
+        else:
+            _print("%s was already dead" % (x,))
```
erikjohnston: Dendron does a sigterm, waits for it to die, then does a sigkill. We probably want to replicate that.

pydron.py
hawkowl: We don't, though? Should we?

erikjohnston: Oh, I thought that was what:

```
wait_for_start(running[x], urls[x])
```

was doing on 113. Though it doesn't appear that `urls` is defined?

PR #503, opened 2018-10-05
Tests for e2e key backup interface

dbkr:


tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
...
+         my ( $content ) = @_;
+         log_if_fail "Content", $content;
+
+         assert_json_keys( $content, "first_message_index" );
+
+         assert_json_keys( $content, "forwarded_count" );
+
+         assert_json_keys( $content, "is_verified" );
+
+         assert_json_keys( $content, "session_data" );
```
erikjohnston: You can do:

```perl
assert_json_keys( $content, qw( first_message_index forwarded_count is_verified session_data ) )
```

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
...
+         my ( $content ) = @_;
+         log_if_fail "Content", $content;
+
+         do_request_json_for( $user,
+            method  => "GET",
+            uri     => "/unstable/room_keys/keys/!abcd/1234",
+            params  => {
+               version => $current_version,
+            }
+         );
```
erikjohnston: Does this really not fail?

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
...
+            params  => {
+               version => $old_version,
+            },
+            content => {
+               first_message_index => 3,
+               forwarded_count     => 0,
+               is_verified         => JSON::false,
+               session_data        => "anopaquestring",
+            }
+             );
```
erikjohnston: weird indentation

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
...
+   do => sub {
+      my ( $user ) = @_;
+
+      do_request_json_for( $user,
+         method  => "POST",
+         uri     => "/unstable/room_keys/version",
+         content => {
+            algorithm => "m.megolm_backup.v1",
+            auth_data => "anopaquestring",
+         }
```
erikjohnston: I think its going to be beneficial if we create `matrix_create_backup_version` etc functions. There's already a fair bit of duplication and it'll make it easier to write more backup tests.

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
+my $fixture = local_user_fixture();
+
+my $current_version; # FIXME: is there a better way of passing the backup version between tests?
```
erikjohnston: I would prefer if each test was a standalone test, rather than there being implicit dependencies (it makes it harder to reason about each test and change later on).

I think once the APIs are split out into separate functions that will reduce boiler plate enough to make it feasible to just start from scratch for each test. There are also `multi_test` if you want to do multiple tests in one run through.

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -0,0 +1,387 @@
...
+         assert_json_keys( $content, "first_message_index" );
+
+         assert_json_keys( $content, "forwarded_count" );
+
+         assert_json_keys( $content, "is_verified" );
+
+         assert_json_keys( $content, "session_data" );
+
+         $content->{first_message_index} == 1 or
+            die "Expected first message index to match submitted data";
```
erikjohnston: I think you want to use `assert_eq( $content->{first_message_index}, 1, ".." )` as it will have better logging on failures (in terms of logging what the actual values were?)

dbkr:
This should be better! ptal :)

PR #505, opened 2018-10-11
add 3 more tests for gappy sync

ara4n:
and ensure that the failing test for disabled LL on gappy syncs works


tests/31sync/15lazy-members.pl
```diff
@@ -242,11 +245,97 @@
       })->then( sub {
          matrix_join_room( $charlie, $room_id );
       })->then( sub {
+         repeat( sub {
```
erikjohnston: You want `try_repeat` I think, `repeat` is deprecated

PR #506, opened 2018-10-16
Fix some pain when installing the deps

richvdh:
A bunch of fixes to make this stuff work better. See the commits.


richvdh:
@matrixbot retest this please

PR #513, opened 2018-10-25
Tests for room version upgrades

richvdh:
tests for https://github.com/matrix-org/synapse/pull/4091

~~builds on top of #514~~

PR #515, opened 2018-10-26
test upgrading with odd PLs

richvdh:
Make sure that upgrades work where we have permission to send the tombstone, but not set other state.

Follow-on from #513. Tests for https://github.com/matrix-org/synapse/pull/4099.

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+use List::Util qw( first );
+
 # TODO: switch this to '2' once that is released
 my $TEST_NEW_VERSION = 'vdh-test-version';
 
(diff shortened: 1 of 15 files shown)


tests/30rooms/60version_upgrade.pl
```diff
@@ -205,22 +205,32 @@ sub upgrade_room_synced {
...
+               return 0 if not exists $sync_body->{rooms}{join}{$new_room_id};
+               my $tl = $sync_body->{rooms}{join}{$new_room_id}{timeline}{events};
+               foreach my $ev ( @$tl ) {
+                  next if $ev->{type} ne 'm.room.power_levels';
+                  $c++;
+                  log_if_fail "PL event $c in new room", $ev;
+                  return $ev if $c == 2;
+               }
+               return 0;
+            },
```
erikjohnston: I'm a bit confused why this is needed. Can't this be in `upgrade_room_synced`? Do we send a two PL in the new room? It feels a bit useless to have a `upgrade_room_sync` if its not finished upgrading when it returns.

richvdh: Yes, we send two PL events in the new room.

The problem here is (a) knowing when the upgrade has finished (or, more to the point, all the resultant events have been synced back) and (b) given that those events could span more than one sync, returning them from `upgrade_room_synced` to the tests so that they can be verified.

PR #516, opened 2018-10-26
Test that we copy the avatar when upgrading rooms

richvdh:
Follow-on from #513. Test for https://github.com/matrix-org/synapse/pull/4100.


richvdh:
(I think this failed because the synapse changes hadn't landed yet)

PR #520, opened 2018-10-31
Add a test against PL takeovers

richvdh:
This (belatedly) adds a test that the attack fixed in https://github.com/matrix-org/synapse/pull/3397 is fixed.


richvdh:
These tests failed because I hadn't yet landed https://github.com/matrix-org/synapse/pull/4124.

Issue #510, opened 2018-10-16
Re-add the ability to pass which tests to run to ./run-tests.pl


turt2live:
plz @hawkowl

richvdh:
seems to have been fixed by #521

PR #521, opened 2018-11-05
pass arguments from Docker to sytest

uhoreg:
so that I can just run one test and not have to wait for ages for the other tests to run


turt2live:
Fixes https://github.com/matrix-org/sytest/issues/510 (probably)

PR #523, opened 2018-11-09
Add regression test for #4169

dbkr:
Test for https://github.com/matrix-org/synapse/pull/4113

If this fails, it keeps going through the 10 versions which isn't ideal. I tried using the repeat+foreach+while to terminate it but couldn't get it to abort. Feedback welcome on the right runes to make this happen.

### DIFF
--- a/tests/41end-to-end-keys/07-backup.pl
+++ b/tests/41end-to-end-keys/07-backup.pl
@@ -359,7 +359,7 @@ test "Can create more than 10 backup versions",
 
       repeat( sub {
          matrix_create_key_backup( $user );
-      }, foreach => [ 0 .. 10 ]);
+      }, foreach => [ 0 .. 10 ], while => sub { $_[0] -> is_done });
    };
 
 
--- a/tests/41end-to-end-keys/07-backup.pl
+++ b/tests/41end-to-end-keys/07-backup.pl
@@ -349,10 +349,9 @@ test "Deleted & recreated backups are empty",
       });
    };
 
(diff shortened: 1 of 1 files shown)


richvdh:
I think you can do something like:

```perl
      try_repeat( sub {
         matrix_create_key_backup( $user );
      }, foreach => [ 0 .. 10 ], while => sub { $_[0] -> is_done } );
```

whether it's worth doing so, I couldn't say...

richvdh:
the code lgtm but I'm not reassured by the failures!

dbkr:
That's better. Ptal!

tests/41end-to-end-keys/07-backup.pl
```diff
@@ -347,6 +349,19 @@
       });
    };
 
+test "Can create more than 10 backup versions",
+   requires => [ $fixture ],
+
+   bug => 'synapse/4169',
```
richvdh: I don't think this does what you want. It marks the test as an "expected failure".

dbkr: I assume it only does this while the bug is still open though?

erikjohnston: You'd be wrong :) The string isn't interpreted at all, and certainly doesn't go and check the status of any linked issue (nor is there any consistency in the format of the string)

dbkr: looks like you're right though

PR #524, opened 2018-11-10
[WIP] cross-signing support

uhoreg:
corresponds to https://github.com/matrix-org/synapse/pull/4170


uhoreg:
Not sure why the sytests are failing here, when they were passing for me. :-/ Will investigate.

PR #527, opened 2018-11-14
Remove setting and checking of critical flag on tests

babolivier:
Following up on #451

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -50,10 +50,6 @@ is described in more detail in the following sections.
     Provides an ``ARRAY`` reference giving a list of named requirements and
     fixture objects.
 
-``critical``
-    If true and the test fails, the entire test run will bail out at this
-    point; no further tests will be attempted at all.
-
 A call to ``test`` is a simplified version of ``multi_test`` which produces
 only a single line of test output indicating success or failure automatically.
 A call to ``multi_test`` can make use of additional functions within the body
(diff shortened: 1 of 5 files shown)


richvdh:
Out of interest, which test(s) exactly is failing on dendrite?

babolivier:
The first critical test to not pass is `POST /rooms/:room_id/join can join a room` (though it should pass, as should some other tests that are not passing, that's something I'll investigate once I have decent enough bandwidth to deal with it).

richvdh:
> The first critical test to fail is POST /rooms/:room_id/join can join a room (though it should pass

yeah you see, I do wonder if it wouldn't have been better just to fix that, and then possibly remove the `critical` flag on https://github.com/matrix-org/sytest/pull/527/files#diff-5e9bbfbdb3b409994f44369a7f6025cfL236.

Never mind; we can always revert this if we decide it's too painful.

babolivier:
Yeah, the thinking behind removing it wasn't just "that test doesn't pass, let's fix the test rather than fixing the actual issue", but rather following up on a previous discussions in which everyone seemed happy with removing the feature. Also, regardless on how important the test is, given that some tests that are scheduled to happen later on don't rely on that afaik, I'm more in favor of an approach where we skip all tests relying on users joining rooms if that fails, and let the human reading sytest's output decide if that's important, rather than skipping the rest of the whole tests suite.

PR #529, opened 2018-11-26
Add better support for coverage in CI

hawkowl:


docker/synapse_sytest.sh
```diff
@@ -72,4 +72,12 @@ rsync --ignore-missing-args -av server-0 server-1 /logs --include "*/" --include
 mkdir -p /logs/sytest
 perl /tap-to-junit-xml.pl --puretap --input=/logs/results.tap --output=/logs/sytest/results.xml "SyTest"
 
+# Upload coverage to codecov, if running on CircleCI
+if [ -n "$CIRCLECI" ]
+then
+    /venv/bin/coverage combine || IGN=$?
```
richvdh:
```suggestion
    /venv/bin/coverage combine || true
```

Although, why do we want to ignore the error here and not at line 80?

hawkowl: codecov doesn't return non-0 if it fails searching, only if it fails uploading

richvdh: too manynegatives I think ... this line is a search rather than an upload? So according to what you wrote, it always returns 0 here, so the `|| true` is redundant?

PR #530, opened 2018-12-03
utf8 test for pushrule

hawkowl:


hawkowl:
Tests for https://github.com/matrix-org/synapse/issues/4165

PR #532, opened 2018-12-07
Use the from query parameter when querying /messages

babolivier:
While Synapse supports requests to `/messages` without a `from` query parameter, the specs mentions it as required (cf https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-rooms-roomid-messages), therefore SyTest should always call this endpoint with this parameter set.

This PR adds a request to `/sync` before every request to `/messages` and uses the value of the `prev_batch` field from the `/sync` response as the value for `/message`'s `from` parameter.

### DIFF
--- a/tests/10apidoc/34room-messages.pl
+++ b/tests/10apidoc/34room-messages.pl
@@ -138,12 +138,20 @@ test "GET /rooms/:room_id/messages returns a message",
       matrix_send_room_text_message( $user, $room_id,
          body => "Here is the message content",
       )->then( sub {
+         matrix_sync($user)
+      })->then( sub {
+         my ( $sync_body ) = @_;
+         my $token = $sync_body->{next_batch};
+
          do_request_json_for( $user,
             method => "GET",
             uri    => "/r0/rooms/$room_id/messages",
 
             # With no params this does "forwards from END"; i.e. nothing useful
-            params => { dir => "b" },
+            params => {
+                dir => "b",
+                from => $token,
+            },
          )
       })->then( sub {
          my ( $body ) = @_;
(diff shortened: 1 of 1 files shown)


tests/10apidoc/34room-messages.pl
```diff
@@ -138,12 +138,20 @@ sub matrix_send_room_text_message
       matrix_send_room_text_message( $user, $room_id,
          body => "Here is the message content",
       )->then( sub {
+         matrix_sync($user)
```
richvdh:
```suggestion
         matrix_sync( $user )
```

tests/10apidoc/34room-messages.pl
```diff
@@ -138,12 +138,20 @@ sub matrix_send_room_text_message
...
+      })->then( sub {
+         my ( $sync_body ) = @_;
+         my $token = $sync_body->{next_batch};
+
          do_request_json_for( $user,
             method => "GET",
             uri    => "/r0/rooms/$room_id/messages",
 
             # With no params this does "forwards from END"; i.e. nothing useful
```
richvdh: this comment is pretty redundant

tests/10apidoc/34room-messages.pl
```diff
@@ -138,12 +138,20 @@ sub matrix_send_room_text_message
...
-            params => { dir => "b" },
+            params => {
+                dir => "b",
+                from => $token,
```
richvdh: The spec says:

> This token can be obtained from a prev_batch token returned for each room by the sync API, or from a start or end token returned by a previous request to this endpoint.

This is neither.

PR #533, opened 2018-12-10
Configure Dendrite to log to file

babolivier:
Add a section to Dendrite's generated configuration file to log to files on top of logging to the standard output.

I set the log level to `debug` as I assume users will look at log files to understand precisely what's going on when running tests (even though afaik there's no debug logs in Dendrite yet so that doesn't make a difference right now, but might in the future).


babolivier:
It has been, and it does.

PR #535, opened 2018-12-14
support user_type in hmac

neilisfragile:


neilisfragile:
have made change on synapse side backward compatible

PR #539, opened 2018-12-21
Use the from query parameter when querying /messages

babolivier:
This is #532 except I messed up and somehow merged it into `develop`, so I reverted that and am opening the PR again.

Original initial comment from the original PR:

While Synapse supports requests to `/messages` without a `from` query parameter, the specs mentions it as required (cf https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-rooms-roomid-messages), therefore SyTest should always call this endpoint with this parameter set.

This PR adds a request to `/sync` before every request to `/messages` and uses the value of the `prev_batch` field from the `/sync` response as the value for `/message`'s `from` parameter.

Unresolved (at the time I'm writing this) review available in the original PR. Will solve it (soon-ish) then re-request a review.

### DIFF
--- a/tests/10apidoc/34room-messages.pl
+++ b/tests/10apidoc/34room-messages.pl
@@ -141,7 +141,7 @@ test "GET /rooms/:room_id/messages returns a message",
          matrix_sync( $user )
       })->then( sub {
          my ( $sync_body ) = @_;
-         my $token = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{next_batch};
+         my $token = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{prev_batch};
 
          do_request_json_for( $user,
             method => "GET",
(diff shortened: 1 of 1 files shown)


babolivier:
Right, giving it another look, sorry for the delay.

tests/10apidoc/34room-messages.pl
```diff
@@ -138,10 +138,10 @@ sub matrix_send_room_text_message
       matrix_send_room_text_message( $user, $room_id,
          body => "Here is the message content",
       )->then( sub {
-         matrix_sync($user)
+         matrix_sync( $user )
       })->then( sub {
          my ( $sync_body ) = @_;
-         my $token = $sync_body->{next_batch};
+         my $token = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{next_batch};
```
richvdh: `next_batch` doesn't exist here (at least in the spec), so I suspect this PR isn't changing anything...

s/next_batch/prev_batch.

(likewise below)

babolivier: Gah, indeed. Weirdly enough, SyTest doesn't complain about that, nor does Synapse by being given an invalid token (which I'd expect it to do). /me fixes

PR #540, opened 2019-01-07
Add configuration for recaptcha auth to Dendrite

babolivier:
Allows sytest to configure recaptcha registration/login on Dendrite.


babolivier:
Not sure why I didn't merge this PR back in Jan, doing it now.

PR #541, opened 2019-01-16
Add dockerfile for dendrite

anoadragon453:
Adds a Dockerfile for Dendrite which prepares postgres and installs the needed dependencies for sytest and dendrite to run.


anoadragon453:
Not sure why adding two unrelated files would cause the tests to fail...

anoadragon453:
Currently blocked on #452

anoadragon453:
@erikjohnston Is it possible to run commands that require postgres to be running like creating databases in the Dockerfile?

erikjohnston:
> @erikjohnston Is it possible to run commands that require postgres to be running like creating databases in the Dockerfile?

I believe so, but its probably best to leave that to dendrite when it starts?

anoadragon453:
Ok, assumed that was part of the database setup code that you mentioned, but if not then that's fine to leave to the actual running container.

docker/dendrite_sytest.sh
```diff
@@ -0,0 +1,63 @@
...
+    tar --strip-components=1 -xf sytest.tar.gz
+
+fi
+
+# Make sure all Perl deps are installed -- this is done in the docker build so will only install packages added since the last Docker build
+dos2unix ./install-deps.pl
+./install-deps.pl
+
+# Initialise the database files and start the database
+su -c '/usr/lib/postgresql/9.6/bin/initdb -E "UTF-8" --lc-collate="en_US.UTF-8" --lc-ctype="en_US.UTF-8" --username=postgres' postgres
```
erikjohnston: Oh sorry, you *can* and should run this in the dockerfile, just the create database stuff below can stay. You then also don't need to explicitly call start as it'll have been started automatically

docker/dendrite_sytest.sh
```diff
@@ -0,0 +1,63 @@
...
+        wget -q https://github.com/matrix-org/sytest/archive/develop.tar.gz -O sytest.tar.gz
+    }
+
+    tar --strip-components=1 -xf sytest.tar.gz
+
+fi
+
+# Make sure all Perl deps are installed -- this is done in the docker build so will only install packages added since the last Docker build
+dos2unix ./install-deps.pl
+./install-deps.pl
```
erikjohnston: Can you please also run this in the dockerfile, that way most dependencies will be pre-installed but new ones will still be picked up. Otherwise it takes like 15mins

docker/dendrite_sytest.sh
anoadragon453: Mm, you say that but computer says no:

```
+ su -c 'psql -c "CREATE USER dendrite PASSWORD '\''itsasecret'\''" postgres'
psql: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
$
```

erikjohnston: Yeah, its just the initdb that you want to move up (though there is nothing stopping you from starting the database technically)

anoadragon453: By move up did you mean move it to the Dockerfile?

PR #545, opened 2019-01-21
Test sending v2 invites to server

erikjohnston:
This PR is in two parts.

In the second part I've tried to reuse as much code as possible between `invite_server` using v1 and v2 APIs, but it feels fairly ming :/   Alternate suggestions welcome.

Tests https://github.com/matrix-org/synapse/pull/4402

### DIFF
--- a/tests/50federation/35room-invite.pl
+++ b/tests/50federation/35room-invite.pl
@@ -47,7 +47,7 @@ test "Outbound federation can send invites",
       );
    };
 
-test "Inbound federation can receive invites",
+test "Inbound federation can receive invites via v1 API",
    requires => [ local_user_fixture( with_events => 1 ), $main::INBOUND_SERVER,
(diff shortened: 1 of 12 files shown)


tests/50federation/35room-invite.pl
```diff
@@ -142,6 +187,76 @@ sub invite_server
...
+      method   => "PUT",
+      hostname => $first_home_server,
+      uri      => "/v1/invite/$room_id/$invitation->{event_id}",
+
+      content => $invitation,
+   )->then( sub {
+      my ( $response ) = @_;
+
+      # $response seems to arrive with an extraneous layer of wrapping as
+      # the result of a synapse implementation bug (SYN-490).
```
richvdh: s|SYN-490| https://github.com/matrix-org/synapse/issues/1383 |

PR #547, opened 2019-01-22
Add tests for transfer of tags and dm status on room upgrade

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/4412

Adds 2 tests to check that the same room tags and `m.direct` status is set on a room before and after an upgrade.

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -75,27 +75,25 @@ sub is_direct_room {
    my ( $user, $room_id ) = @_;
 
    # Download account data events from sync
-   matrix_sync( $user )->then( sub {
-      my ( $sync_body ) = @_;
-
+   matrix_get_account_data( $user, "m.direct" )->then( sub {
       # Should only have the m.direct event in account_data
-      my $account_data = $sync_body->{account_data}{events};
+      my ( $data ) = @_;
+
+      log_if_fail "m.direct account data", $data;
 
       # Check if the room_id is in the list of direct rooms
-      foreach my $event ( @$account_data ) {
-         if ( $event->{type} eq "m.direct" ) {
-            my $room_ids = $event->{content}{$user->user_id};
+      foreach my $user_id ( keys %{ $data } ) {
+         my $room_ids = $data->{$user_id};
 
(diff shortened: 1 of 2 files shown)


tests/30rooms/60version_upgrade.pl
```diff
@@ -59,6 +61,45 @@ sub upgrade_room {
...
+
+Check if a room ID is considered direct by the given user.
+
+=cut
+
+sub is_direct_room {
+   my ( $user, $room_id ) = @_;
+
+   # Download account data events from sync
+   matrix_sync( $user )->then( sub {
```
richvdh: How about using `GET /_matrix/client/r0/user/{user}/account_data/m.direct` (https://github.com/matrix-org/matrix-doc/issues/1339) ?

anoadragon453: Ah cool.

anoadragon453: Indeed this is already a function: https://github.com/matrix-org/sytest/blob/665210b5e13e4adf9194512d44677a2a9ef7ed0a/tests/10apidoc/44account_data.pl#L71-L79

tests/30rooms/60version_upgrade.pl
```diff
@@ -59,6 +61,43 @@ sub upgrade_room {
    });
 }
 
+=head2 is_direct_room
+
+    is_direct_room( $user, $room_id )->then( sub {
+        my ( $is_direct ) = @_;
+    })
+
+Check if a room ID is considered direct by the given user.
```
richvdh: I found this slightly hard to parse; how about something like:

```suggestion
Check if a room ID is considered to be a direct chat by the given user.
```

tests/30rooms/60version_upgrade.pl
```diff
@@ -59,6 +61,43 @@ sub upgrade_room {
...
+
+         # Return whether the given room ID is in the response
+         foreach my $room (@$room_ids) {
+            if ( $room eq $room_id ) {
+               return Future->done( 1 );
+            }
+         }
+      }
+
+      # Didn't find an direct romo with our room ID
```
richvdh: "an direct romo"?

PR #548, opened 2019-01-23
Print sytest sha1 during CI run

anoadragon453:
To address https://riot.im/experimental/#/room/#synapse-dev:matrix.org/$154816555286JknwK:sw1v.org

Prints the commit hash of sytest that we're using for a run.

### DIFF
--- a/docker/synapse_sytest.sh
+++ b/docker/synapse_sytest.sh
@@ -24,6 +24,9 @@ else
 
 fi
 
+sytest_commit="$(git rev-parse HEAD)" || sytest_commit="Unknown"
+echo "Running on sytest commit $sytest_commit"
+
 # PostgreSQL setup
 if [ -n "$POSTGRES" ]
 then


richvdh:
Sadly, I don't think this is going to work, because we download a tarball of sytest, rather than checking it out from git. I'm not entirely sure why we do that, tbh.

anoadragon453:
Could simply switch to cloning.

richvdh:
yarp. I think @hawkowl wrote the current impl - she may know if there's an advantage to it.

hawkowl:
we could do a shallow clone so it doesn't take forever, but just downloading a tarball was easier at the time.

anoadragon453:
I change change this PR up post-FOSDEM to do a shallow clone since that seems like the simplest solution.

richvdh:
gonna close this for now.

PR #552, opened 2019-01-29
Correctly handle new event ID format in rooms v3

erikjohnston:


tests/10apidoc/31room-state.pl
```diff
@@ -64,7 +64,7 @@
       )->then( sub {
          my ( $body ) = @_;
 
-         assert_json_keys( $body, qw( sender event_id room_id content ));
```
erikjohnston: Note that this is when we've specified ?format=event, i.e. asking for the raw federation format

cpanfile
```diff
@@ -50,3 +50,5 @@ requires 'Protocol::Matrix', '>= 0.02';
 requires 'Struct::Dumb', '>= 0.04';
 requires 'URI::Escape';
 requires 'YAML';
+
+requires 'URI::Escape::XS';
```
richvdh: this file is sorted... can you stick it in the right place please

richvdh: Is there a reason to use this instead of the URI::Escape we already have? Sytest already downloads most of the internet :/

richvdh: (plus if you bump the deps you have to rebuild the docker images and faff)

erikjohnston: Oh, I got confused with `URI::Encode`, which says to use `URI::Escape:XS`

PR #556, opened 2019-01-30
Add basic functionality test for room versions

erikjohnston:
Currently we don't test any room versions other than the default, so
this adds some very basic tests to ensure that some of the core
funcationlity works.


richvdh:
I'm going to land this because I want to land a test for #4532 on top of it

PR #558, opened 2019-01-31
Regression test for redactions in room v3

richvdh:


richvdh:
looks to have failed because it ran against the release branch of synapse, which lacks matrix-org/synapse#4513

PR #559, opened 2019-02-06
test endpoint for updating backup versions

uhoreg:
companion to https://github.com/matrix-org/synapse/pull/4580

### DIFF
--- a/tests/41end-to-end-keys/07-backup.pl
+++ b/tests/41end-to-end-keys/07-backup.pl
@@ -40,6 +40,54 @@ test "Can create backup version",
       });
    };
 
+test "Can update backup version",
+   requires => [ $fixture ],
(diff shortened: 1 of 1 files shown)


richvdh:
sigh. Will pick this up and see what's the matter with it.

uhoreg:
I wonder how I forgot to merge this one when I merged the synapse PR :frowning_face: It looks like it's failing on unrelated tests.  Maybe because it has newer tests that are running against an older synapse?

richvdh:
> Maybe because it has newer tests that are running against an older synapse?

Oh, almost certainly this. I've deleted the old (merged) synapse branch and updated this one again. It passes locally at least...

richvdh:
\o/

PR #563, opened 2019-02-14
Add test for transferring bans on a room upgrade

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/4642

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -437,24 +437,21 @@ test "/upgrade copies ban events to the new room",
       })->then( sub {
          ( $new_room_id, ) = @_;
 
-         matrix_sync_again( $creator );
-      })->then( sub {
-         my ( $sync_body ) = @_;
+         await_sync_timeline_or_state_contains( $creator, $new_room_id, check => sub {
+            my ( $event ) = @_;
 
-         log_if_fail "sync body", $sync_body;
+            return unless $event->{type} eq "m.room.member";
+            return unless $event->{state_key} eq "\@bob:matrix.org";
 
-         my $room = $sync_body->{rooms}{join}{$new_room_id};
+            assert_deeply_eq(
+               $event->{content},
+               $content,
+               "no ban in replacement room",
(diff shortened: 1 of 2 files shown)


tests/30rooms/60version_upgrade.pl
```diff
@@ -408,6 +408,58 @@ sub upgrade_room_synced {
...
+      })->then( sub {
+         my ( $sync_body ) = @_;
+
+         log_if_fail "sync body", $sync_body;
+
+         my $room = $sync_body->{rooms}{join}{$new_room_id};
+
+         my $event = first {
+            $_->{type} eq "m.room.member" && $_->{state_key} eq '@bob:matrix.org',
+         } @{ $room->{timeline}->{events} };
```
erikjohnston: We should probably check if its in the `state` section, as that's also a valid place.

On top of that, we can write this entire thing using `await_sync_timeline_contains`:

```perl
await_sync_timeline_contains( $user, $room, check => sub {
    return unless $event->{type} eq "m.room.member";
    return unless $event->{state_key} eq "@bob:matrix.org";

    assert_deeply_eq(
        $event->{content},
        $content,
        "ban in replacement room",
    );

    return 1;
});
```

Which makes the test a lot shorter, and stops you from having to do the manual syncs.

We will have to create a new `await_sync_timeline_or_state_contains` function, but that is fairly trivial to do.

tests/30rooms/60version_upgrade.pl
```diff
@@ -408,6 +408,55 @@ sub upgrade_room_synced {
...
+            assert_deeply_eq(
+               $event->{content},
+               $content,
+               "no ban in replacement room",
+            );
+
+            return 1;
+         });
+
+         Future->done(1);
```
erikjohnston: Remove this, as otherwise you're not using the result of `await_sync_timeline_or_state_contains`

anoadragon453: Ah lol. Happy to see that wasn't the reason it passed.

PR #564, opened 2019-02-22
Test history visibility works for backfill

erikjohnston:
Synapse PR: https://github.com/matrix-org/synapse/pull/4718

### DIFF
--- a/tests/30rooms/30history-visibility.pl
+++ b/tests/30rooms/30history-visibility.pl
@@ -1,5 +1,6 @@
 use JSON qw( encode_json );
 use URI::Escape qw( uri_escape );
+use Future::Utils qw( repeat );
 
 use constant { YES => 1, NO => !1 };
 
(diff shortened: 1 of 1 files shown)


tests/30rooms/30history-visibility.pl
```diff
@@ -533,6 +534,43 @@ sub test_history_visibility
       });
    };
 
+test "Backfill works correctly with history visilibty set to joined",
```
richvdh: visilibty

tests/30rooms/30history-visibility.pl
```diff
@@ -533,6 +534,43 @@ sub test_history_visibility
...
+
+      matrix_set_room_history_visibility( $user, $room_id, "joined" )
+      ->then( sub {
+         repeat( sub {
+            my $msgnum = $_[0];
+
+            matrix_send_room_text_message( $user, $room_id, body => "Message $msgnum" );
+         }, foreach => [ 1 .. 100 ]);
+      })->then( sub {
+         matrix_join_room( $another_user, $room_alias );
```
richvdh: what's the point of this?

tests/30rooms/30history-visibility.pl
```diff
@@ -533,6 +534,43 @@ sub test_history_visibility
...
+   do => sub {
+      my ( $user, $room_id, $room_alias, $another_user, $remote_user, $room_alias_name ) = @_;
+
+      matrix_set_room_history_visibility( $user, $room_id, "joined" )
+      ->then( sub {
+         repeat( sub {
+            my $msgnum = $_[0];
+
+            matrix_send_room_text_message( $user, $room_id, body => "Message $msgnum" );
+         }, foreach => [ 1 .. 100 ]);
```
richvdh: could we use fewer than 100 messages?

tests/30rooms/30history-visibility.pl
```diff
@@ -533,6 +534,43 @@ sub test_history_visibility
...
+      })->then( sub {
+         matrix_join_room( $remote_user, $room_alias );
+      })->then( sub {
+         matrix_get_room_messages( $remote_user, $room_id, limit => 10 )
+      })->then( sub {
+         my ( $body ) = @_;
+         log_if_fail "messages body", $body;
+
+         my $chunk = $body->{chunk};
+         @$chunk == 1 or die "Expected 1 message";
```
richvdh: s/message/event/

tests/30rooms/30history-visibility.pl
erikjohnston: I was padding out so that we didn't get any events from before the `m.room.history_visibility` in the backfill, but that's a bit fragile so I've rejigged the test.

tests/30rooms/30history-visibility.pl
erikjohnston: I've added a comment, but the bug only manifests if there is a state event in the chunk. (Initially I thought it also had to appear in the auth chain, hence using a join, but I don't think that's true)

tests/30rooms/30history-visibility.pl
```diff
@@ -534,24 +534,27 @@ sub test_history_visibility
...
 
             matrix_send_room_text_message( $user, $room_id, body => "Message $msgnum" );
-         }, foreach => [ 1 .. 100 ]);
+         }, foreach => [ 1 .. 10 ]);
       })->then( sub {
          matrix_join_room( $another_user, $room_alias );
       })->then( sub {
          matrix_send_room_text_message( $user, $room_id, body => "2" );
       })->then( sub {
+         # We now send a state event to ensure they're correctly handled in
```
richvdh: erm; is it not the *other* join that we're doing to test this?

This join is necessary for us to see anything at all?

tests/30rooms/30history-visibility.pl
```diff
@@ -534,24 +534,27 @@ sub test_history_visibility
...
+         # backfill, this was a bug in synapse (c.f. #1943)
```
richvdh: s/,/./

tests/30rooms/30history-visibility.pl
erikjohnston: Oh I'm such a crank. Sorry, yes, imagine that comment on the previous join.

PR #565, opened 2019-02-25
Make sytest support auth rate limiting

babolivier:
Paired with https://github.com/matrix-org/synapse/pull/4735


babolivier:
Merging this as matrix-org/synapse#4735 just got merged.

PR #566, opened 2019-02-25
Check that event_id is given over state fetching over federation

hawkowl:


erikjohnston:
What are all the changes in `Synapse.pm`?

hawkowl:
@erikjohnston a mismerge (oops)

PR #567, opened 2019-03-01
Make the userdir synced not rely on being able to search for yourself

hawkowl:


tests/52user-directory/01public.pl
```diff
@@ -5,21 +5,32 @@
...
+         ( $searching_user ) = @_;
+
          matrix_create_room( $user,
             preset => "public_chat",
          );
+      })->then( sub {
+         my ( $room_id ) = @_;
+         matrix_join_room( $searching_user, $room_id );
       })->then( sub {
          ( $room_id ) = @_;
```
richvdh: this looks bogus. join_room doesn't return a room id. (looks like the function-level $room_id is unused and should be vaped?)

tests/52user-directory/01public.pl
```diff
@@ -44,17 +55,22 @@
       my ( $user ) = @_;
 
       my $room_id;
```
richvdh: this is redundant I think?

tests/52user-directory/01public.pl
```diff
@@ -44,17 +55,22 @@
...
 
-      matrix_set_displayname( $user, $displayname )
-      ->then( sub {
          matrix_create_room( $user,
             preset => "private_chat",
          );
       })->then( sub {
-         ( $room_id ) = @_;
-
          matrix_get_user_dir_synced( $user, $displayname );
```
richvdh: this user still seems to be looking for themselves?

tests/52user-directory/01public.pl
```diff
@@ -384,12 +400,25 @@
...
       my $room_id;
+      my $searching_user;
 
       my $displayname = generate_random_displayname();
       my $second_displayname = generate_random_displayname();
+      my $searching_displayname = generate_random_displayname();
 
       matrix_set_displayname( $user, $displayname )
       ->then( sub {
+
```
richvdh: ugly whitespace makes ugly perl uglier

tests/52user-directory/01public.pl
```diff
@@ -384,12 +400,25 @@
...
+         matrix_create_user_on_server( $user->http,
```
richvdh: I'm not quite sure why we are creating a user that is never used here?

tests/52user-directory/01public.pl
```diff
@@ -435,30 +466,43 @@ sub generate_random_displayname
 }
 
 
-# Get the user direectory after a change has been made. This creates a new user
+# Get the user directory after a change has been made. This creates a new user
```
richvdh:
```suggestion
# Get the user directory after a change has been made. This creates two new users
```

tests/52user-directory/01public.pl
```diff
@@ -435,30 +466,43 @@ sub generate_random_displayname
...
 # and then polls the user directory until we see it. This is to get around the
```
richvdh:
```suggestion
# and then polls the user directory until one sees the other. This is to get around the
```

tests/52user-directory/01public.pl
hawkowl: done

tests/52user-directory/01public.pl
hawkowl: fixed

tests/52user-directory/01public.pl
hawkowl: fixed

tests/52user-directory/01public.pl
hawkowl: they have to join the room so that the dir is populated

richvdh: they don't seem to join a room?

PR #573, opened 2019-03-13
Make running sytest docker work in worker mode

erikjohnston:
This also includes moving a bunch of setup into the docker image so that
the containers start up a lot faster.

This does make the docker images about ~1.1GB, which is quite a lot bigger than before.

Depends on https://github.com/matrix-org/synapse/pull/4853

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get -qq install -y \
     build-essential \
     perl \
     wget \
-    postgresql \
+    postgresql-9.6 \
     postgresql-client \
     libpq-dev \
     libssl-dev \
--- a/docker/Dockerfile-synapsepy2
+++ b/docker/Dockerfile-synapsepy2
@@ -8,7 +8,7 @@ ENV PGDATA=/var/lib/postgresql/data
 
 RUN su -c '/usr/lib/postgresql/9.6/bin/initdb -E "UTF-8" --lc-collate="en_US.UTF-8" --lc-ctype="en_US.UTF-8" --username=postgres' postgres
 
-# Turn of all the fsync stuff for postgres
+# Turn off all the fsync stuff for postgres
 RUN mkdir -p /etc/postgresql/9.6/main/conf.d/
 RUN echo "fync=off" > /etc/postgresql/9.6/main/conf.d/fsync.conf
 RUN echo "full_page_writes=off" >> /etc/postgresql/9.6/main/conf.d/fsync.conf
@@ -22,7 +22,7 @@ RUN /venv/bin/pip install -q --no-cache-dir matrix-synapse[all]
 RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2
 
 # Uninstall matrix-synapse package so it doesn't collide with the version we try
-# Zand test
+# and test
 RUN /venv/bin/pip uninstall -q --no-cache-dir -y matrix-synapse
 
 # The dockerfile context, when ran by the buildscript, will actually be the
--- a/docker/Dockerfile-synapsepy2
+++ b/docker/Dockerfile-synapsepy2
@@ -1,14 +1,35 @@
 FROM matrixdotorg/sytest:latest
 
-RUN apt-get -qq install -y python python-dev python-virtualenv
+RUN apt-get -qq update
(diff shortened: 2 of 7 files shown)


docker/Dockerfile-synapsepy2
```diff
@@ -1,14 +1,35 @@
...
 # /src is where we expect Synapse to be
 RUN mkdir /src
 
+# Create the virutal env upfront so we don't need to keep reinstall dependencies
+RUN $PYTHON -m virtualenv -p $PYTHON /venv/
+RUN /venv/bin/pip install -q --no-cache-dir matrix-synapse[all]
+RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2
+
+# Uninstall matrix-synapse package so it doesn't collide with the version we try
+# Zand test
```
richvdh: Zand

docker/Dockerfile-synapsepy3
```diff
@@ -1,14 +1,35 @@
...
 # /src is where we expect Synapse to be
 RUN mkdir /src
 
+# Create the virutal env upfront so we don't need to keep reinstall dependencies
+RUN $PYTHON -m virtualenv -p $PYTHON /venv/
+RUN /venv/bin/pip install -q --no-cache-dir matrix-synapse[all]
+RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2
+
+# Uninstall matrix-synapse package so it doesn't collide with the version we try
+# Zand test
```
richvdh: Zand

docker/Dockerfile-synapsepy2
```diff
@@ -1,14 +1,35 @@
...
-RUN apt-get -qq install -y python python-dev python-virtualenv
+RUN apt-get -qq update
+RUN apt-get -qq install -y python python-dev python-virtualenv dos2unix eatmydata
 
 ENV PYTHON=python2
+ENV PGDATA=/var/lib/postgresql/data
+
+RUN su -c '/usr/lib/postgresql/9.6/bin/initdb -E "UTF-8" --lc-collate="en_US.UTF-8" --lc-ctype="en_US.UTF-8" --username=postgres' postgres
+
+# Turn of all the fsync stuff for postgres
```
richvdh: off

docker/Dockerfile-synapsepy3
```diff
@@ -1,14 +1,35 @@
 FROM matrixdotorg/sytest:latest
```
richvdh: it'd be really nice to combine these two into one dockerfile with an ARG in the same way that synapse's dockerfile works.

richvdh: given that the lifespan of the py2 version is only a fortnight, not important, I guess.

docker/README.md
```diff
@@ -18,7 +18,7 @@ $ docker run --rm -it -v /path/to/synapse\:/src -v /path/to/where/you/want/logs\
 
 This will run on the same branch in SyTest as the Synapse checkout, if possible, but will fall back to using develop.
 
-If you want to use an existing checkout of SyTest, mount it to `/test` inside the container by adding `-v /path/to/sytest\:/test` to the docker command.
+If you want to use an existing checkout of SyTest, mount it to `/test` inside the container by adding `-v /path/to/sytest\:/test` to the docker command (note that this will be written to).
```
richvdh: nit: wrapping is a thing

docker/pydron.py
```diff
@@ -143,6 +90,8 @@ def do_GET(self):
     server = HTTPServer(('', PORT_NUMBER), WebHandler)
     server.serve_forever()
 
+except KeyboardInterrupt:
```
richvdh: what's this for, ooi?

docker/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get -qq install -y \
     build-essential \
     perl \
     wget \
-    postgresql \
+    postgresql-9.6 \
```
richvdh: it feels weird that we are doing this here rather than in Dockerfile-synapsepy2 or whatever

docker/pydron.py
erikjohnston: This (empirically) captures when we ask docker to stop half way through a job.

PR #575, opened 2019-03-20
Start workers in parallel to speed up spinup

richvdh:


docker/pydron.py
```diff
@@ -73,7 +74,13 @@ def _print(out):
             + " --config-path="
             + configs[i]
         )
-        subprocess.run(shlex.split(base), check=True)
+        worker_processes[appname] = subprocess.Popen(shlex.split(base))
```
erikjohnston: How does this interact with stdout/stderr? It would be good to ensure that they don't go missing (or get entangled) as they sometimes have useful info when synapse fails to start

richvdh: they go to pydron's stderr/stdout, as they did previously.

So yes, they do get entangled. I guess we could buffer them up and then spew them out sequentially in the second loop.

PR #579, opened 2019-03-22
Don't assume that push rule changes propogate instantly

erikjohnston:
This test broke in worker mode, so we do the checks in a loop until they pass.


richvdh:
prop**a**gate, ftr

PR #581, opened 2019-03-22
Don't assume /notifications updates instantly

erikjohnston:
This test broke in worker mode.

### DIFF
--- a/tests/61push/09_notifications_api.pl
+++ b/tests/61push/09_notifications_api.pl
@@ -30,41 +30,45 @@ test "Notifications can be viewed with GET /notifications",
             body => "Test message 2",
          );
       })->then( sub {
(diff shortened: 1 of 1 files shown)


tests/61push/09_notifications_api.pl
```diff
@@ -30,41 +30,45 @@
...
-         do_request_json_for( $user1,
-            method  => "GET",
-            uri     => "/unstable/notifications",
-         );
-      })->then( sub {
-         my ( $body ) = @_;
+         retry_until_success {
+            do_request_json_for( $user1,
+               method  => "GET",
+               uri     => "/unstable/notifications",
```
richvdh: we should really change these to be /r0/, though leaving them for another PR makes some sense

PR #585, opened 2019-03-22
Fix race in /notifications test in worker mode

erikjohnston:
There was a race where we would send the message before the event
creator had realised it should be calculating notifications for that
user, and so nothing would ever come down `/notifications`.

### DIFF
--- a/tests/61push/09_notifications_api.pl
+++ b/tests/61push/09_notifications_api.pl
@@ -24,34 +24,53 @@ test "Notifications can be viewed with GET /notifications",
       })->then( sub {
          my ( $event_id ) = @_;
 
+         # We need to send a read receipt before the server will start
+         # calculating notifications.
          matrix_advance_room_receipt( $user1, $room_id, "m.read" => $event_id );
       })->then( sub {
-         matrix_send_room_text_message( $user2, $room_id,
-            body => "Test message 2",
-         );
-      })->then( sub {
+         # It may take a while for the server to start calculating
+         # notifications, so we repeatedly send message and check if anything
+         # turns up in `/notifications`
          retry_until_success {
-            do_request_json_for( $user1,
-               method  => "GET",
-               uri     => "/unstable/notifications",
-            )->then( sub {
-               my ( $body ) = @_;
+            matrix_send_room_text_message( $user2, $room_id,
+               body => "Test message 2",
+            )->then(sub {
+               do_request_json_for( $user1,
+                  method  => "GET",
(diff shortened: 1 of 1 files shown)


tests/61push/09_notifications_api.pl
```diff
@@ -24,34 +24,53 @@
...
+                  assert_json_keys( $body, "notifications" );
 
-               assert_json_keys( $notifs->[0], qw( room_id actions event read ts ) );
-               assert_ok( exists $notifs->[0]{profile_tag}, "profile_tag defined" );
+                  my $notifs = $body->{notifications};
 
-               my $notif = $notifs->[0];
-               assert_eq( $notif->{read}, JSON::false );
+                  # We just want something to turn up
+                  scalar @{ $notifs } or die "no notificaions";
```
richvdh:
```suggestion
                  scalar @{ $notifs } or die "no notifications";
```

erikjohnston:
> surely `/notifications` hits a different worker to `/send` so it is not a given that a `/notifications` request immediately after the send will see the message?

Yup, but we wrap the checks in a retry loop, so if it doesn't see that message immediately, it will hopefully pick it up next loop round, i.e. something like:

1. Send message A
2. Call `/notifications` and it returns nothing, so we loop
3. Send message B
4. Call `/notifications` and it returns message A (and maybe B), so we continue the test

The retry loop is actually guarding two race conditions: 1) we need to repeatedly send messages as the first few might not use the new push rules and 2) we need to call `/notifications` repeatedly as it might take a while for messages to propagate. Hence why we both send a new message and check `/notifications` in the same loop.

tests/61push/09_notifications_api.pl
```diff
@@ -24,34 +24,53 @@
...
+         # We need to send a read receipt before the server will start
+         # calculating notifications.
          matrix_advance_room_receipt( $user1, $room_id, "m.read" => $event_id );
       })->then( sub {
-         matrix_send_room_text_message( $user2, $room_id,
-            body => "Test message 2",
-         );
-      })->then( sub {
+         # It may take a while for the server to start calculating
+         # notifications, so we repeatedly send message and check if anything
```
richvdh:
```suggestion
         # notifications, so we repeatedly send messages and check if anything
```

richvdh:
> The retry loop is actually guarding two race conditions: 1) we need to repeatedly send messages as the first few might not use the new push rules and 2) we need to call /notifications repeatedly as it might take a while for messages to propagate. 

Perhaps you could stick this in the comment.

PR #586, opened 2019-03-22
Fix rejected pusher test in worker mode

erikjohnston:
It assumed that pusher changes would propagate instantly.

### DIFF
--- a/tests/61push/08_rejected_pushers.pl
+++ b/tests/61push/08_rejected_pushers.pl
@@ -23,17 +23,15 @@ sub wait_for_push
    my ( $path, $response ) = @_;
 
    await_http_request( $path, sub {
-     my ( $request ) = @_;
-     my $body = $request->body_from_json;
-
-     return unless $body->{notification}{type};
-     return unless $body->{notification}{type} eq "m.room.message";
-     return 1;
-   })->then( sub {
       my ( $request ) = @_;
+      my $body = $request->body_from_json;
 
+      # Respond to all requests, even if we filiter them out
       $request->respond_json( $response // {} );
-      Future->done( $request );
+
+      return unless $body->{notification}{type};
+      return unless $body->{notification}{type} eq "m.room.message";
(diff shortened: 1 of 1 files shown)


tests/61push/08_rejected_pushers.pl
```diff
@@ -74,45 +72,50 @@ sub wait_for_push
...
-              uri     => "/r0/pushers",
-         )->then( sub {
-            my ( $body ) = @_;
-
-            assert_json_keys( $body, qw( pushers ) );
-            @{ $body->{pushers} } == 2 or die "Expected two pushers";
-
-            Future->done(1);
-         });
+         retry_until_success {
```
richvdh: why is this needed? Don't the update and the GET both hit the master?

tests/61push/08_rejected_pushers.pl
```diff
@@ -23,17 +23,15 @@ sub wait_for_push
...
-     my $body = $request->body_from_json;
-
-     return unless $body->{notification}{type};
-     return unless $body->{notification}{type} eq "m.room.message";
-     return 1;
-   })->then( sub {
       my ( $request ) = @_;
+      my $body = $request->body_from_json;
 
+      # Respond to all requests, even if we filiter them out
```
richvdh:
```suggestion
      # Respond to all requests, even if we filter them out
```

tests/61push/08_rejected_pushers.pl
erikjohnston: Currently, yes, but there's little reason for that so I figured I may as well change it to not assume that while I'm there anyway.

tests/61push/08_rejected_pushers.pl
erikjohnston: Where has my spell checker gone :/

richvdh: apparently he lives in Cambridge :angry:

PR #589, opened 2019-03-27
Fix race when calling /messages in worker mode

erikjohnston:
To be honest, I'm a bit confused as to how this raced since `/messages` is still on master


tests/30rooms/13guestaccess.pl
```diff
@@ -37,28 +37,31 @@
       })->then( sub {
          matrix_send_room_text_message( $guest_user, $room_id, body => "sup" );
       })->then( sub {
-         matrix_get_room_messages( $user, $room_id, limit => 1 );
-      })->then( sub {
-         my ( $body ) = @_;
-         log_if_fail "Body:", $body;
+         retry_until_success {
```
richvdh: comment might be helpful.

PR #597, opened 2019-04-02
tests for cross-signing

uhoreg:
corresponds to https://github.com/matrix-org/synapse/pull/4970


tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -0,0 +1,642 @@
...
+              # private key: 2lonYOM6xYKdEsO+6KrC766xBcHnYnim1x/4LFGF8B0
+              "user_id" => $user_id,
+              "usage" => ["master"],
+              "keys" => {
+                  "ed25519:nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk"
+                      => "nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk",
+              },
+          },
+      })->main::expect_http_401->then( sub {
+         # uploading a self-signing key with no master key
```
richvdh: can you split this to a separate test case?

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -0,0 +1,642 @@
...
+Get a user's keys
+
+=cut
+
+sub matrix_get_keys {
+   my ( $from_user, $target_user_id ) = @_;
+
+   do_request_json_for( $from_user,
+       method  => "POST",
+       uri     => "/unstable/keys/query",
```
richvdh: this should be r0 (likewise below)

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -0,0 +1,642 @@
...
+
+=head2 matrix_get_keys
+
+   matrix_get_keys( $user, $keys )
+
+Get a user's keys
+
+=cut
+
+sub matrix_get_keys {
```
richvdh: this isn't device-keys specific, so probably belongs in `tests/41end-to-end-keys/01-upload-key.pl`

uhoreg:
Probably needs re-review since I've made some more invasive changes since the last review.

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -162,28 +144,48 @@
...
    # TODO(paul): I don't really know what's parametric about this
    my ( $user, %params ) = @_;
 
    my $device_keys = $params{device_keys} // {};
+   $device_keys->{user_id} = $user->user_id;
+   $device_keys->{device_id} = $user->device_id;
 
    do_request_json_for( $user,
       method => "POST",
       uri    => "/unstable/keys/upload",
```
richvdh: make this be /r0/ while you're here?

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -162,28 +144,48 @@
...
-push our @EXPORT, qw( matrix_put_e2e_keys );
+push our @EXPORT, qw( matrix_put_e2e_keys matrix_get_e2e_keys );
 
 sub matrix_put_e2e_keys
 {
    # TODO(paul): I don't really know what's parametric about this
    my ( $user, %params ) = @_;
 
    my $device_keys = $params{device_keys} // {};
+   $device_keys->{user_id} = $user->user_id;
```
richvdh: I'm not entirely comfortable with modifying the input like this. maybe copy the hash before adding entries?

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -162,28 +144,48 @@
...
       method => "POST",
       uri    => "/unstable/keys/upload",
 
       content => {
-         device_keys => {
-            user_id => $user->user_id,
-            device_id => $user->device_id,
-            device_keys => $device_keys,
-         },
+         device_keys => $device_keys,
```
richvdh: looks like you're changing the API for this function, but haven't changed all the placed it is used?

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -70,29 +69,38 @@
...
-                 # private key: 2lonYOM6xYKdEsO+6KrC766xBcHnYnim1x/4LFGF8B0
-                 "user_id" => $user_id,
-                 "usage" => ["self_signing"],
-                 "keys" => {
-                     "ed25519:nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk"
-                         => "nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk",
-                 },
-             },
-         });
+         Future->done( 1 );
```
richvdh: I think you can safely omit this whole `then` block.

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -70,29 +69,38 @@
...
+              # private key: 2lonYOM6xYKdEsO+6KrC766xBcHnYnim1x/4LFGF8B0
+              "user_id" => $user_id,
+              "usage" => ["self_signing"],
+              "keys" => {
+                  "ed25519:nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk"
+                      => "nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk",
+              },
+          },
       })->main::expect_http_400->then( sub {
          Future->done( 1 );
```
richvdh: ditto

tests/41end-to-end-keys/01-upload-key.pl
uhoreg: AFAICT, other than the new uses in the cross-signing test, the only time the `device_keys` is specified is adding a dummy value to make synapse think that there's a device update.  This is based on the fact that the spec doesn't define a nested `device_keys` property, nor an `updated` property within `device_keys`.  So I don't think that the changed API makes a difference.  (I re-ran the tests, and they seem to pass.)

tests/41end-to-end-keys/01-upload-key.pl
uhoreg: Fixed.  I think I've done that correctly, but someone who actually knows Perl may want to confirm.

tests/41end-to-end-keys/01-upload-key.pl
richvdh: fair enough!

tests/41end-to-end-keys/01-upload-key.pl
richvdh: looks about right!

PR #601, opened 2019-04-03
Fix up rejected pusher tests and add helper functions

erikjohnston:
The sytest HTTP server will drop incoming requests on the floor if no tests are actively awaiting on the path, which is problematic for things like pushers where synapse waits for previous requests to complete before attempting new ones. This leads to races where a request is sent before/after the test has called `await_http_request`. To fix this, we add a concept of "default responders", i.e. handlers that will get called for registered paths when no tests are currently awaiting.

We also add a `wait_for_while` helper function that repeatedly calls a function until a future resolves, making it easier to e.g. keep sending messages until the sytest receives a push

### DIFF
--- a/tests/01http-server.pl
+++ b/tests/01http-server.pl
@@ -9,7 +9,9 @@ my $DIR = dirname( __FILE__ );
 
 struct Awaiter => [qw( pathmatch filter future )];
 
-push our @EXPORT, qw( ServerInfo await_http_request TEST_SERVER_INFO );
+struct DefaultResponder => [qw( pathmatch responder )];
+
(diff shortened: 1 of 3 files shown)


tests/01http-server.pl
```diff
@@ -108,6 +110,8 @@
 # List of Awaiter structs
 my @pending_awaiters;
 
+my @default_responders;
```
richvdh: should this not be scoped to the HTTPServer instance?

erikjohnston: I guess??? The awaiters aren't but I don't know why

PR #602, opened 2019-04-03
Test for not kicking users who aren't in a room

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/4999

### DIFF
--- a/tests/30rooms/15kick.pl
+++ b/tests/30rooms/15kick.pl
@@ -39,16 +39,13 @@ test "Users cannot kick users who have already left a room",
                     state_key => $kicked_user->user_id,
                 )->then( sub {
                     my ( $body ) = @_;
-                    $body->{membership} eq "leave" or die "";
+                    $body->{membership} eq "leave" or
+                        die "Expected kicked user membership to be 'leave'";
 
                     Future->done ( 1 );
                 })
             }
         })->then( sub {
-            my ( $body ) = @_;
-            $body->{membership} eq "leave" or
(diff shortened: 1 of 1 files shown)


tests/30rooms/15kick.pl
```diff
@@ -0,0 +1,54 @@
...
+            matrix_get_room_state( $creator, $room_id,
+                type      => "m.room.member",
+                state_key => $kicked_user->user_id,
+            )
+        })->then( sub {
+            my ( $body ) = @_;
+            $body->{membership} eq "leave" or
+                die "Expected kicked user membership to be 'leave'";
+
+            retry_until_success {
```
erikjohnston: You probably want the `retry_until_success` to cover the checking of the room state, rather than the kicking

tests/30rooms/15kick.pl
```diff
@@ -0,0 +1,54 @@
...
+           uri    => "/r0/rooms/$room_id/kick",
+
+           content => { user_id => $kicked_user->user_id, reason => "testing" },
+        )->then( sub {
+            retry_until_success {
+                matrix_get_room_state( $creator, $room_id,
+                    type      => "m.room.member",
+                    state_key => $kicked_user->user_id,
+                )
+            }
```
erikjohnston: This will always succeed, as it will either return the join or the leave. I'd add 

```perl
)->then( sub {
   my ( $body ) = @_;
   $body->{membership} eq "leave" or die "";

   Future->done ( 1 );
})
```

in the retry loop too

PR #605, opened 2019-04-26
Don't disable federation when using SQLite

babolivier:
Supports config changes for https://github.com/matrix-org/synapse/pull/5078


richvdh:
closing for now as per https://github.com/matrix-org/synapse/pull/5078

PR #611, opened 2019-05-22
Install go in dendrite docker file, remove extra dep from synapse

anoadragon453:
Dendrite's docker file didn't have `go` installed which is necessary for CircleCI on Dendrite's repo. It also had an incorrect path for the `tap-to-junit-xml.pl` script.

Additionally `dos2unix` was being installed in the `sytest` docker image, and then again in the `sytest-synapse*` images, so that was removed.

### DIFF
--- a/docker/Dockerfile-dendrite
+++ b/docker/Dockerfile-dendrite
@@ -1,6 +1,12 @@
 FROM matrixdotorg/sytest:latest
 
-RUN apt-get -qq install -y dos2unix
+# Install Go 1.11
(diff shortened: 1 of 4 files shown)


anoadragon453:
You *can* do it with [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) apparently, but looks like a lot more faff that just updating the golang URL when we want to update.

anoadragon453:
Ftr we're using go 1.11 for CI as it's our lowest supported go version.

PR #613, opened 2019-05-27
Test for correct error when logging out with invalid or missing access token

aaronraimist:
For https://github.com/matrix-org/synapse/pull/5256


erikjohnston:
Wooooo

PR #615, opened 2019-05-28
Add regression tests for synapse#5269

erikjohnston:
c.f. https://github.com/matrix-org/synapse/pull/5274

### DIFF
--- a/tests/50federation/52soft-fail.pl
+++ b/tests/50federation/52soft-fail.pl
@@ -323,3 +323,214 @@ test "Inbound federation accepts a second soft-failed event",
          );
       });
    };
+
+
+test "Inbound federation correctly handles soft failed events as extremities",
+   # this is mostly a regression test for https://github.com/matrix-org/synapse/issues/5269.
+   requires => [
+      $main::OUTBOUND_CLIENT, $main::INBOUND_SERVER, $main::HOMESERVER_INFO[0],
+      local_user_and_room_fixtures(
+         user_opts => { with_events => 1 },
+      ),
+      federation_user_id_fixture(),
+   ],
+
+   do => sub {
+      my (
+         $outbound_client, $inbound_server, $info, $creator, $room_id,
+         $remote_user_id,
+      ) = @_;
+      my $first_home_server = $info->server_name;
+
+      my $local_server_name = $outbound_client->server_name;
+
+      my $room;
+
+      # We're going to construct a room graph like:
+      #
+      #        J1
+      #       /  \
+      #      /    \
+      #    PL1     M1
+      #     |      |
+      #     |     SF2
+      #     |      |
+      #      \    SF1
+      #       \   /
+      #         M2
+      #         |
+      #         M3
+      #
+      # Where time is flowing downwards.
+      #
+      #  J1  = join of remote user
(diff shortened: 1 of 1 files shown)


tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      #
+      #  J1  = join of remote user
+      #  PL1 = creator of room blocks SF event sending
+      #  M1  = remote user sends a permitted message
+      #  SF1 = remote user sends a soft-failed message
+      #  SF2 = remote user sends a soft-failed message
+      #  M2  = remote user sends a permitted message
+      #  M3  = creator sends a message
+      #
+       # Since the banning of SF events happens before SF1 and SF2, we expect the
```
richvdh:
```suggestion
      # Since the banning of SF events happens before SF1 and SF2, we expect the
```

tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      # local server to soft fail SF1 and SF2 when they are received.
+      #
+      # We should therefore end up with M2 as the forward-extremitiy of the
```
richvdh: There's a leap of logic missing here:

```suggestion
      # M1 and PL1 will therefore become forward-extremities of the room. However, M2 will
      # be accepted, and replace M1 and PL1 as the extremities of the room. M2 should 
      # therefore be the sole prev_event of M3.
```

tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      # room, and hence the prev_event of M3.
```
richvdh:
```suggestion
      # room, and hence the prev_event of M3.
```

tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      my $room;
+
+      # We're going to construct a room graph like:
+      #
+      #        J1
+      #       /  \
+      #      /    \
+      #    PL1     M1
+      #     |      |
+      #     |     SF2
```
richvdh:
```suggestion
      #     |     SF1
```

tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      #     |      |
+      #      \    SF1
```
richvdh:
```suggestion
      #      \    SF2
```

tests/50federation/52soft-fail.pl
```diff
@@ -323,3 +323,214 @@
...
+      #         |
+      #         M3
+      #
+      # Where time is flowing downwards.
+      #
+      #  J1  = join of remote user
+      #  PL1 = creator of room blocks SF event sending
+      #  M1  = remote user sends a permitted message
+      #  SF1 = remote user sends a soft-failed message
+      #  SF2 = remote user sends a soft-failed message
```
richvdh: I think SF2 is unnecessary in this test? you could get rid of it, if you could be bothered.

erikjohnston: Yes, ish, though it validates the logic a bit better since the test then checks the correct handling of chains of soft failed events.

PR #621, opened 2019-06-05
Use CA-signed TLS keys for testing

richvdh:
Now that synapse requires real certs, we should present real certs.

We could just turn off the cert validation in synapse, but it seems nicer
to use a fake CA.

The CA key/cert were generated with:

   openssl genrsa -out ca.key 2048
   openssl req -new -x509 -key ca.key -days 3650 -out ca.crt

We could generate the CA cert and key dynamically, but it's easier to store
them.


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -145,16 +147,20 @@ sub start
...
         ( -f $log_config_file ) ? ( log_config => $log_config_file ) : (),
-        tls_certificate_path => "$cwd/keys/tls-selfsigned.crt",
-        tls_private_key_path => "$cwd/keys/tls-selfsigned.key",
-        tls_dh_params_path => "$cwd/keys/tls.dh",
+        tls_certificate_path => $self->{paths}{cert_file},
+        tls_private_key_path => $self->{paths}{key_file},
+        federation_custom_ca_list => [
+           "$cwd/keys/ca.crt",
+        ],
         use_insecure_ssl_client_just_for_testing_do_not_use => 1,
```
erikjohnston: Can we not turn this off then? Or what? I'm a bit confused about what has real certs and what doesn't.

richvdh: this setting controls synapse's 'simple http client', which is used for non-federation requests, including (it turns out) some to IS servers. There's no way to set a custom CA for that, hence the setting.

This should probably be in a comment in the sytest code...

PR #627, opened 2019-06-19
Read in white/blacklist test files as utf-8

anoadragon453:
Was causing characters such as `£` to transform into another character, thus meaning tests that contained those characters would not be considered part of the testfile.


ara4n:
lgtm fwiw

PR #628, opened 2019-06-21
Route pagination requests to client reader

erikjohnston:


tests/30rooms/04messages.pl
```diff
@@ -352,3 +352,39 @@
          })
       });
    };
+
+
+test "/messages does not require a from param",
```
richvdh: https://matrix.org/docs/spec/client_server/r0.5.0#get-matrix-client-r0-rooms-roomid-messages says it does.

richvdh: right now, the fact that synapse doesn't return a 400 when this happens is a synapse bug. If we want to go the other way, it'll need an MSC defining what it means to not give a `from` param...

erikjohnston: Ugh. Ok, I just wanted a regression test but it doesn't seem obvious to me we should support not supplying a `from`

PR #631, opened 2019-06-26
Have the room member tests retry a few times

anoadragon453:
These tests were pretty flaky in a multi-process homeserver (like Dendrite). The events took a few moments to persist, and thus a race condition was produced where SyTest would request the status right afterwards (before the events finished persisting).

We now retry a few times before calling it failed.

(Should `retry_until_success` be moved inside of `matrix_get_room_state`?)

### DIFF
--- a/tests/10apidoc/33room-members.pl
+++ b/tests/10apidoc/33room-members.pl
@@ -35,23 +35,21 @@ test "POST /rooms/:room_id/join can join a room",
          $body->{room_id} eq $room_id or
             die "Expected 'room_id' to be $room_id";
 
-         Future->done(1);
-      });
-   },
+         # Retry getting the state a few times, for it may take some time to
+         # propagate in a multi-process homeserver
+         retry_until_success {
+            matrix_get_room_state( $user, $room_id,
+               type      => "m.room.member",
+               state_key => $user->user_id,
+            )->then( sub {
+               my ( $body ) = @_;
 
-   check => sub {
-      my ( $user, $room_id, undef ) = @_;
+               $body->{membership} eq "join" or
+                  die "Expected membership to be 'join'";
 
-      matrix_get_room_state( $user, $room_id,
-         type      => "m.room.member",
-         state_key => $user->user_id,
-      )->then( sub {
-         my ( $body ) = @_;
-
-         $body->{membership} eq "join" or
-            die "Expected membership to be 'join'";
-
-         Future->done(1);
+               Future->done(1);
+            })
+         }
       });
    };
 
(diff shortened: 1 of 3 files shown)


tests/10apidoc/33room-members.pl
```diff
@@ -178,10 +172,14 @@ sub matrix_join_room
...
-         )
+         # Retry getting the state a few times, for it may take some time to
+         # propagate in a multi-process homeserver
+         retry_until_success {
+            matrix_get_room_state( $user, $room_id,
+               type      => "m.room.member",
+               state_key => $user->user_id,
+            )
+         }
       })->then( sub {
```
richvdh: should this not be inside the `retry_until_success` call?

tests/10apidoc/33room-members.pl
```diff
@@ -35,23 +35,21 @@
...
-         type      => "m.room.member",
-         state_key => $user->user_id,
-      )->then( sub {
-         my ( $body ) = @_;
-
-         $body->{membership} eq "join" or
-            die "Expected membership to be 'join'";
-
-         Future->done(1);
+         # Retry getting the state a few times, for it may take some time to
```
richvdh: this reads really awkwardly.

```suggestion
         # Retry getting the state a few times, as it may take some time to
```
or 'since'?

tests/10apidoc/33room-members.pl
richvdh: (likewise various other places below)

erikjohnston:
> I'm not quite sure what I think about this. Tightlooping on get_room_state doesn't seem optimal, though I guess it's ok.

FWIW it does have a backoff delay, so it doesn't hammer tooooo much

> In other places I think we wait for the change to come down `/sync`

We do tend to use `*_synced` as they're just more convenient, though there's nothing actually that says if the event has made it to the sync servers its made it to the necessary other workers....

tests/10apidoc/33room-members.pl
anoadragon453: The whole subroutine?

anoadragon453:
Fixed up with advice from rich [in #synapse-dev](https://matrix.to/#/!yZHTGeDKZUeKaqeTeU:matrix.org/$1561714939182466yLcUv:amorgan.xyz?via=matrix.org&via=vector.modular.im&via=amorgan.xyz).

tests/10apidoc/33room-members.pl
richvdh: this still says 'for' ?

tests/10apidoc/36room-levels.pl
```diff
@@ -56,7 +56,9 @@
             content => $levels,
          )
       })->then( sub {
-         matrix_get_room_state( $user, $room_id, type => "m.room.power_levels" )
+         retry_until_success {
+            matrix_get_room_state( $user, $room_id, type => "m.room.power_levels" )
+         }
       })->then( sub {
```
richvdh: needs to be in the `retry_until_success` block

PR #632, opened 2019-06-26
Value for "typing" in /typing requests should be bool instead of int

Cnly:
Spec: https://matrix.org/docs/spec/client_server/r0.5.0#put-matrix-client-r0-rooms-roomid-typing-userid

Signed-off-by: Alex Chen <minecnly@gmail.com>


richvdh:
looks about right.

PR #638, opened 2019-07-03
Remove "bug" & "fixed" parameters, and reimplement functionality with a blacklist

anoadragon453:
Sytest has a feature that allows you to mark a test as expected fail, along with a bug ID with the reasoning for it. This is nice, but really only works in the context of a single homeserver, where tests were being disabled for everyone because something didn't work in Synapse.

Since homeserver-specific test white/blacklist was added, we can recreate the functionality of the `bug` parameter using a blacklist instead.

This PR pairs with https://github.com/matrix-org/synapse/pull/5611

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -55,8 +55,6 @@ our $BIND_HOST = "localhost";
 # and the like.
 our $TEST_RUN_ID = strftime( '%Y%m%d_%H%M%S', gmtime() );
 
-my %FIXED_BUGS;
-
 my $STOP_ON_FAIL;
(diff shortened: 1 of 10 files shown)


run-tests.pl
```diff
@@ -107,8 +107,6 @@
 
    'p|port-range=s' => \(my $PORT_RANGE = "8800:8899"),
 
-   'F|fixed=s' => sub { $FIXED_BUGS{$_}++ for split m/,/, $_[1] },
```
richvdh: need to remove the declaration of `%FIXED_BUGS`

PR #639, opened 2019-07-04
Update the SyTest docker images to ones that work on Buildkite

hawkowl:


docker/Dockerfile-synapsepy35
```diff
@@ -10,7 +10,7 @@ RUN su -c '/usr/lib/postgresql/9.6/bin/initdb -E "UTF-8" --lc-collate="en_US.UTF
 
 # Turn off all the fsync stuff for postgres
 RUN mkdir -p /etc/postgresql/9.6/main/conf.d/
-RUN echo "fsync=off" > /etc/postgresql/9.6/main/conf.d/fsync.conf
+RUN echo "fync=off" > /etc/postgresql/9.6/main/conf.d/fsync.conf
```
anoadragon453: Are you sure this is right?

hawkowl: no i am not

richvdh:
what's the background here? why don't they currently work on buildkite? Are we not currently using them on buildkite?

hawkowl:
@richvdh when I first got it working, I vendored in a buildkite-specific `build_synapse.sh`. This upstreams that, and fixes some issues (like us using entrypoints instead of cmd which makes it harder for the BK plugin to use).

These are the current images being used on buildkite.

docker/synapse_sytest.sh
```diff
@@ -5,6 +5,13 @@
 
 set -ex
 
+if [ -n "$BUILDKITE" ]
+then
+    SYNAPSE_DIR=`pwd`
```
richvdh: could we not set `workdir` in the buildkite config (https://github.com/buildkite-plugins/docker-buildkite-plugin#workdiroptional-string) which would solve this and make the readme (which says: mount the server to be tested at /src) more truthful?

hawkowl: see https://github.com/matrix-org/synapse/pull/5664

richvdh: so we can get rid of this condition now?

hawkowl: yes

PR #644, opened 2019-07-08
Create new SSL certs for each test run

richvdh:
Fixes a problem where the sytests would suddenly start failing because they
were still using the certs taht we generated a month ago.


richvdh:
thanks

PR #651, opened 2019-07-17
Remove trailing slashes from some requests

anoadragon453:
Dendrite is currently quite strict in that it doesn't allow trailing slashes anywhere that the spec does not supply them (this is expected to become more lax in the future but doing so elegantly is currently a bit tricky with our HTTP library). 

In the meantime, we should update SyTest to follow the spec with regards to trailing slashes anyways.

Synapse is still happy: https://buildkite.com/matrix-dot-org/synapse/builds/2870

### DIFF
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -123,7 +123,7 @@ sub send_transaction
    $self->do_request_json(
       method   => "PUT",
       hostname => $params{destination},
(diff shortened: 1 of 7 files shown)


richvdh:
> this is expected to become more lax in the future

Don't make it more lax - accepting unspecced endpoints is a bug, not a feature!

anoadragon453:
Fair enough :D

PR #659, opened 2019-07-23
UISI test

JorikSchellekens:
Tests for https://github.com/matrix-org/synapse/issues/5441


erikjohnston:
The tests are broadly good, but the problem with killing one of the servers is that they are used for the entire test run, rather than being set up for each test. I think a good alternative here is to use the sytest server implementation and then we can simulate the server "going down" by having it return e.g. a 503.

JorikSchellekens:
The server is restarted at the end. Does the sytest server implementation emulate users actions such as joining rooms and uploading keys? Otherwise this test requires a lot of mocking.

erikjohnston:
> Does the sytest server implementation emulate users actions such as joining rooms and uploading keys? Otherwise this test requires a lot of mocking.

Yup, look in `test/50federation/` which all use the perl sytest federation thing.

JorikSchellekens:
I've changed it over to the sytest federation server

tests/50federation/40devicelists.pl
```diff
@@ -284,3 +284,132 @@
...
+            room_version => "5",
+         )->then( sub {
+            my ( $room_id ) = @_;
+            $outbound_client->join_room(
+               server_name => $local_server_info->server_name,
+               room_id     => $room_id,
+               user_id     => $outbound_client_user,
+            )
+         })->then( sub {
+            do_request_json_for( $local_user,
```
erikjohnston: Can you just add some comments at the steps here please, e.g. here we do "First we successfully request the keys while the remote server is up", etc.

tests/50federation/40devicelists.pl
```diff
@@ -284,3 +284,132 @@
...
+            }),
+            $inbound_server->await_request_user_keys_query()->then( sub {
+               my ( $req ) = @_;
+               $OUTPUT->diag("responding");
+               $req->respond_json({}, code => 503);
+               Future->done(1)
+            })
+         );
+         Future->wait_any(
+            @respond_503,
```
erikjohnston: Won't `Future->wait_any` return here when one of the above `await_*` returns? I think you probably want to not wait on them and just return the `do_request_json`?

tests/05homeserver.pl
```diff
@@ -4,7 +4,7 @@
 
 my $N_HOMESERVERS = 2;
 
-my @servers;
+our @SERVERS;
```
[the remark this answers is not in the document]
JorikSchellekens: Right

tests/50federation/40devicelists.pl
JorikSchellekens: Rookie question: if I do so how do I start the two futures I've created using await*?

tests/05homeserver.pl
```diff
@@ -135,4 +135,4 @@
          })
       },
    );
-} 0 .. $N_HOMESERVERS-1;
+} 0 .. $N_HOMEservers-1;
```
erikjohnston: Can we just not change this file? :)

tests/50federation/40devicelists.pl
```diff
@@ -284,3 +284,125 @@
...
+         );
+         do_request_json_for( $local_user,
+            method  => "POST",
+            uri     => "/r0/keys/query",
+            content => {
+               device_keys => {
+                  $outbound_client_user => {}
+               }
+            }
+         )->then( sub {
```
erikjohnston: I would probably move this up a level so that we don't have nested callbacks

Issue #653, opened 2019-07-19
SyTest asserts non-required JSON key "unsigned" in /event/ tests


Cnly:
Found https://github.com/matrix-org/sytest/blob/19abb6046614bbbeda8dd5153f00a56794c3e2c4/tests/30rooms/51event.pl#L41 and https://github.com/matrix-org/sytest/blob/19abb6046614bbbeda8dd5153f00a56794c3e2c4/tests/30rooms/51event.pl#L115.

---

changed links to permalinks ~anoa

PR #668, opened 2019-07-31
Don't require unsigned for /rooms/{roomId}/event/{eventId}

anoadragon453:
Fixes https://github.com/matrix-org/sytest/issues/653

The spec [doesn't mention](https://matrix.org/docs/spec/client_server/unstable#get-matrix-client-r0-rooms-roomid-event-eventid) `unsigned`, so we shouldn't require it here.

### DIFF
--- a/tests/30rooms/51event.pl
+++ b/tests/30rooms/51event.pl
@@ -38,7 +38,7 @@ test "/event/ on joined room works",
          )->then( sub {
             my ( $body ) = @_;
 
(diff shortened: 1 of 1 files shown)


turt2live:
The spec does mention it elsewhere btw, but it shouldn't be considered a required field.

erikjohnston:
> The spec doesn't mention unsigned, so we shouldn't require it here.

Hmm, that spec should really reference the event definitions.

PR #669, opened 2019-07-31
Add test that tombstone cannot reference itself

erikjohnston:
Tests https://github.com/matrix-org/synapse/pull/5801


tests/30rooms/60version_upgrade.pl
```diff
@@ -811,7 +811,22 @@ sub upgrade_room_synced {
       )->main::expect_matrix_error( 404, 'M_NOT_FOUND' );
    };
 
-# upgrade with other local users
-# upgrade with remote users
+test "Cannot send tombstone event that point to the same room.",
```
anoadragon453:
```suggestion
test "Cannot send tombstone event that points to the same room",
```

PR #679, opened 2019-08-22
Add v2 lookup support to sytest identity server

anoadragon453:
Adds the `/v2/hash_details` and `/v2/lookup` endpoints to the built-in sytest identity server.

Synapse PR: https://github.com/matrix-org/synapse/pull/5897


anoadragon453:
Oops, hold up. Found an edge case.

anoadragon453:
Ok, all better now.

PR #680, opened 2019-08-30
Remove disused Jenkins scripts

hawkowl:
burrrrrrrrrrrrrrrrrrrrrrrrrrrrrn

also see https://github.com/matrix-org/synapse/pull/5938/


richvdh:
(though what's with the broken CI?)

PR #681, opened 2019-08-30
CI for SyTest

hawkowl:


hawkowl:
(merging unilaterally as it's CI junk)

PR #682, opened 2019-08-30
Have the in-built identity server support v2

anoadragon453:
Add all of the v2 identity server endpoints to Sytest's built-in identity server. This is independent of any of the Synapse changes currently in-flight.

Additionally, change the identity server tests to use these v2 endpoints.

This essentially just involves switching the endpoints to `v2` and including an `id_access_token` parameter at the top-level of the JSON, or in a query parameter.

I tried to reuse functions wherever possible.


anoadragon453:
Closing this and moving it to a different branch, as Synapse currently has a `anoa/msc2140` branch as well.

PR #684, opened 2019-09-02
Fix event id calculation

richvdh:
Fix the event id calculated for events in v3 etc rooms: we have to calculate
the hash ~~*before*~~ *after* signing.

Turns out this was working by good fortune before (we never actually checked
the event ids).

### DIFF
--- a/lib/SyTest/Federation/Datastore.pm
+++ b/lib/SyTest/Federation/Datastore.pm
@@ -190,6 +190,7 @@ sub create_event
       for qw( type auth_events content depth prev_events room_id sender );
 
    my $room_version = delete $fields{room_version} // 1;
+   my $event_id_suffix = delete $fields{event_id_suffix};
 
    my $event = {
       %fields,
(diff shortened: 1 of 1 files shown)


lib/SyTest/Federation/Datastore.pm
```diff
@@ -190,27 +190,30 @@ sub create_event
...
-         # calculate the event id from the hash
-         $event_id = id_for_event( $event, $room_version );
       }
+      $self->sign_event( $event );
+   } else {
+      die "event with explicit event_id in room v$room_version"
+         if defined $event_id;
+
+      $self->sign_event( $event );
+      $event_id = id_for_event( $event, $room_version );
```
erikjohnston: Isn't this now calculating the hash after signing?

richvdh: that's what I (wish I) said

erikjohnston: Why does it matter if we hash before or after we sign? The signatures should be stripped before we calculate the hash?

richvdh: yes, but the `hashes` aren't.

erikjohnston: Oh, and `sign_event` also adds the hash. Right, got you. So this basically says we need to calculate the event ID hash *after* we've added the content hashes.

PR #686, opened 2019-09-02
Optionally return event id from create_and_insert_event

richvdh:
Return the event id as well as the event if it's called in array context;
that's more convenient than having to call id_for_event separately.

### DIFF
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -225,7 +225,11 @@ sub create_event
 
 =head2 create_and_insert_event
 
-   $event = $room->create_and_insert_event( %fields )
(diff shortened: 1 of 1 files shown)


lib/SyTest/Federation/Room.pm
```diff
@@ -237,13 +241,14 @@ sub create_and_insert_event
...
 
-   my $event = $self->create_event( %fields );
+   my ( $event, $event_id ) = $self->create_event( %fields );
 
    $self->_insert_event( $event );
 
    $self->{prev_events} = [ $event ];
 
-   return $event;
+   return $event unless wantarray;
```
erikjohnston: TIL

PR #689, opened 2019-09-03
Have the in-built identity server support v2

anoadragon453:
Add all of the v2 identity server endpoints to Sytest's built-in identity server. This is independent of any of the Synapse changes currently in-flight.

Additionally, change the identity server tests to use these v2 endpoints.

This essentially just involves switching the endpoints to v2 and including an `id_access_token` parameter at the top-level of the JSON, or in a query parameter. I tried to reuse functions wherever possible.

The `id_access_token` is hard-coded, as there isn't much point retrieving it via OpenID as we're using a built-in Sytest identity server, which we don't need to test. Sydent's OpenID functionality is handled by matrix-is-tester.

### DIFF
--- a/lib/SyTest/Identity/Server.pm
+++ b/lib/SyTest/Identity/Server.pm
@@ -223,12 +223,14 @@ sub validate_identity
 {
    my $self = shift;
    my ( $medium, $address, $client_secret ) = @_;
+
    my $sid = "session_${\ $self->{sid}++ }";
-   $self->on_pubkey( $req, $key_name );
+
    $self->{validated}{$sid} = {
       medium       => $medium,
       address      => $address,
    };
+
    return $sid;
 }
 
--- a/lib/SyTest/Identity/Server.pm
+++ b/lib/SyTest/Identity/Server.pm
@@ -17,7 +17,10 @@ my $crypto_sign = Crypt::NaCl::Sodium->sign;
 
 my $next_token = 0;
 
-my $id_access_token = "testing";
+# Perpetually correct access token for authenticating with v2 Identity Service API endpoints.
+# v2 endpoint calls to this identity server should include this value for their
+# `id_access_token` parameter
+my $ID_ACCESS_TOKEN = "swordfish";
 
 sub _init
 {
@@ -44,6 +47,14 @@ sub _init
    return $self->SUPER::_init( @_ );
 }
 
+=head2
+
+   $self->rotate_keys();
+
+Creates new ed25519 public/private key pairs for this server.
+
+=cut
+
 sub rotate_keys
 {
    my $self = shift;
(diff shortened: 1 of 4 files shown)


anoadragon453:
Added docstrings for every function.

lib/SyTest/Identity/Server.pm
```diff
@@ -17,7 +17,10 @@ my $crypto_sign = Crypt::NaCl::Sodium->sign;
 
 my $next_token = 0;
 
-my $id_access_token = "testing";
+# Perpetually correct access token for authenticating with v2 Identity Service API endpoints.
+# v2 endpoint calls to this identity server should include this value for their
+# `id_access_token` parameter
+my $ID_ACCESS_TOKEN = "swordfish";
```
richvdh: :heart:

lib/SyTest/Identity/Server.pm
```diff
@@ -42,6 +47,14 @@ sub _init
    return $self->SUPER::_init( @_ );
 }
 
+=head2
```
richvdh:
```suggestion
=head2 rotate_keys
```

likewise for a bunch of the others.

btw in case you weren't aware: you need `perldoc lib/SyTest/Identity/Server.pm` to see the output

lib/SyTest/Identity/Server.pm
```diff
@@ -42,6 +47,14 @@ sub _init
...
+
+   $self->rotate_keys();
```
richvdh: $self is unconventional here

```suggestion
   $server->rotate_keys();
```

likewise for the others

lib/SyTest/Identity/Server.pm
```diff
@@ -55,232 +68,527 @@ sub rotate_keys
    };
 }
 
+=head2
```
richvdh:
```suggestion
=head2 on_request

    $server->on_request( $request );
```

lib/SyTest/Identity/Server.pm
```diff
@@ -55,232 +68,527 @@ sub rotate_keys
...
 
-            $req->respond_json( \%resp, code => 400 );
-            return;
-         }
+   if (
+      $req->query_param("id_access_token") and
+      $req->query_param("id_access_token") eq $ID_ACCESS_TOKEN
+   ) {
+      # We found it!
+      return 1
```
richvdh:
```suggestion
      return 1;
```

lib/SyTest/Identity/Server.pm
```diff
@@ -55,232 +68,527 @@ sub rotate_keys
...
+   # the request has a body
+   my $body = eval { $req->body_from_json };
+
+   if (
+      $body and
+      $body->{id_access_token} and
+      $body->{id_access_token} eq $ID_ACCESS_TOKEN
+   ) {
+      # We found it!
+      return 1
```
richvdh:
```suggestion
      return 1;
```

lib/SyTest/Identity/Server.pm
```diff
@@ -55,232 +68,527 @@ sub rotate_keys
...
-      }
-      else {
-         # Unknown algorithm provided
-         $resp{error} = "Unknown algorithm";
-         $resp{errcode} = "M_INVALID_PARAM";
+   # Couldn't find an access token
+   $resp{error} = "Missing id_access_token parameter";
+   $resp{errcode} = "M_MISSING_PARAM";
+   $req->respond_json( \%resp, code => 400 );
+   return 0
```
richvdh:
```suggestion
   return 0;
```

lib/SyTest/Identity/Server.pm
```diff
@@ -57,111 +68,133 @@ sub rotate_keys
...
       # We found it!
-      return
+      return 1
    }
 
    # Check the JSON body for the token. This isn't required for all endpoints so only try if
    # the request has a body
-   my $found = 0;
-   eval {
-      # We use an eval in case this request doesn't have a JSON body
```
richvdh: it's worth keeping this comment

lib/SyTest/Identity/Server.pm
anoadragon453: I was not, thank you!

lib/SyTest/Identity/Server.pm
```diff
@@ -729,16 +730,16 @@ sub invites_for
    return $self->{invites}{ join "\0", $medium, $address };
 }
 
-=head2
+=head2 get_access_token
 
-   $self->get_access_token();
+   $server->get_access_token();
```
richvdh: I'm not going to make you go through all these yet again, but for future reference: this *is* the example, so you should make it look like what you've written under 'Example' and get rid of the separate example.

lib/SyTest/Identity/Server.pm
```diff
@@ -747,9 +748,9 @@ sub get_access_token
    return $ID_ACCESS_TOKEN;
 }
 
-=head2
+=head2 name
 
-   $self->name():
+   $server->name():
```
richvdh: s/:/;/

(fyi: worth noting that parens after a fn call are optional in perl, so you can write this as `$server->name;`)

anoadragon453: mind blown

PR #690, opened 2019-09-04
Test public room list pagination

erikjohnston:


tests/30rooms/70publicroomslist.pl
```diff
@@ -135,3 +135,105 @@
...
+      my ( $user ) = @_;
+
+      my $num_rooms;
+      my $next_batch;
+      my $prev_batch;
+
+      # A hash from room ID to number of times we saw the room.
+      my $counts = {};
+
+      # First we fill up the room list a bit (note there will probably arleady
```
richvdh:
```suggestion
      # First we fill up the room list a bit (note there will probably already
```

tests/30rooms/70publicroomslist.pl
```diff
@@ -135,3 +135,105 @@
...
+
+   check => sub {
+      my ( $user ) = @_;
+
+      my $num_rooms;
+      my $next_batch;
+      my $prev_batch;
+
+      # A hash from room ID to number of times we saw the room.
+      my $counts = {};
```
richvdh: normally you'd use a regular hash rather than a hashref here:

```suggestion
      my %counts;
```

doesn't really matter though

erikjohnston: Originally I did that then everything went horribly wrong and it worked with a hashref and I still don't really understand perl :sob:

PR #697, opened 2019-09-10
Accept access_token instead of id_access_token on the IS, accept Authorization header

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/6013 which has context

We now accept `access_token` instead of `id_access_token` and do so using the `Authorization` header instead of in the JSON body, as [MSC2140](https://github.com/matrix-org/matrix-doc/pull/2140/files#diff-c03a26de5ac40fb532de19cb7fc2aaf7R80) states.

### DIFF
--- a/lib/SyTest/Identity/Server.pm
+++ b/lib/SyTest/Identity/Server.pm
@@ -19,7 +19,7 @@ my $next_token = 0;
 
 # Perpetually correct access token for authenticating with v2 Identity Service API endpoints.
 # v2 endpoint calls to this identity server should include this value for their
-# `id_access_token` parameter
+# `access_token` parameter
 my $ID_ACCESS_TOKEN = "swordfish";
 
 sub _init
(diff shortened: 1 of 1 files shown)


lib/SyTest/Identity/Server.pm
```diff
@@ -137,13 +137,13 @@ sub on_request
...
-the request body.
+A helper method that takes an HTTP request and checks if an C<access_token> parameter
+matching C<$ID_ACCESS_TOKEN> is present in either the query parameters or the Authorization
+header (after the Bearer declaration).
 
-Returns C<0> or C<1> depending on whether a correct C<id_access_token> value was found.
+Returns C<0> or C<1> depending on whether a correct C<access_token> value was found.
 
-Responds to the HTTP request with an error message if no C<id_access_token> value was found.
+Responds to the HTTP request with an error message if no C<access_token> value was found.
```
richvdh: can you fix the text of said error message (it still says `id_access_token`)

anoadragon453:
The tests will fail until we merge https://github.com/matrix-org/synapse/pull/6013 unfortunately.

anoadragon453:
And it's been merged!

PR #698, opened 2019-09-11
add some logging for flaky sync test

richvdh:


richvdh:
I totally tested that... :/

PR #701, opened 2019-09-11
federated_rooms_fixture

richvdh:
A test fixture which creates one or more rooms and joins them over
federation.

We have quite a lot of tests that do this, and the boilerplate is boilerplatey.


richvdh:
sorry about the size of the diff here... it looks worse than it is. It's better if you hide whitespace changes.

PR #704, opened 2019-09-17
Implement a mail server to help with 3pid testing

richvdh:
To help us test the 3pid validation flows, we're going to need a simple SMTP
server. This starts one up, but doesn't do anything with it yet.


lib/SyTest/MailServer.pm
```diff
@@ -0,0 +1,124 @@
+package SyTest::MailServer;
```
richvdh: much of this was cargo-culted from the HTTP equivalents...

PR #705, opened 2019-09-17
Configure synapse to use the test mail server

richvdh:
Tweak the generated synapse config so that it uses the test mail server.

Based on #704

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -142,6 +142,7 @@ sub start
 
    my $listeners = [ $self->generate_listeners ];
    my $bind_host = $self->{bind_host};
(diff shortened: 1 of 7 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -255,6 +257,14 @@ sub start
            recaptcha_private_key    => $self->{recaptcha_config}->{private_key},
         ) : (),
 
+        $self->{smtp_server_config} ? (
+           email => {
+              smtp_host => $self->{smtp_server_config}->{host},
+              smtp_port => $self->{smtp_server_config}->{port},
+              notif_from => 'syanapse@localhost',
```
anoadragon453:
```suggestion
              notif_from => 'synapse@localhost',
```

PR #707, opened 2019-09-18
Add support for handling email validation challenges

richvdh:
As of MSC2078, the homeserver is allowed to send its own email validation when we do a /requestToken, so stubbing it out in the test Identity Server no longer cuts the mustard.

These tests should pass against current develop (and master). `tests/12login/01threepid-and-password.pl` should also pass against `anoa/account_3pid_homeserver_send_email` (https://github.com/matrix-org/synapse/pull/6042), except that there seem to be implementation errors over there which need fixing. `tests/54identity.pl` certainly *won't* pass against that branch, because it tests things that that branch conciously disables.

~~This PR is based on #705 and #706.~~

### DIFF
--- a/tests/12login/01threepid-and-password.pl
+++ b/tests/12login/01threepid-and-password.pl
@@ -45,17 +45,53 @@ sub validate_email_for_user {
          Future->done;
       }),
 
-      # we now expect a callout to our test ID server.
-      await_id_validation( $id_server, $address ),
+      # depending on the server under test, we should expect either a callout to
+      # our test ID server, or an email from the homeserver.
+      #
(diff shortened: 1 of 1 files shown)


richvdh:
I don't think POSTs are supported on the confirmation link?

anoadragon453:
> I don't think POSTs are supported on the confirmation link?

They are :D https://matrix.org/docs/spec/identity_service/unstable#post-matrix-identity-v2-validate-msisdn-submittoken

anoadragon453:
Conclusion: We removed the POST as Synapse doesn't prompt the user to use it yet.

PR #710, opened 2019-09-26
Generalise validate_email_for_user to work for other /requestToken endpoints

anoadragon453:
This came up during implementation of #695. Registration calls the `/r0/register/email/requestToken` path rather than `/account/3pid/email/requestToken`.

Additionally remove the requirement for a $user parameter to be required, as none of the /requestToken endpoints require authentication.


anoadragon453:
I'll leave it out for now, but I'll keep that in mind :)

PR #714, opened 2019-10-04
Add Perl results.tap parser

anoadragon453:
The aim of this is to replace the [format_tap.py](https://github.com/matrix-org/synapse/blob/develop/.buildkite/format_tap.py) file that's required to show the following Buildkite annotation for failed Sytest tests:

image

This perl script generates a line-for-line same output as the python version. This is useful so that we don't need to include a python binary for Dendrite's sytest environment.


format_tap.pl
```diff
@@ -0,0 +1,49 @@
...
+my $parser = TAP::Parser->new( { source => $tap_file } );
+my $in_error = 0;
+my @out = ( "### TAP Output for $build_name" );
+
+while ( my $result = $parser->next ) {
+   if ( $result->is_test ) {
+      # End an existing error block
+      if ( $in_error == 1 ) {
+         push ( @out, "" );
+         push ( @out, "</pre></code></details>" );
```
erikjohnston:
```suggestion
         push( @out, "</pre></code></details>" );
```

format_tap.pl
```diff
@@ -0,0 +1,49 @@
...
+
+my $parser = TAP::Parser->new( { source => $tap_file } );
+my $in_error = 0;
+my @out = ( "### TAP Output for $build_name" );
+
+while ( my $result = $parser->next ) {
+   if ( $result->is_test ) {
+      # End an existing error block
+      if ( $in_error == 1 ) {
+         push ( @out, "" );
```
erikjohnston:
```suggestion
         push( @out, "" );
```

format_tap.pl
```diff
@@ -0,0 +1,49 @@
...
+         my $number = $result->number;
+         my $description = $result->description;
+
+         push(@out, "FAILURE Test #$number: ``$description``");
+         push(@out, "");
+         push(@out, "<details><summary>Show log</summary><code><pre>");
+      }
+   } elsif ( $result->is_comment and $in_error == 1 ) {
+      # Print error contents
+      push ( @out, $result->raw );
```
erikjohnston:
```suggestion
      push( @out, $result->raw );
```

format_tap.pl
```diff
@@ -0,0 +1,49 @@
...
+   }
+}
+
+# Print out the contents of @out, leaving off the last little formatting bits
+foreach my $line ( @out[0..$#out-3] ) {
+   # The -l in the hashbang makes print append a newline to the content
```
erikjohnston: ugh, can't we just add `print \n`? Otherwise running `perl format_tap.pl` does something different.

format_tap.pl
```diff
@@ -0,0 +1,49 @@
...
+         push(@out, "");
+         push(@out, "<details><summary>Show log</summary><code><pre>");
+      }
+   } elsif ( $result->is_comment and $in_error == 1 ) {
+      # Print error contents
+      push ( @out, $result->raw );
+   }
+}
+
+# Print out the contents of @out, leaving off the last little formatting bits
```
erikjohnston: What formatting bits?

erikjohnston:
Also, why isn't this killing off the old script? Lets not keep both.

anoadragon453:
> Also, why isn't this killing off the old script? Lets not keep both.

The old script is in Synapse's repo. I need to make a PR for it and changing the CI.

erikjohnston:
At least you can change the `synapse_sytest` script to use this?

erikjohnston:
Also, please don't put this in the root of the repo

PR #717, opened 2019-10-05
ensure that server ACLs are enforced

richvdh:
Some tests that I found on a branch in my working copy. Require https://github.com/matrix-org/synapse/pull/6170.


richvdh:
flaky test is flaky

PR #718, opened 2019-10-05
Regression test for sync cache consistency bug

richvdh:
(though note also that this requires synapse to run in worker mode, with some
replication-torturing patches, to show up the issue.)

(Another test I found sitting in my working copy)


erikjohnston:
Is there a synapse issue?

richvdh:
there may have been once, but apparenly I wrote this 6 months ago

richvdh:
flaky test is unrelated

PR #719, opened 2019-10-08
Update a test to give better diagnostics

richvdh:


tests/30rooms/13guestaccess.pl
```diff
@@ -311,97 +312,94 @@
...
+            room_alias_name => $aliasname,
+         )->on_done( sub {
+            ( $room_id, undef ) = ( @_ );
+         });
 
-            matrix_set_room_guest_access( $user, $room_id, "can_join" );
-         }),
+         if( $settings->{guest_access} ) {
+            $f = $f->then( sub {
+               matrix_set_room_guest_access( $user, $room_id, $settings->{guest_access} );
```
erikjohnston: I'd have been tempted to use `initial_state` on `/createRoom` API

richvdh: fair. I was just sticking with the way the test did it originally, but with less C&P.

richvdh:
I'm merging this despite the failing test, because the test failure isn't new.

PR #722, opened 2019-10-10
Enable testing of log contexts for synapse

erikjohnston:
Flag was added in https://github.com/matrix-org/synapse/pull/6127


erikjohnston:
Looking across some builds it doesn't seem to have slowed things down, most of the differences look like normal variance.

PR #733, opened 2019-10-25
Fix conversion of numbers into strings

richvdh:
Wrap numbers in JSON bodies

This should help avoid the idiocy where logging the body of a federation 
request turns the numbers into strings (which then screws up their later
use in other JSON-encoded things).


richvdh:
Note that this replicates the behaviour of [`do_request_json` etc](https://github.com/matrix-org/sytest/blob/74e692e8f47c5c5e7a609dab01c254fe2034977a/lib/SyTest/HTTPClient.pm#L115), where it seems to work well.

PR #735, opened 2019-10-29
Tests for room dir search

erikjohnston:
Requires https://github.com/matrix-org/synapse/pull/6268

### DIFF
--- a/tests/30rooms/70publicroomslist.pl
+++ b/tests/30rooms/70publicroomslist.pl
@@ -248,3 +248,44 @@ test "Can paginate public room list",
          Future->done( 1 );
       })
    };
+
+test "Can search public room list",
+   requires => [ $main::HOMESERVER_INFO[0], local_user_fixture() ],
+
+   check => sub {
+      my ( $info, $local_user ) = @_;
+
+      my $room_id;
+
+      matrix_create_room( $local_user,
+         visibility      => "public",
+         name            => "Test Name",
+         topic           => "Test Topic Wombles",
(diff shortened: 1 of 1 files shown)


erikjohnston:
Good point, should be fixed now

tests/30rooms/70publicroomslist.pl
```diff
@@ -248,3 +248,46 @@
...
+            do_request_json_for( $local_user,
+               method => "POST",
+               uri    => "/r0/publicRooms",
+
+               content => {
+                  filter => {
+                     generic_search_term => "wombles",  # Search case insesitively
+                  }
+               },
+               )->then( sub {
```
richvdh:
```suggestion
            )->then( sub {
```

tests/30rooms/70publicroomslist.pl
```diff
@@ -248,3 +248,46 @@
...
+
+               assert_json_keys( $body, qw( chunk ) );
+
+               # We only expect to find a single result
+               assert_eq( scalar @{ $body->{chunk} }, 1 );
+
+               assert_eq( $body->{chunk}[0]{room_id}, $room_id );
+
+               Future->done( 1 );
+            })
```
richvdh: can you add an on_fail here (like https://github.com/matrix-org/sytest/blob/0b009ceaf2fec56edcc8a0015b548a306f79d990/tests/30rooms/70publicroomslist.pl#L104) so that when it fails we can see *why* it is failing?

PR #736, opened 2019-10-31
Add simple /purge_room admin API test

erikjohnston:
Depends on https://github.com/matrix-org/synapse/pull/6307

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -447,3 +447,30 @@ sub await_message_in_room
       log_if_fail "Found event $event_id for $user_id";
    })
 }
+
+
+test "/purge_room",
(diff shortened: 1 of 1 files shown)


erikjohnston:
That's probably fair, though we probably want to think about what to do about the other tests in the file

erikjohnston:
I guess one thing that this does test is if they work in worker mode?

richvdh:
still kinda think we want to test that in `synapse/tests` rather than here... even if that means doing a bunch of annoying boilerplatey work to get a pair of workers set up in a trial env.

richvdh:
> what to do about the other tests in the file

kill them

kill them all

MWAHAHAHAHAHA

PR #741, opened 2019-11-07
[DINSIC] Tests for user info endpoint

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse-dinsic/pull/12

PR #743, opened 2019-11-12
[DINSIC] Enable account validity on tests

anoadragon453:
Enables account validity tracking in Synapse, such that we can test it.

Requires https://github.com/matrix-org/synapse/pull/6359


anoadragon453:
This was actually done in https://github.com/matrix-org/sytest/pull/741

PR #744, opened 2019-11-18
Update the SyTest Docker files

hawkowl:
This has py35 + 3.7, as well as go 1.10 and 1.13.


docker/base-buster.Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM debian:buster
```
richvdh: can we use a build arg rather than cutting and pasting the entire dockerfile?

richvdh: compare: https://github.com/matrix-org/sytest/pull/745

hawkowl: yeah that works better, done

PR #746, opened 2019-11-19
Fix perl 5.28 incompatibility

richvdh:
The SvPOK hack emits an error on perl 5.28 (`Use of strings with code points over 0xFF as arguments to bitwise xor (^) operator is not allowed`). Let's try a call to `is_utf8`.


hawkowl:
Merged with https://github.com/matrix-org/sytest/commit/70e5ef215d939a5d2870fe586c17967be7c1f4b6

PR #747, opened 2019-11-19
Add test for removing redactions from search results

anoadragon453:
Adds a test that sends two messages, redacts one of them, then searches for both. It should only get one, the non-redacted event, back.

Synapse PRs: https://github.com/matrix-org/synapse/pull/6377, https://github.com/matrix-org/synapse/pull/6522


anoadragon453:
I've fixed the test to use `scalar @$results` instead of `length @$results`, as `length` apparently always returns `1`, which was the value we were testing for.

PR #748, opened 2019-11-25
Don't modify source volumes when running in docker

richvdh:
I've fallen into this trap several times now: I don't expect the docker image
to modify my local files.


richvdh:
I think @turt2live introduced it; I asked him about it once but the answer was a bit handwavey. I don't think it's the right solution to whatever the problem was, anyway; if it's necessary it should be done in the docker image construction.

PR #749, opened 2019-11-25
Fix a flaky cross-signing test

richvdh:
The basic problem in this test is that there are three things that can cause a
"devices/changed" sync update:
 * inviting the user
 * the user joining
 * the user doing his cross-signing

... and it's hard to be sure that we're getting the right update. If we make
sure that we wait for each update, it fixes the race condition.


richvdh:
@uhoreg perhaps you'd like to review this one :)

PR #750, opened 2019-11-26
Don't add outliers to forward extremities.

erikjohnston:
The inbuilt federation library blindly added all events as foreward
extremities. Normally this would be fine as adding all state events in
the correct order removed all superfluous extremities, however if not
done in the right order, or there was a gap, then new events would get
sent out with a random assortment of prev events. This led to some
events being rejected due to the resolved state being not what was
expected.

### DIFF
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -280,7 +280,7 @@ sub create_and_insert_event
 
    $room->insert_event( $event );
 
-Inserts a new event into the database, updating the rooms view of the forward
+Inserts a new event into the database, updating the room's view of the forward
 extremities (i.e. event IDs to use as the prev events of the the next
 generated event).
 
@@ -308,7 +308,7 @@ sub insert_event
 
    $room->insert_outlier_event( $event );
 
-Inserts a new event into the database, without updating the rooms forward
+Inserts a new event into the database, *without* updating the room's forward
 extremities (i.e. event IDs to use as the prev events of the the next
 generated event).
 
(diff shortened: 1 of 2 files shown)


lib/SyTest/Federation/Room.pm
```diff
@@ -268,20 +268,28 @@ sub create_and_insert_event
...
 
    return $event unless wantarray;
    return ( $event, $event_id );
 }
 
+=head2 insert_event
+
+   $event = $room->insert_event( $event );
+
+Inserts a new event into the database, updating prev_events.
```
richvdh: perhaps you could explain what prev_events here is, and why I should care. I think I'd read this description and expect it to update `prev_events` on the event itself, but that doesn't appear to be the case.

similarly below.

lib/SyTest/Federation/Room.pm
```diff
@@ -268,20 +268,30 @@ sub create_and_insert_event
...
 
    return $event unless wantarray;
    return ( $event, $event_id );
 }
 
+=head2 insert_event
+
+   $room->insert_event( $event );
+
+Inserts a new event into the database, updating the rooms view of the forward
```
richvdh:
```suggestion
Inserts a new event into the database, updating the room's view of the forward
```

lib/SyTest/Federation/Room.pm
```diff
@@ -293,7 +303,18 @@ sub insert_event
    extract_by { $to_remove{ $self->id_for_event($_) } } @$prev_events;
 }
 
-sub _insert_event
+
+=head2 insert_outlier_event
+
+   $room->insert_outlier_event( $event );
+
+Inserts a new event into the database, without updating the rooms forward
```
richvdh:
```suggestion
Inserts a new event into the database, *without* updating the room's forward
```

PR #752, opened 2019-11-26
Move scripts which aren't in docker out of `docker`

richvdh:
Since these scripts were in the docker directory, I thought they would be in the docker image. Turns out, not so much.

`bootstrap.sh`, which *is* in the docker image, is updated to cope with the scripts being either in the old place or the new, so that we can bump the docker image without breaking existing branches.

Also a couple of bits of cleanup.


richvdh:
(of course CI on this branch is going to fail until we update the docker image)

richvdh:
now passing modulo flaky worker tests, for which see https://github.com/matrix-org/synapse/pull/6429

PR #753, opened 2019-11-26
Enable replication torturing for synapse worker mode

richvdh:
this helps find race conditions in the worker-mode tests.


richvdh:
turns out lots of our tests are flakey :/

erikjohnston:
Can't we just add this to the invocation in CI?

richvdh:
I guess so. Would that be preferable?

erikjohnston:
I think so, if for no other reason than it'd be nice to still be able to run the worker tests without torture replication on (as it takes so long)

richvdh:
https://github.com/matrix-org/pipelines/pull/18 then

PR #754, opened 2019-11-27
support comments in blacklist/whitelist files

richvdh:


run-tests.pl
```diff
@@ -222,9 +222,14 @@ sub usage
 if ( $BLACKLIST_FILE ) {
    open( my $blacklist_data, "<:encoding(UTF-8)", $BLACKLIST_FILE ) or die "Couldn't open blacklist file for reading: $!\n";
    while ( my $test_name = <$blacklist_data> ) {
-      # Trim whitespace
+      # Trim whitespace and comments
       chomp $test_name;
-      $TEST_BLACKLIST{$test_name} = 1;
+      $test_name =~ s/#.*//;
+      $test_name =~ s/^\s*//;
+      $test_name =~ s/\s*$//;
```
erikjohnston: Any reason not to chomp?

richvdh: we do chomp, at line 226.

Note that chomp just removes trailing newlines; it does not remove other trailing whitespace.

erikjohnston: Oh right

PR #756, opened 2019-11-28
Add tests for ephemeral messages

babolivier:
Add tests for https://github.com/matrix-org/synapse/pull/6409

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -255,6 +255,8 @@ sub start
 
         require_membership_for_aliases => "false",
 
+        enable_ephemeral_messages => "true",
+
         $self->{recaptcha_config} ? (
            recaptcha_siteverify_api => $self->{recaptcha_config}->{siteverify_api},
            recaptcha_public_key     => $self->{recaptcha_config}->{public_key},
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -255,6 +255,7 @@ sub start
 
         require_membership_for_aliases => "false",
 
+        # Enable ephemeral message support (MSC2228)
         enable_ephemeral_messages => "true",
 
         $self->{recaptcha_config} ? (
--- a/tests/30rooms/04messages.pl
+++ b/tests/30rooms/04messages.pl
@@ -369,7 +369,7 @@ test "Ephemeral messages received from clients are correctly expired",
           matrix_get_room_messages( $user, $room_id, filter => $filter )
       })->then( sub {
          my ( $body ) = @_;
-         log_if_fail "Response body", $body;
+         log_if_fail "Response body before expiry", $body;
 
          my $chunk = $body->{chunk};
          @$chunk == 1 or
@@ -384,7 +384,7 @@ test "Ephemeral messages received from clients are correctly expired",
          matrix_get_room_messages($user, $room_id, filter => $filter )
       })->then( sub {
          my ( $body ) = @_;
-         log_if_fail "Response body", $body;
+         log_if_fail "Response body after expiry", $body;
 
          my $chunk = $body->{chunk};
          @$chunk == 1 or
--- a/tests/30rooms/04messages.pl
+++ b/tests/30rooms/04messages.pl
@@ -364,34 +364,44 @@ test "Ephemeral messages received from clients are correctly expired",
          content => {
             msgtype                          => "m.text",
             body                             => "This is a message",
-            "org.matrix.self_destruct_after" => $now_ms + 250,
+            "org.matrix.self_destruct_after" => $now_ms + 1000,
          },
       )->then( sub {
-         matrix_get_room_messages( $user, $room_id, filter => $filter )
+         my $iter = 0;
+         retry_until_success {
+            matrix_get_room_messages( $user, $room_id, filter => $filter )->then( sub {
+               $iter++;
+               my ( $body ) = @_;
+               log_if_fail "Iteration $iter: response body before expiry", $body;
+
+               my $chunk = $body->{chunk};
+
+               @$chunk == 1 or
+                  die "Expected 1 message";
+
+               # Make sure we can read the message's content before it expires.
+               assert_eq( $chunk->[0]{content}{body}, "This is a message",
+                  'chunk[0] content body' );
+
+               Future->done( 1 )
+            })->on_fail( sub {
+               my ( $exc ) = @_;
+               chomp $exc;
+               log_if_fail "Iteration $iter: not ready yet: $exc";
+            });
+         }
       })->then( sub {
-         my ( $body ) = @_;
-         log_if_fail "Response body before expiry", $body;
-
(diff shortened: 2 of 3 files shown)


tests/30rooms/04messages.pl
```diff
@@ -349,3 +349,55 @@
...
+      my $filter = '{"types":["m.room.message"]}';
+
+      matrix_send_room_message( $user, $room_id,
+         content => {
+            msgtype                          => "m.text",
+            body                             => "This is a message",
+            "org.matrix.self_destruct_after" => $now_ms + 1000,
+         },
+      )->then( sub {
+          matrix_get_room_messages( $user, $room_id, filter => $filter )
```
richvdh: I'm no fan of the 3-space indentation convention, but let's try to stick to it.

There are more instances below.

tests/30rooms/04messages.pl
```diff
@@ -349,3 +349,55 @@
...
+         @$chunk == 1 or
+            die "Expected 1 message";
+
+         # Make sure we can read the message's content before it expires.
+         assert_eq( $chunk->[0]{content}{body}, "This is a message",
+            'chunk[0] content body' );
+
+         sleep( 2 );
+
+         retry_until_success {
```
richvdh: this really isn't great, because it means that the test will never fail properly, and worse, it won't tell you why it's failing, because the assertion error gets swallowed.

one way to improve it is to make sure that the error gets logged, with an `on_fail` handler:

```perl
my $iter = 0;
retry_until_success {
   $iter++;
   matrix_get_room_messages( ... ) ->then( sub {
      # ...
      log_if_fail "Iteration $iter: response body after expiry", $body;
      # ...
      Future->done(1);
   })->on_fail( sub {
      my ( $exc ) = @_;
      chomp $exc;
      log_if_fail "Iteration $iter: not ready yet: $exc";
   });
};

File: tests/50federation/31room-send.pl [hunk 3]
```diff
@@ -145,3 +146,71 @@
...
    };
+
+test "Ephemeral messages received from servers are correctly expired",
+   requires => [ local_user_and_room_fixtures(), federation_user_id_fixture(),
+                 $main::OUTBOUND_CLIENT ],
+
+   do => sub {
+      my ( $local_user, $room_id, $federated_user, $outbound_client ) = @_;
+
+      my $now_ms = int( time() * 1000 );
```
richvdh: this should probably be after the join.

tests/30rooms/04messages.pl
```diff
@@ -349,3 +349,55 @@
...
+      my ( $user, $room_id ) = @_;
+
+      my $now_ms = int( time() * 1000 );
+      my $filter = '{"types":["m.room.message"]}';
+
+      matrix_send_room_message( $user, $room_id,
+         content => {
+            msgtype                          => "m.text",
+            body                             => "This is a message",
+            "org.matrix.self_destruct_after" => $now_ms + 1000,
```
richvdh: I wonder if we can reduce this and the sleep time, to get the test to run quicker.

tests/50federation/31room-send.pl
```diff
@@ -145,3 +146,71 @@
...
+      my $filter = '{"types":["m.room.message"]}';
+
+      $outbound_client->join_room(
```
richvdh: `federated_rooms_fixture` is a useful way to cut out some of this boilerplate fyi. no real need to change this unless you want to.

tests/50federation/31room-send.pl
```diff
@@ -145,3 +146,71 @@
          });
       });
    };
+
+test "Ephemeral messages received from servers are correctly expired",
+   requires => [ local_user_and_room_fixtures(), federation_user_id_fixture(),
+                 $main::OUTBOUND_CLIENT ],
```
richvdh: Note that sytest spins up *two* synapse instances, and where we're not doing anything magical with the protocol that makes it hard to do so, generally I prefer to test federation stuff by sending messages between those two servers rather than using the sytest federation impl: it gives a better end-to-end test. You can create users on the second server with `remote_user_fixture`: see other tests for further details.

Again, a thought for the future rather than one that needs changing here.

tests/50federation/31room-send.pl
```diff
@@ -145,3 +146,71 @@
...
+
+         my $chunk = $body->{chunk};
+         @$chunk == 1 or
+            die "Expected 1 message";
+
+         # Make sure we can read the message's content before it expires.
+         assert_eq( $chunk->[0]{content}{body}, "This is a message",
+            'chunk[0] content body' );
+
+         sleep( 2 );
```
richvdh: again, sleep ftl

tests/50federation/31room-send.pl
```diff
@@ -145,3 +146,71 @@
...
+
+         retry_until_success {
```
richvdh: again, needs `on_fail` or similar

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -255,6 +255,8 @@ sub start
 
         require_membership_for_aliases => "false",
 
+        enable_ephemeral_messages => "true",
```
richvdh: this could do with a comment. what does enabling it do?

tests/30rooms/04messages.pl
babolivier: Ugh Pycharm why

tests/30rooms/04messages.pl
babolivier: I'm reducing this to 250ms and the delay to 0.5s.

babolivier: Actually, nevermind, because `--torture-replication` means that if that is too low, when we want to check the message before it expires, the message has already expired by the time it makes it to the `client_reader`...

I've set it to `1000 / 1.5` in the local test and `2000 / 2.5` in the federated tests because those are the values for which I managed to make it work somewhat reliably on my local machine.

tests/30rooms/04messages.pl
```diff
@@ -349,3 +349,72 @@
...
+         },
+      )->then( sub {
+         my $iter = 0;
+         retry_until_success {
```
richvdh: we ought to be able to use sync for this instead.

tests/50federation/31room-send.pl
```diff
@@ -145,3 +146,58 @@
...
+      matrix_invite_user_to_room( $local_user, $federated_user, $room_id )->then( sub {
+         matrix_join_room( $federated_user, $room_id )
+      })->then( sub {
+         my $now_ms = int( time() * 1000 );
+
+         matrix_send_room_message( $local_user, $room_id,
+            content => {
+               msgtype                          => "m.text",
+               body                             => "This is a message",
+               "org.matrix.self_destruct_after" => $now_ms + 500,
```
richvdh: I'm surprised that the expiry time for the federated test is lower than that for the local test, and it seems to be different to what you wrote previously.

babolivier: Yeah, I've tried tweaking this one a bit to see if I could lower it. I'll probably try a few other values later (and harmonise them between both tests).

PR #757, opened 2019-12-02
Add sytest blacklist for DINUM synapse

anoadragon453:
Have DINUM sytest make use of a sytest blacklist file for synapse builds.

synapse-dinsic PR that adds the blacklist file: https://github.com/matrix-org/synapse-dinsic/pull/17


anoadragon453:
Tests are expected to fail due to a deprecated sytest option that synapse-dinsic still uses.

PR #760, opened 2019-12-05
Test v2 endpoint for /send_join

babolivier:
One half of testing the implementation of MSC1802 (https://github.com/matrix-org/synapse/issues/6293). Also adds support for `await_request_[...]` on non-v1 endpoints.

### DIFF
--- a/lib/SyTest/Federation/Server.pm
+++ b/lib/SyTest/Federation/Server.pm
@@ -402,22 +402,20 @@ sub mk_await_request_pair
       }
    };
 
-   # Don't specify the version prefix in the function name if the version is v1 so we
-   # don't break existing function calls.
-   my ( $await_request_function_name, $on_request_function_name );
+   # Don't specify the version prefix in the await function name if the version is v1 so
+   # we don't break existing function calls.
+   my ( $await_request_function_name );
    if( $versionprefix eq "v1" ) {
       $await_request_function_name = "await_request_$shortname";
-      $on_request_function_name = "on_request_federation_$shortname"
    }
    else {
       $await_request_function_name = "await_request_$versionprefix\_$shortname";
-      $on_request_function_name = "on_request_federation_$versionprefix\_$shortname";
    }
 
    no strict 'refs';
    no warnings 'redefine';
    *{"${class}::$await_request_function_name"} = $awaitfunc;
-   *{"${class}::$on_request_function_name"} = $on_requestfunc;
+   *{"${class}::on_request_federation_$versionprefix\_$shortname"} = $on_requestfunc;
 }
 
 __PACKAGE__->mk_await_request_pair(
--- a/lib/SyTest/Federation/Server.pm
+++ b/lib/SyTest/Federation/Server.pm
@@ -402,10 +402,22 @@ sub mk_await_request_pair
       }
    };
 
+   # Don't specify the version prefix in the function name if the version is v1 so we
+   # don't break existing function calls.
+   my ( $await_request_function_name, $on_request_function_name );
+   if( $versionprefix eq "v1" ) {
+      $await_request_function_name = "await_request_$shortname";
+      $on_request_function_name = "on_request_federation_$shortname"
+   }
+   else {
+      $await_request_function_name = "await_request_$versionprefix\_$shortname";
+      $on_request_function_name = "on_request_federation_$versionprefix\_$shortname";
+   }
+
    no strict 'refs';
    no warnings 'redefine';
-   *{"${class}::await_request_$shortname"} = $awaitfunc;
-   *{"${class}::on_request_federation_v1_$shortname"} = $on_requestfunc;
+   *{"${class}::$await_request_function_name"} = $awaitfunc;
+   *{"${class}::$on_request_function_name"} = $on_requestfunc;
 }
 
 __PACKAGE__->mk_await_request_pair(
--- a/lib/SyTest/Federation/Server.pm
+++ b/lib/SyTest/Federation/Server.pm
@@ -315,10 +315,10 @@ sub on_request_federation_v1_send_join
 sub mk_await_request_pair
 {
    my $class = shift;
-   my ( $shortname, $paramnames ) = @_;
+   my ( $shortname, $versionprefix, $paramnames ) = @_;
    my @paramnames = @$paramnames;
 
-   my $okey = "awaiting_$shortname";
+   my $okey = "awaiting_$versionprefix\_$shortname";
 
    my $awaitfunc = sub {
       my $self = shift;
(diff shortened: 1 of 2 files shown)


lib/SyTest/Federation/Server.pm
```diff
@@ -312,13 +312,40 @@ sub on_request_federation_v1_send_join
...
+      map { $_->[0] } @{ $event->{auth_events} }
+   );
+   my @state_events = $room->current_state_events;
+
+   $room->insert_event( $event );
+
+   Future->done( json => {
+      auth_chain => \@auth_chain,
+      state      => \@state_events,
+   } );
```
erikjohnston: Can we make `on_request_federation_v1_send_join` just wrap this rather than duplicating the code?

tests/50federation/30room-join.pl
```diff
@@ -327,6 +326,227 @@ sub assert_is_valid_pdu {
    };
 
 
+
+test "Inbound federation supports v2 /send_join",
```
erikjohnston: Can we not reuse the `Inbound federation can receive v1 room-join requests` test here? I.e. something like:

```perl
foreach my $version ( qw( v1 v2 ) ) {
   test "Inbound federation supports $version /send_join",
      ...
}
```

or something?

tests/50federation/30room-join.pl
```diff
@@ -327,6 +326,227 @@ sub assert_is_valid_pdu {
...
+         log_if_fail "State", \%state;
+
+         # TODO: lots more checking. Requires spec though
+         Future->done(1);
+      });
+   };
+
+
+
+test "Outbound federation falls back to v1 /send_join if v2 isn't available",
```
erikjohnston: This is presumably the same as the outbound v2 /send_join test, but only awaiting on v1? Can we dedupe as per above?

richvdh:
I have some thoughts on this too... bear with me.

lib/SyTest/Federation/Server.pm
```diff
@@ -373,66 +402,80 @@ sub mk_await_request_pair
       }
    };
 
+   # Don't specify the version prefix in the await function name if the version is v1 so
+   # we don't break existing function calls.
+   my ( $await_request_function_name );
+   if( $versionprefix eq "v1" ) {
+      $await_request_function_name = "await_request_$shortname";
```
richvdh: I think we should probably aim to switch to consistently using version-prefixed method names.

No problem keeping this around for now, but in addition to the prefixed name. So you end up with something like:

```perl
*{"${class}::await_request_${versionprefix}_${shortname}"} = $awaitfunc;
# deprecated alternative name for v1 endpoints
*{"${class}::await_request_${shortname}"} = $awaitfunc if ${versionprefix} eq "v1";
```

tests/50federation/30room-join.pl
```diff
@@ -52,103 +52,118 @@ sub assert_is_valid_pdu {
 }
 push our @EXPORT, qw( assert_is_valid_pdu );
 
-test "Outbound federation can send room-join requests",
-   requires => [ local_user_fixture(), $main::INBOUND_SERVER,
-                 federation_user_id_fixture() ],
+foreach my $versionprefix ( qw( v1 v2 ) ) {
+   test "Inbound federation supports $versionprefix /send_join",
```
richvdh: wait. what is this testing?

tests/50federation/30room-join.pl
```diff
@@ -52,103 +52,118 @@ sub assert_is_valid_pdu {
...
 
-      Future->needs_all(
-         # Await PDU?
+         my $await_request_send_join;
 
-         $inbound_server->await_request_make_join( $room_id, $user->user_id )->then( sub {
-            my ( $req, $room_id, $user_id ) = @_;
+         if( $versionprefix eq "v1" ) {
+            $await_request_send_join = $inbound_server->await_request_v1_send_join( $room_id );
+         }
```
richvdh: surely in the v1 case we should expect that the v2 request happens first; we can then 404 it and wait for the v1 request

tests/50federation/30room-join.pl
babolivier: That a client requesting to join a remote room results in a v1 `make_join` request and a v[12] `send_join` request, and that the server can process the responses. I just realised I mixed up "inbound" and "outbound" in the test names, I've fixed that in https://github.com/matrix-org/sytest/pull/760/commits/78f8a530f6e09a78eda19e12f8a5c3e2c917cb9e

tests/50federation/30room-join.pl
babolivier: Why? SyTest already 404s requests on the endpoints it's not expecting requests on.

richvdh: doesn't it spit out some warnings about it?

babolivier: Nope:

```
$ ./run-tests.pl tests/50federation/30room-join.pl
Running tests/50federation/30room-join.pl...
  Testing if: Outbound federation can query v1 /send_join... 
Tests: 0 / 17 | OKSignature ok
subject=CN = localhost
Getting CA Private Key
Signature ok
subject=CN = localhost
 # Started test HTTPS Server at localhost:42521
 # Started test SMTP Server at localhost:34521
 # Starting Homeserver using SyTest::HomeserverFactory::Synapse=HASH(0x556a57bad378)
 # Starting server-0
 # Clearing SQLite database at :memory:
  Testing if: Outbound federation can query v1 /send_join... 
Tests: 0 / 17 | OKSignature ok
subject=CN = localhost
 # Generating config for port 8800
 # Creating config for server 0 with command python -m synapse.app.homeserver --config-path /home/brendan/Documents/matrix/sytest-stable/server-0/config.yaml --server-name localhost:8800 --generate-config --report-stats=no
  Testing if: Outbound federation can query v1 /send_join... 
 # Starting server 0 for port 8800 with command python -m synapse.app.homeserver --config-path /home/brendan/Documents/matrix/sytest-stable/server-0/config.yaml --server-name localhost:8800
 # Connecting to server 8800
 # Connected to server 8800
 # Started server-0
  Testing if: Outbound federation can query v1 /send_join... PASS
  Testing if: Outbound federation can query v2 /send_join... PASS
  Testing if: Outbound federation passes make_join failures through to the client... PASS
  Testing if: Inbound federation can receive v1 /send_join... PASS
  Testing if: Inbound federation can receive v2 /send_join... PASS
...
```

richvdh: fair enough. can you chuck in a comment that you're relying on that behaviour?

babolivier: Done in https://github.com/matrix-org/sytest/pull/760/commits/5db16de9fda73c1558565cc2f5181cfde10cd7a9.

PR #761, opened 2019-12-06
Add a passing test for recovery from gappy device list updates

ara4n:
as part of narrowing down https://github.com/matrix-org/synapse/issues/6399 and generally putting more tests around DL sync logic.


erikjohnston:
Excitingly it looks like the test fails on workers

PR #762, opened 2019-12-09
Test that changing an avatar to a non-image type media is disallowed

anoadragon453:
Associated synapse-dinsic PR: https://github.com/matrix-org/synapse-dinsic/pull/19

Adds a test to check that a plaintext file can't be used as an avatar. Also changes the "$datum updates affect room member events" test to upload an actual image as an avatar, instead of a plaintext file :)


anoadragon453:
This ended up being far too much work, and something that should be tested in a Synapse unittest, rather than Sytest.

PR #764, opened 2019-12-11
Deflake "Inviting an AS-hosted user asks the AS server"

richvdh:
We need to make sure the AS is ready for the event before synapse sends it,
otherwise we might miss it.

### DIFF
--- a/tests/60app-services/03passive.pl
+++ b/tests/60app-services/03passive.pl
@@ -2,7 +2,7 @@ my $user_fixture = local_user_fixture();
 
 my $room_fixture = room_fixture( $user_fixture );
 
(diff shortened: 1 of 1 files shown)


tests/60app-services/03passive.pl
```diff
@@ -22,24 +22,25 @@
             });
          });
 
-      matrix_invite_user_to_room( $creator, $user_id, $room_id )
-         ->SyTest::pass_on_done( "Sent invite" )
```
richvdh: this is pretty redundant, since `matrix_invite_user_to_room` does a log_on_fail when the invite is sent. And since this is the only `pass_on_done`, we may as well convert from a `multi_test` to a regular `test`.

PR #765, opened 2019-12-11
Sanity-checking of some params

richvdh:
Some sanity-checking I added while I was trying to figure out where I was
getting an undefined event from.

### DIFF
--- a/lib/SyTest/Federation/Protocol.pm
+++ b/lib/SyTest/Federation/Protocol.pm
@@ -26,6 +26,7 @@ use Protocol::Matrix qw(
    encode_base64_unpadded
 );
 
+use Carp;
 
 use Exporter 'import';
 our @EXPORT_OK = qw(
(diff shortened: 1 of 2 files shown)


lib/SyTest/Federation/Protocol.pm
```diff
@@ -26,6 +26,7 @@ use Protocol::Matrix qw(
    encode_base64_unpadded
 );
 
+use Carp;
```
erikjohnston: What's Carp?

lib/SyTest/Federation/Room.pm
```diff
@@ -385,14 +385,21 @@ sub make_join_protoevent
       $self->get_current_state_event( "m.room.create" ),
       $self->get_current_state_event( "m.room.join_rules" ),
    );
+   my $auth_events = $self->make_event_refs( @auth_events );
+
```
erikjohnston: Do we want to do the same checks for `auth_events`?

lib/SyTest/Federation/Protocol.pm
richvdh: https://perldoc.perl.org/Carp.html, which exports `croak`.

lib/SyTest/Federation/Room.pm
richvdh: I guess.

richvdh: actually no, that would be redundant. there's already a `grep defined` at line 384.

PR #771, opened 2019-12-12
Make the replication torture level adjustable

richvdh:


richvdh:
https://perldoc.perl.org/Getopt/Long.html#Summary-of-Option-Specifications may be helpful to the reviewer...

lib/SyTest/HomeserverFactory/Synapse.pm
```diff
@@ -109,7 +109,7 @@ sub get_options
 
    return (
       'dendron-binary=s' => \$self->{args}{dendron_binary},
-      'torture-replication+' => \$self->{args}{torture_replication},
+      'torture-replication:50' => \$self->{args}{torture_replication},
```
erikjohnston: The documentation you linked to seemed to suggest that the colon should be followed by a type, rather than a value?

richvdh: image

erikjohnston: oooooooooooooooooooooooooooooooo

erikjohnston: that's special

PR #772, opened 2019-12-16
workaround for https://github.com/matrix-org/synapse/issues/6536

richvdh:
(https://github.com/matrix-org/synapse/issues/6536)

I feel like there are going to be a lot of these

### DIFF
--- a/tests/50federation/38receipts.pl
+++ b/tests/50federation/38receipts.pl
@@ -8,15 +8,20 @@ test "Outbound federation sends receipts",
    do => sub {
       my ( $creator_user, $room_id, $federated_user_id, $outbound_client, $inbound_server ) = @_;
 
-      my $event_id;
+      my ( $event_id, $room );
 
       $outbound_client->join_room(
          server_name => $creator_user->server_name,
          room_id     => $room_id,
          user_id     => $federated_user_id,
       )->then( sub {
+         ( $room ) = @_;
+         log_if_fail "Joined room $room_id";
+
+         # workaround for https://github.com/matrix-org/synapse/issues/6536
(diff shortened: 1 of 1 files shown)


tests/50federation/38receipts.pl
```diff
@@ -8,15 +8,20 @@
...
       $outbound_client->join_room(
          server_name => $creator_user->server_name,
          room_id     => $room_id,
          user_id     => $federated_user_id,
       )->then( sub {
+         ( $room ) = @_;
+         log_if_fail "Joined room $room_id";
+
+         # workaround for https://github.com/matrix-org/synapse/issues/6536
+         delay( 0.5 );
```
erikjohnston: I'd really rather avoid adding delays if possible given how long it takes to run sytest anyway. Can we not wait for the join to come down a sync? I know that's not quite the same, but its a heuristic we use elsewhere for similar things

richvdh: well... we're waiting for `latest_events_in_room` to get flushed, not the events stream: and part of the problem is that `latest_events_in_room` is flushed *after* the events stream.

so yeah we *could* wait for a sync and it would *probably* work, but really we'd just be using it as a proxy for "wait a bit", which we may as well do with a delay.

Compromise: we could make it only delay in worker mode?

erikjohnston: Yeah, doing that would help. It's fine for now I guess, I just don't want to end up in a situation where we delay in loads of tests, which I could see happening more and more as we continue to split things apart

richvdh: ... the right solution might be to fix the bug?

PR #775, opened 2019-12-20
Add support for synapse multi DB setups.

erikjohnston:
This allows specifying a `databases.yaml` file that contains multiple
databases.

I don't really know if this is the right way of doing it.

Uses https://github.com/matrix-org/synapse/pull/6580

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -376,6 +376,10 @@ sub _get_dbconfigs
 
       $db_configs{"main"} = \%db_config;
    }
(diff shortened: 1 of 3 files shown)


erikjohnston:
> I assume we can't yet delete _get_db_config? We should have a plan for it not to hang around as a massive c&p forever though. What's the plan for it? could you put a comment or something in?

Its used by dendrite config, which then pulls out the args and constructs a postgres URI string. My thought was that we probably don't want to try and use the same config format for dendrite and synapse TBH.

PR #777, opened 2020-01-05
Add test for ensuring power levels are correctly handled during upgrade

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/6633

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -14,6 +14,7 @@
 
 use Future::Utils qw( repeat );
 use List::Util qw( all first none );
+use URI::Escape qw( uri_escape );
 
 push our @EXPORT, qw ( upgrade_room_synced $TEST_NEW_VERSION );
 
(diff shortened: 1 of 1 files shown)


tests/30rooms/60version_upgrade.pl
```diff
@@ -373,6 +373,81 @@ sub upgrade_room_synced {
...
+         assert_deeply_eq(
+            $new_room_pl_event->{content},
+            $pl_content,
+            "power levels in replacement room",
+         );
+
+         # Check that the power levels in the old room have not changed
+         # by making sure there is not a power levels event for the old room
+         # in the sync response
+         my $old_room = $sync_body->{rooms}{join}{$room_id};
```
richvdh: have you checked this actually failed under the old code?

My suspicion is that it will not, since the bug is that we are modifying the existing PL event, rather than creating a new one. Maybe try fetching the state with `/r0/room/{roomId}/state/m.room.power_levels` or sth?

anoadragon453: It does fail with:

```
 | Got JSON::number(100), expected JSON::number(50) at {users}{@anon-20200114_100001-7:localhost:8800} for power levels in replacement room at tests/30rooms/60version_upgrade.pl line 435.
```

But perhaps fetching state is a more reliable method anyways?

anoadragon453: Switched. Test still fails on old code, passes on new.

tests/30rooms/60version_upgrade.pl
```diff
@@ -337,7 +338,7 @@ sub upgrade_room_synced {
       matrix_change_room_power_levels(
          $creator, $room_id, sub {
             ( $pl_content ) = @_;
-            $pl_content->{users}->{'@test:xyz'} = 40;
+            $pl_content->{users}->{'@test:xyz'} = JSON::number(40);
```
richvdh: oh heh, that's what I fixed with #789. Arguably yours is a better fix.

PR #778, opened 2020-01-06
enable frozen dicts in synapse

richvdh:


richvdh:
cf https://github.com/matrix-org/synapse/pull/6642

PR #783, opened 2020-01-21
Change postgres in docker image to use C collation/ctype.

erikjohnston:
This is what Synapse recommends, so we should test it.

c.f. https://github.com/matrix-org/synapse/pull/6734


erikjohnston:
Ack. Any reason this can't be done as an automated build on docker hub?

richvdh:
none that I'm aware of.

richvdh:
(it may already be automated somehow, but I don't think so)

erikjohnston:
Looking at it I think its a bit of a PITA as we can't specify different `DEBIAN_VERSION` build arg for different builds.

PR #787, opened 2020-01-22
 Change client_secret to be within [0-9a-zA-Z.=_-] 

anoadragon453:
Modify `client_secret` so that it can only be digits, to fit in the range of `[0-9a-zA-Z.=_-]`.

Synapse PR: https://github.com/matrix-org/synapse/pull/6767

### DIFF
--- a/tests/54identity.pl
+++ b/tests/54identity.pl
@@ -1,7 +1,7 @@
 sub bind_email_for_user {
    my ( $user, $address, $id_server, %params ) = @_;
 
-   my $client_secret = join "", map { chr 40 + rand 86 } 1 .. 20;
+   my $client_secret = join "", map { chr 48 + rand 10 } 1 .. 20;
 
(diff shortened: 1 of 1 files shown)


tests/54identity.pl
```diff
@@ -1,7 +1,7 @@
 sub bind_email_for_user {
    my ( $user, $address, $id_server, %params ) = @_;
 
-   my $client_secret = join "", map { chr 40 + rand 86 } 1 .. 20;
+   my $client_secret = join "", map { chr 48 + rand 10 } 1 .. 20 + "Az.=_-";
```
erikjohnston:
```suggestion
   // Generate a client secret that is just numerics with `"Az.=_-"` on the end.
   my $client_secret = (join "", map { chr 48 + rand 10 } 1 .. 20) . "Az.=_-";
```

PR #789, opened 2020-01-28
Fix power levels being sent as strings

richvdh:
Yes, `log_if_fail` modifies its inputs. FUN TIMES.


richvdh:
flaky test is flaly

PR #792, opened 2020-01-28
Add some tests to check for adding/removing aliases

richvdh:
Make sure that users can add and remove aliases even if the m.room.aliases PL
is restricted.

### DIFF
--- a/tests/30rooms/05aliases.pl
+++ b/tests/30rooms/05aliases.pl
@@ -7,6 +7,7 @@ my $alias_localpart = "#☕";
 my $room_alias;
 
 my $creator_fixture = local_user_fixture();
+my $second_user_fixture = local_user_fixture();
 
 my $room_fixture = room_fixture( $creator_fixture );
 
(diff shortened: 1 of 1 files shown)


tests/30rooms/05aliases.pl
```diff
@@ -101,31 +102,119 @@
...
+      my ( $creator, $other_user, $room_id, $room_alias ) = @_;
+      _test_can_create_and_delete_alias( $room_id, $other_user, $room_alias );
+   };
 
-      do_request_json_for( $user,
-         method => "PUT",
-         uri    => "/r0/directory/room/$room_alias",
 
-         content => { room_id => $room_id },
+test "Regular users can add and delete aliases when m.room.aliases is restricted",
```
erikjohnston: Will this only pass on old room versions?

richvdh: it should pass on all room versions. The idea of MSC2260 is that you can still add aliases, but if you don't have perms in the room there will be no m.room.aliases event.

PR #793, opened 2020-01-29
Test that remote device list is refetched on leave/rejoin

erikjohnston:
Test for matrix-org/synapse#6797

### DIFF
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -248,10 +248,13 @@ sub join_room
       $self->do_request_json(
          method   => "PUT",
          hostname => $server_name,
-         uri      => "/v2/send_join/$room_id/$event_id",
+         uri      => "/v1/send_join/$room_id/$event_id",
          content  => $member_event,
       )->then( sub {
          my ( $join_body ) = @_;
+
+         # /v1/send_join has an extraneous [ 200, ... ] wrapper (see MSC1802)
+         $join_body = $join_body->[1];
 
          my $room = SyTest::Federation::Room->new(
             datastore => $store,
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -248,13 +248,10 @@ sub join_room
       $self->do_request_json(
          method   => "PUT",
          hostname => $server_name,
-         uri      => "/v1/send_join/$room_id/$event_id",
+         uri      => "/v2/send_join/$room_id/$event_id",
          content  => $member_event,
       )->then( sub {
          my ( $join_body ) = @_;
-
-         # /v1/send_join has an extraneous [ 200, ... ] wrapper (see MSC1802)
-         $join_body = $join_body->[1];
 
          my $room = SyTest::Federation::Room->new(
             datastore => $store,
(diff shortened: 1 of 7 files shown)


lib/SyTest/Federation/Client.pm
```diff
@@ -248,14 +248,11 @@ sub join_room
       $self->do_request_json(
          method   => "PUT",
          hostname => $server_name,
-         uri      => "/v1/send_join/$room_id/$event_id",
+         uri      => "/v2/send_join/$room_id/$event_id",
```
richvdh: I've got a feeling we were deliberately leaving this for now, since (a) /v2/send_join isn't specced yet; (b) compatibility with dendrite.

tests/50federation/30room-join.pl
```diff
@@ -86,7 +80,18 @@ sub assert_is_valid_pdu {
             # that 404 response from SyTest and that fallback mechanism to test that the
```
richvdh: this comment seems to be somewhat incorrect.

Also: this change precludes testing against servers which don't support the v2 endpoint. Is it necessary? If so, Can you make the /v2/ request optional? (Just make a future, and cancel it once the v1 request arrives)

tests/50federation/30room-join.pl
```diff
@@ -839,10 +839,12 @@ sub assert_is_valid_pdu {
                event => $proto,
             } );
 
+            log_if_fail "make_join resp", $proto;
+
             Future->done;
          }),
 
-         $inbound_server->await_request_send_join( $room_id )->then( sub {
+         $inbound_server->await_request_v2_send_join( $room_id )->then( sub {
```
richvdh: again, /v2/send_join isn't specced yet, so this seems premature at best.

tests/50federation/40devicelists.pl
```diff
@@ -266,6 +266,119 @@
...
 
+test "Server correctly resyncs when server leaves and rejoins a room",
+   requires => [ $main::INBOUND_SERVER, federated_rooms_fixture() ],
+
+   check => sub {
+      my ( $inbound_server, $user, $federated_user_id, $room ) = @_;
+
+      # At first the server shares a room with the federated user, who at that
+      # point has a single device. The server will then leave and then rejoin
+      # the room, in the mean time the federated user has added a device, but
```
richvdh:
```suggestion
      # the room. In the mean time the federated user has added a device, but
```

erikjohnston:
Ok, I've tried to move stuff back to not relying on `/v2` being called. I don't know if adding a new `await_request_v1_send_join_reject_v2` is the right approach going forwards, but it at least works for now

lib/SyTest/Federation/Server.pm
```diff
@@ -545,4 +557,40 @@ sub on_event
    return 1;
 }
 
+=head2 await_request_v1_send_join_reject_v2
+
+   $fut = $server->await_request_v1_send_join_reject_v2( $room_id )
```
richvdh: could do with documenting what the $fut returns:

```
my ( $request, $room_id, $event_id ) = $fut->get;
```

tests/50federation/30room-join.pl
```diff
@@ -859,7 +848,6 @@ sub assert_is_valid_pdu {
             @auth_chain = grep { $_->{type} ne 'm.room.create' } @auth_chain;
 
             $req->respond_json(
-               # /v1/send_join has an extraneous [200, ...] wrapper (see MSC1802)
```
richvdh: this is still valid...

PR #794, opened 2020-01-31
Remove test which uses unspecced /events?dir=b

richvdh:
I've got annoyed with trying to make this test pass. It uses "/events?dir=b"
which isn't even specced, but it seems to expect it to block until an event
arrives, which is (a) strange and (b) unreliable in a worker setup.

### DIFF
--- a/tests/30rooms/30history-visibility.pl
+++ b/tests/30rooms/30history-visibility.pl
@@ -96,8 +96,8 @@ sub test_history_visibility
 }
 
 foreach my $i (
-   [ "Guest", sub { guest_user_fixture( with_events => 1 ) } ],
-   [ "Real", sub { local_user_fixture( with_events => 1 ) } ]
+   [ "Guest", sub { guest_user_fixture() } ],
+   [ "Real", sub { local_user_fixture() } ]
 ) {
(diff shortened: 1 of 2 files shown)


tests/30rooms/30history-visibility.pl
```diff
@@ -96,8 +96,8 @@ sub test_history_visibility
 }
 
 foreach my $i (
-   [ "Guest", sub { guest_user_fixture( with_events => 1 ) } ],
-   [ "Real", sub { local_user_fixture( with_events => 1 ) } ]
+   [ "Guest", sub { guest_user_fixture() } ],
```
richvdh: guest_user_fixture doesn't do anything with its params

tests/30rooms/30history-visibility.pl
```diff
@@ -96,8 +96,8 @@ sub test_history_visibility
...
+   [ "Real", sub { local_user_fixture() } ]
```
richvdh: `local_user_fixture` *does* take a with-events param: it makes it call `/events` to initialise an `eventstream_token` field in the user... which we weren't using.

PR #797, opened 2020-02-06
Don't push Dendrite images to Docker Hub

neilalexander:
We use autobuild now for the Dendrite Sytest images so remove them from `push.sh`. We also don't need the Go 1.11 images anymore either.


neilalexander:
@erikjohnston We are requiring Go 1.13 going forward so no longer any need to test Go 1.11.

PR #803, opened 2020-02-14
Add whitelist information to dendrite buildkite annotation

anoadragon453:
Takes the output of dendrite's `show-expected-fail-tests.sh` script and appends it to the buildkite annotation output.

This allows us to see which tests we need to add/remove to/from the whitelist/blacklist without having to dig down in the logs.


scripts/dendrite_sytest.sh
```diff
@@ -62,6 +62,9 @@ rsync -r --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /
 if [ $TEST_STATUS -ne 0 ]; then
     # Build the annotation
     perl /sytest/scripts/format_tap.pl /logs/results.tap "$BUILDKITE_LABEL" >/logs/annotate.md
+    # If show-expected-fail-tests logged something, put it in the annotation
+    # We want nice output though, so surround the text output in code tags
```
richvdh: > surround the text output in code tags

where does this happen?

anoadragon453: Wow, this totally fell off my radar.

It's supposed to happen inside of `format_tap.pl`, which was *probably* supposed to be part of this PR, but perhaps I didn't push that change?

Time to go inspect my desktop...

anoadragon453: Aha, no they're here! https://github.com/matrix-org/dendrite/pull/870/files

richvdh: ok, but still, that means this comment is in the wrong place?

scripts/dendrite_sytest.sh
```diff
@@ -52,7 +52,7 @@ fi
 
 # Check for new tests to be added to the test whitelist
 /src/show-expected-fail-tests.sh /logs/results.tap /src/sytest-whitelist \
-    /src/sytest-blacklist || TEST_STATUS=$?
+    /src/sytest-blacklist > /src/test-whitelist.log || TEST_STATUS=$?
```
richvdh, line 55: could this file have a name which better represents what it does?

Also: I'd recommend against writing to `/src`. It means that we'll be writing stuff to somebody's working copy of dendrite, and means we can't run it against a read-only mount of the dendrite source.

scripts/dendrite_sytest.sh
anoadragon453: Ah, yes quite.

PR #805, opened 2020-02-17
Update tests that assumed PL50 for room upgrade, fix others

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/6834

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -505,24 +505,6 @@ test "/upgrade copies important state to the new room",
       }
 
       $f->then( sub {
-         # to make things harder, we now restrict our ability to change each of
-         # those states: the server should make sure it sets up the state
-         # *before* it replicates the PL.
-         matrix_change_room_power_levels(
-            $creator, $room_id, sub {
-               my ( $levels ) = @_;
-               foreach my $k ( keys %STATE_DICT ) {
-                  $levels->{events}->{$k} = 80;
-               }
-               # Modify tombstone requirements to 50 so we are able to upgrade the room
-               $levels->{events}->{"m.room.tombstone"} = 50;
-
-               $levels->{users}->{$creator->user_id} = 50;
-            },
-         );
-      })->then( sub {
-         matrix_sync( $creator );
-      })->then( sub {
          upgrade_room_synced(
             $creator, $room_id,
             new_version => $TEST_NEW_VERSION,
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -505,7 +505,7 @@ test "/upgrade copies important state to the new room",
 
       $f->then( sub {
          matrix_sync( $creator );
-      }->then( sub {
+      })->then( sub {
          upgrade_room_synced(
             $creator, $room_id,
             new_version => $TEST_NEW_VERSION,
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -374,94 +374,6 @@ test "/upgrade copies the power levels to the new room",
       });
    };
 
-test "/upgrade preserves the power level of the upgrading user in old and new rooms",
-   requires => [
-      local_user_and_room_fixtures(),
-      local_user_fixture(),
-      qw( can_upgrade_room_version can_change_power_levels ),
-   ],
-
-   do => sub {
-      my ( $creator, $room_id, $upgrader ) = @_;
-
-      my ( $pl_content, $new_room_id );
-
-      # Note that this test assumes that moderators by default are allowed to upgrade rooms
-
-      matrix_join_room_synced(
-         $upgrader, $room_id
-      )->then( sub {
-         # Make the joined user a moderator
-         matrix_change_room_power_levels(
-            $creator, $room_id, sub {
-               ( $pl_content ) = @_;
-               $pl_content->{users}->{$upgrader->user_id} = JSON::number(50);
-               log_if_fail "PL content in old room", $pl_content;
-            }
-         )
-      })->then( sub {
-         matrix_sync( $upgrader );
-      })->then( sub {
-         upgrade_room_synced(
-            $upgrader, $room_id,
-            expected_event_counts => { 'm.room.power_levels' => 1 },
-            new_version => $TEST_NEW_VERSION,
-         );
-      })->then( sub {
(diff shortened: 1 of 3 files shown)


tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -687,8 +687,10 @@ sub sync_until_user_in_device_list
 
       my ( $room_id, $from_token, $to_token );
 
-      matrix_create_room( $creator,
+      matrix_create_room($creator,
```
erikjohnston:
```suggestion
      matrix_create_room( $creator,
```

tests/30rooms/60version_upgrade.pl
```diff
@@ -510,6 +422,9 @@ sub upgrade_room_synced {
                foreach my $k ( keys %STATE_DICT ) {
                   $levels->{events}->{$k} = 80;
                }
+               # Modify tombstone requirements to 50 so we are able to upgrade the room
+               $levels->{events}->{"m.room.tombstone"} = 50;
```
erikjohnston: Why? Doesn't the creator have 100?

tests/30rooms/60version_upgrade.pl
```diff
@@ -547,6 +462,96 @@ sub upgrade_room_synced {
       });
    };
 
+test "/upgrade preserves the power level of the upgrading user in old and new rooms",
```
erikjohnston: Why move this test?

tests/30rooms/60version_upgrade.pl
anoadragon453: Just a couple lines below we modify creator to be PL50.

tests/30rooms/60version_upgrade.pl
anoadragon453: It looks better lower down the file :D 

(just an accident)

tests/30rooms/60version_upgrade.pl
erikjohnston: Hmm, ok, but why? If its trying to test that moderators can upgrade rooms by default then that sounds we're making the wrong change here.

anoadragon453: Originally moderators were allowed to update rooms by default, and that caused a bug where power levels could split brain. Technically someone could put a room back into this state, so it may still be worth testing?

Originally this PR just deleted this test, but rich thought it good to keep it as it's still a possibility.

Would you modify it in another way?

erikjohnston: Just don't bother changing the creator to a mod? The test name suggests its trying to test that state gets copied over correctly, so testing that they can do so as a mod was probably just a cheeky extra test

anoadragon453: AH. No I was thinking of `"/upgrade preserves the power level of the upgrading user in old and new rooms"`. This is a separate test. Yes, we can just pull the PL50 change out of here.

tests/30rooms/60version_upgrade.pl
```diff
@@ -501,21 +504,6 @@ sub upgrade_room_synced {
...
-            $creator, $room_id, sub {
-               my ( $levels ) = @_;
-               foreach my $k ( keys %STATE_DICT ) {
-                  $levels->{events}->{$k} = 80;
-               }
-               $levels->{users}->{$creator->user_id} = 50;
-            },
-         );
-      })->then( sub {
-         matrix_sync( $creator );
```
erikjohnston: Why are we removing this sync?

tests/30rooms/60version_upgrade.pl
```diff
@@ -386,18 +386,21 @@ sub upgrade_room_synced {
...
-               ( $pl_content ) = @_;
-               $pl_content->{users}->{$upgrader->user_id} = JSON::number(50);
-               log_if_fail "PL content in old room", $pl_content;
-            }
+            ( $pl_content ) = @_;
+            $pl_content->{users}->{$upgrader->user_id} = JSON::number(50);
+
+            # Note that this test assumes that moderators by default are allowed to upgrade rooms
+            # Change the PL rules to allow moderators to send tombstones
+            $pl_content->{events}->{"m.room.tombstone"} = JSON::number(50);
```
erikjohnston: I'd be tempted to instead have the creator upgrade tbh

tests/30rooms/60version_upgrade.pl
```diff
@@ -386,18 +386,21 @@ sub upgrade_room_synced {
...
          $upgrader, $room_id
       )->then( sub {
          # Make the joined user a moderator
          matrix_change_room_power_levels(
             $creator, $room_id, sub {
-               ( $pl_content ) = @_;
-               $pl_content->{users}->{$upgrader->user_id} = JSON::number(50);
-               log_if_fail "PL content in old room", $pl_content;
-            }
+            ( $pl_content ) = @_;
```
erikjohnston: Needs more ident

tests/30rooms/60version_upgrade.pl
anoadragon453: I presumed it was required to sync the power level changes, but perhaps it's there for some other reason?

(In general I'm not often sure why a `matrix_sync` is done except in only obvious scenarios).

erikjohnston: Its so that `matrix_sync_again` only shows changes from that point

anoadragon453: `upgrade_room_synced` syncs again though? But perhaps with an ephemeral copy of the sync token?

erikjohnston: the `_synced` stuff *should* reset the sync token to what it was before

tests/30rooms/60version_upgrade.pl
anoadragon453: I should stick this in a comment, but the bug that this tested for was a moderator getting upgraded to an admin in the old room. So changing the upgrader to be an admin initially may not preserve the test case?

tests/30rooms/60version_upgrade.pl
anoadragon453: Huh! Ok, good to know. I'll add it back.

tests/30rooms/60version_upgrade.pl
erikjohnston: hmm, Ok

anoadragon453: I'll add a comment.

PR #806, opened 2020-02-17
Support running dendrite in sqlite

Kegsay:
As promised. Tested this with a few configurations:
 - `POSTGRES=1` still runs with postgres
 - `POSTGRES=0` runs in sqlite
 - Omitting the env var runs in sqlite.

I felt this was preferable to having another env var `SQLITE` which then becomes confusing for situations like `POSTGRES=1 SQLITE=1`, and this has parity with Synapse afaict. Pipeline PR will be forthcoming after I have fixed the issues sytest has flagged up.

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -102,12 +102,6 @@ sub _get_config
       $db_uri .= sprintf( " sslmode=%s", $db_config{args}->{sslmode} );
    }
 
-   # POSTGRES not set or is 0, use sqlite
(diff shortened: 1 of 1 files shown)


anoadragon453:
@Kegsay Heads up you'll need to merge latest sytest develop into your branch to fix the synapse failures.

Kegsay:
Thanks, I'm going to flip this to WIP and ask for a re-review later.

The problem is that this PR makes a single sqlite database for all microservices, which is fine if it weren't for `database is locked` errors. Rather than enable WAL (incompatible with WASM) or implement a longish `busy_timeout` (icky), it would be better to have each microservice using its own `.db` file. That's what I'll change this code to.

Kegsay:
It's 4 lines now ;) PTAL

PR #807, opened 2020-02-18
Route /publicised_groups/ to client reader

erikjohnston:
This is to test that we can point publicised groups API to client reader.


erikjohnston:
I'm not a huge fan of that given people might add a different API like `/publicised_groups_to_delete` or something

PR #809, opened 2020-02-18
Tests for /rooms/:room_id/aliases

richvdh:
cf https://github.com/matrix-org/synapse/pull/6939

### DIFF
--- a/lib/SyTest/HTTPClient.pm
+++ b/lib/SyTest/HTTPClient.pm
@@ -83,6 +83,8 @@ sub do_request
    my $self = shift;
    my %params = @_;
 
+   croak "must give a method" unless $params{method};
+
    my $uri = $self->full_uri_for( %params );
 
    # Also set verify_mode = 0 to not complain about self-signed SSL certs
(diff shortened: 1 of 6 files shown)


babolivier:
> lgtm, though may be worth adding a test for server admins being able to query the aliases of a room for which they're not in?

IMHO that should happen in the unit tests of the Synapse PR since server admins aren't a Matrix spec thing.

anoadragon453:
> server admins aren't a Matrix spec thing

Very good point. We should probably convert https://github.com/matrix-org/sytest/blob/d0c105b30f0db89ab118f39052e1e3885946d973/tests/48admin.pl to unit tests at some point as well.

richvdh:
> IMHO that should happen in the unit tests of the Synapse PR since server admins aren't a Matrix spec thing.

+1

> We should probably convert https://github.com/matrix-org/sytest/blob/d0c105b30f0db89ab118f39052e1e3885946d973/tests/48admin.pl to unit tests at some point as well.

+1

PR #811, opened 2020-02-19
Fix flaky 50federation/40publicroomlist.pl

clokep:
This ports the changes (done in #354) from 30rooms/70publicroomslist.pl to 50federation/40publicroomlist.pl in order to fix a flaky test.

This test didn't seem to be causing any issues before matrix-org/synapse#6941, but the changes should be OK to merge before that.


clokep:
Most of this is just fixing indentation, so it helps to [look at the diff without whitespace](https://github.com/matrix-org/sytest/pull/811/files?w=1).

PR #812, opened 2020-02-19
Update tests for propagating canonical alias when upgrading a room

clokep:
Expands tests for matrix-org/synapse#6941 by ensuring the alt_aliases get moved when a room is upgraded.

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -679,7 +679,8 @@ test "/upgrade moves aliases to the new room",
          matrix_put_room_state( $creator, $room_id,
             type    => "m.room.canonical_alias",
             content => {
-               alias => $room_alias_1,
+                alias => $room_alias_1,
+                alt_aliases => [$room_alias_2],
             },
          );
       })->then( sub {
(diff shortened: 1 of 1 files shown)


tests/30rooms/60version_upgrade.pl
```diff
@@ -679,7 +679,8 @@ sub upgrade_room_synced {
          matrix_put_room_state( $creator, $room_id,
             type    => "m.room.canonical_alias",
             content => {
-               alias => $room_alias_1,
+                alias => $room_alias_1,
+                alt_aliases => [$room_alias_2],
```
richvdh: our coding style (see https://github.com/matrix-org/sytest/blob/develop/STYLE-GUIDE.rst) suggests: 

```suggestion
                alt_aliases => [ $room_alias_2 ],
```

(plus three-space indents)

I'm certainly no great fan of that style, but we are fairly consistent with it. I don't feel strongly enough to block this PR on it, but fyi for future reference.

clokep: 👍 I'll make sure I didn't miss this in any other places too.

PR #817, opened 2020-02-21
Stop checking for aliases in the public room list.

clokep:
The `/publicRooms` API is going to stop returning the "aliases" key, so SyTest needs to stop looking for it.

Much of the changes here are de-indenting, the [diff looks a bit better without whitespace](https://github.com/matrix-org/sytest/pull/817/files?w=1).

### DIFF
--- a/tests/50federation/40publicroomlist.pl
+++ b/tests/50federation/40publicroomlist.pl
@@ -63,38 +63,32 @@ test "Name/topic keys are correct",
                   my $topic = $room->{topic};
                   my $canonical_alias = $room->{canonical_alias};
 
-                  my $aliases = $room->{aliases};
-                  defined $aliases or next;
+                  foreach my $alias_local (keys %rooms) {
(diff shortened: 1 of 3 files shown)


tests/50federation/40publicroomlist.pl
```diff
@@ -63,38 +63,32 @@
...
 
-                        my $room_config = $rooms{$alias_local};
+                     log_if_fail "Alias", $alias_local;
+                     log_if_fail "Room", $room;
 
-                        log_if_fail "Alias", $alias_local;
-                        log_if_fail "Room", $room;
+                     assert_eq($room->{num_joined_members}, 1, "num_joined_members");
 
-                        assert_eq($canonical_alias, $alias, "canonical alias");
```
clokep: I think it is worth pointing on that this assertion was removed entirely, I do not think there's anything to replace it with.

PR #818, opened 2020-02-25
Add tests for a user with sufficient power level removing an alias.

clokep:
SyTests for the behavior in matrix-org/synapse#6986.

A user should be able to delete an alias if they have sufficient power level (enough to modify the canonical alias event). There's already a separate test for a user not being able to delete aliases if they don't have power levels.

I verified that against a develop of master (i.e without matrix-org/synapse#6986) the added test here fails.

### DIFF
--- a/tests/30rooms/05aliases.pl
+++ b/tests/30rooms/05aliases.pl
@@ -254,8 +254,15 @@ test "Users with sufficient power-level can delete other's aliases",
            content => {},
          )
       })->then( sub {
-         Future->done(1);
-      })
+         my ( $res ) = @_;
+         log_if_fail "Unable to delete alias", $res;
+
+         do_request_json_for(
+            $user,
+            method => "GET",
+            uri  => "/r0/directory/room/$room_alias",
+         );
+      })->main::expect_http_404;
    };
 
 test "Can delete canonical alias",
--- a/tests/30rooms/05aliases.pl
+++ b/tests/30rooms/05aliases.pl
@@ -222,17 +222,24 @@ test "Users can't delete other's aliases",
    };
 
 test "Users with sufficient power-level can delete other's aliases",
-   requires => [ $creator_fixture, $room_fixture, local_user_fixture(), room_alias_fixture(),
+   requires => [ $creator_fixture, local_user_fixture(), room_alias_fixture(),
                  qw( can_create_room_alias )],
 
    do => sub {
-      my ( $user, $room_id, $other_user, $room_alias ) = @_;
+      my ( $user, $other_user, $room_alias ) = @_;
+      my $room_id;
 
-      matrix_change_room_power_levels(
-         $user, $room_id, sub {
-            $_[0]->{users}->{$other_user->user_id} = 100;
-         },
+      matrix_create_and_join_room(
+         [ $user, $other_user ],
       )->then( sub {
+         ( $room_id ) = @_;
+
+         matrix_change_room_power_levels(
+            $user, $room_id, sub {
+               $_[0]->{users}->{$other_user->user_id} = 100;
+            },
+         )
+      })->then( sub {
          do_request_json_for( $user,
            method => "PUT",
            uri    => "/r0/directory/room/$room_alias",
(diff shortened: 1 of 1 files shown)


clokep:
Oh, hmm...this is probably going to fail on Dendrite. How do we deal with features being added to Synapse before they're ported to Dendrite?

clokep:
Looks like the dendrite build passed, so never mind. 👍

anoadragon453:
> Looks like the dendrite build passed, so never mind. +1

I'm afraid it's quite confusing. Dendrite is currently marked as 'soft fail', so if it fails it doesn't prevent merging. On github you see a tick, but if you go to [buildkite](https://buildkite.com/matrix-dot-org/sytest/builds/608#b6f655d4-6a92-447e-8ef0-483beee7373e you'll see it did in fact fail.

tests/30rooms/05aliases.pl
```diff
@@ -226,6 +226,37 @@ sub _test_can_create_and_delete_alias {
       })
    };
 
+test "Users with sufficient power-level can delete other's aliases",
+   requires => [ $creator_fixture, $room_fixture, local_user_fixture(), room_alias_fixture(),
+                 qw( can_create_room_alias )],
+
+   do => sub {
+      my ( $user, $room_id, $other_user, $room_alias ) = @_;
+      my $server_name = $user->http->server_name;
```
richvdh: seems unused?

tests/30rooms/05aliases.pl
```diff
@@ -226,6 +226,37 @@ sub _test_can_create_and_delete_alias {
...
+         },
+      )->then( sub {
+         do_request_json_for( $user,
+           method => "PUT",
+           uri    => "/r0/directory/room/$room_alias",
+
+           content => { room_id => $room_id },
+         )
+      })->then( sub {
+         do_request_json_for( $other_user,
```
richvdh: shouldn't `other_user` be a member of the room?

tests/30rooms/05aliases.pl
```diff
@@ -226,6 +226,37 @@ sub _test_can_create_and_delete_alias {
...
+           method => "DELETE",
+           uri    => "/r0/directory/room/$room_alias",
+
+           content => {},
+         )
+      })->then( sub {
+         Future->done(1);
```
richvdh: this is a no-op

richvdh: maybe we should request the alias with `$user` and check that it has been deleted?

tests/30rooms/05aliases.pl
clokep: Seems to be defined in most of the tests in this file and unused. I'll clean it up.

tests/30rooms/05aliases.pl
clokep: Having a final `Future->done(1);` seems to be common in this file. Should they be removed?

For this particular test we can assert something here, yeah.

tests/30rooms/05aliases.pl
clokep: It doesn't seem to be a requirement. We can have the user join though if you think it is best.

clokep: This seems to overlap with the comments at https://github.com/matrix-org/synapse/pull/6986#discussion_r384598785. We should sync about the expected behavior.

tests/30rooms/05aliases.pl
clokep: If I remove this then I get an error and the test fails:

```
Expected __ANON__(tests/30rooms/05aliases.pl line 257) to return a Future at /sytest/run-tests.pl line 745.
```

clokep: I figured out a better way to handle this.

tests/30rooms/05aliases.pl
clokep: I added this check!

richvdh:
> Having a final Future->done(1); seems to be common in this file.

For reference: each `then` sub needs to return a (successful) Future. So it's normal for the final `then` block to do its final checks and then return `Future->done` (ie, an already-complete Future). But in this case, the last thing that was happening was a call to `do_request_json_for` - which returns a Future - so there was no need for an extra call to `Future->done`, and your final `then` block was a no-op.

Issue #819, opened 2020-02-26
"guest user can set displaynames" is flaky


richvdh:
eg https://buildkite.com/matrix-dot-org/synapse/builds/7348#f0749e42-bda8-48f9-95fb-3f862aa7fca4

```
# Started: 2020-02-26 09:55:08.990
# Ended: 2020-02-26 09:55:10.021
# Got 7, expected "creeper" for Room displayname at tests/30rooms/13guestaccess.pl line 190.
# 0.167456: Registered new user @anon-20200226_095239-167:localhost:8831
# 0.919001: User @7:localhost:8831 joined room
# { room_id => "!QgjZoiPzZxNvhctfuI:localhost:8831" }
```

clokep:
I saw a couple more instances of this, both on worker runs.

The assertion that fails is:

https://github.com/matrix-org/sytest/blob/128d7d31900abc02767c7572ed19bb88c37590a7/tests/30rooms/13guestaccess.pl#L190

clokep:
Been able to reproduce this locally (although it seems to happen only 1 out of every 4 times or so):

```
docker run --rm -it -e POSTGRES=true -e WORKERS=true -v /Users/clokep/matrix/synapse\:/src:ro -v /Users/clokep/matrix/sytest/logs\:/logs -v /Users/clokep/matrix/sytest\:/sytest:ro matrixdotorg/sytest-synapse:py35 tests/30rooms/13guestaccess.pl
```

clokep:
I attempted to fix this (see 5b1960d727bda09ef980893a70baf2927b48c003), but that seems to have caused a bunch of other tests to have gone flaky.

clokep:
I think the failed tests on ^ were the other failures recently fixed. I rebased and that passed without issue.

PR #821, opened 2020-02-26
Add a test for an event with an invalid signature

clokep:
This reproduces the issue in matrix-org/synapse#6978.

Note that I created this test by copying, so it is a bit duplicative. I'm unsure how much of it is worth keeping.

On current develop (sans matrix-org/synapse#6996) this fails on the `log_if_fail "send_join event", $event;` line.

### DIFF
--- a/tests/50federation/30room-join.pl
+++ b/tests/50federation/30room-join.pl
@@ -959,7 +959,7 @@ test "Outbound federation rejects m.room.create events with an unknown room vers
       )
    };
 
-test "Outbound federation event with an invalid signature should not break",
+test "Event with an invalid signature in the send_join response should not cause room join to fail",
    requires => [ local_user_fixture(), $main::INBOUND_SERVER,
                  federation_user_id_fixture() ],
 
--- a/tests/50federation/30room-join.pl
+++ b/tests/50federation/30room-join.pl
@@ -870,7 +870,7 @@ test "Outbound federation rejects send_join responses with no m.room.create even
 
             # XXX currently synapse fails with a 500 here. I'm not really convinced that's
             # a thing we want to enforce, but we don't really have a specced way to say
-            # "a remote server did something wierd".
+            # "a remote server did something weird".
             Future->done(1);
          }),
       )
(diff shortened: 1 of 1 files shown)


clokep:
These fail, as expected, until matrix-org/synapse#6996 is merged into develop.

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,117 @@ sub assert_is_valid_pdu {
...
             # XXX currently synapse fails with a 500 here. I'm not really convinced that's
             # a thing we want to enforce, but we don't really have a specced way to say
-            # "a remote server did something wierd".
+            # "a remote server did something weird".
+            Future->done(1);
+         }),
+      )
+   };
+
+test "Outbound federation event from other origin",
```
richvdh: some more words about what we're testing here, maybe?

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,117 @@ sub assert_is_valid_pdu {
...
+   requires => [ local_user_fixture(), $main::INBOUND_SERVER,
+                 federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $user, $inbound_server, $creator_id ) = @_;
+
+      my $versionprefix     = "v2";
```
richvdh: unused?

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,117 @@ sub assert_is_valid_pdu {
...
+      # The event was generated with a valid signature, modify it with an
+      # invalid signature.
+      $event->{origin} = "other-server:12345";
+
+      my $await_request_send_join;
+
+      $await_request_send_join = $inbound_server->await_request_v2_send_join( $room_id );
+
+      Future->needs_all(
+         # Await PDU?
```
richvdh: what does this mean?

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,117 @@ sub assert_is_valid_pdu {
...
+         type        => "test",
+         room_id     => $room_id,
+         state_key   => "",
+         content     => {
+            body    => "Test",
+         },
+      );
+
+      # The event was generated with a valid signature, modify it with an
+      # invalid signature.
```
richvdh: this isn't changing the signature so much as invalidating the existing one

clokep: Yes, I failed to update this when modifying the code.

tests/50federation/30room-join.pl
clokep: Uh, this came across in copying and pasting. I'll remove it.

clokep:
Next run of these should succeed since the fix is now merged to develop. 👍

clokep:
This is now hitting #819.

clokep:
This is ready for review, but includes the changes from #823 to fix the build. I can rebase them out if that gets merged first.

tests/50federation/30room-join.pl
clokep: I updated this. 👍

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,110 @@ sub assert_is_valid_pdu {
...
             # XXX currently synapse fails with a 500 here. I'm not really convinced that's
             # a thing we want to enforce, but we don't really have a specced way to say
-            # "a remote server did something wierd".
+            # "a remote server did something weird".
+            Future->done(1);
+         }),
+      )
+   };
+
+test "Outbound federation event with an invalid signature should not break",
```
richvdh: I'd suggest being more specific about where we are inserting the broken event:

```suggestion
test "Event with an invalid signature in the send_join response should not cause room join to fail",
```

tests/50federation/30room-join.pl
```diff
@@ -953,7 +953,110 @@ sub assert_is_valid_pdu {
...
+   requires => [ local_user_fixture(), $main::INBOUND_SERVER,
+                 federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $user, $inbound_server, $creator_id ) = @_;
+
+      my $local_server_name = $inbound_server->server_name;
+      my $datastore         = $inbound_server->datastore;
+
+      my $room_alias = "#50fed-room-alias:$local_server_name";
```
richvdh: this probably needs to be unique; I think this alias is used elsewhere in the test suite?

clokep: It probably should be. Is there any pattern or something I should use here? Or should I be using the fixture that generates a room alias?

richvdh: well there was some idea of making the alias reflect the name of the test file, but clearly it's not very closely followed.

The trouble with the fixture (iirc) is that it wants to generate an alias that is suitable for the system under test rather than for the sytest federation server.

clokep: I made it unique then.

PR #822, opened 2020-03-02
Fix race in self signing key tests.

erikjohnston:
Though this does raise a bit of a scary thought: there's no guarantee that just because we've told clients about a device change doesn't mean the rest of the workers know about it yet


clokep:
I think the code looks good, but I'm not confident in evaluating the implications of what you've said above:

> there's no guarantee that just because we've told clients about a device change doesn't mean the rest of the workers know about it yet

erikjohnston:
Merging to fix flakey test. We should think about scary thoughts separately.

PR #823, opened 2020-03-02
Fix frontend proxy worker binding to wrong port.

erikjohnston:


clokep:
Looks like the builds are passing with this!

erikjohnston:
Thanks Patrick! I've tidied up some related bits and pieces while we're here

scripts/synapse_sytest.sh
```diff
@@ -16,15 +16,26 @@ cd "$(dirname $0)/.."
...
+    sed -i -r "s/^max_connections.*$/max_connections = 500/" /var/run/postgresql/data/postgresql.conf
+
+    echo -e "fsync = off" >> /var/run/postgresql/data/postgresql.conf
+    echo -e "full_page_writes = off" >> /var/run/postgresql/data/postgresql.conf
 
     # Start the database
     su -c 'eatmydata /usr/lib/postgresql/*/bin/pg_ctl -w -D $PGDATA start' postgres
 
+    su -c psql postgres <<< "show config_file"
+    su -c psql postgres <<< "show max_connections"
```
clokep: I realized we could `grep` for the expected values here if we wanted to enforce failure if the configuration isn't as expected. Something like `su -c grep 500 $(psgl postgres <<< "show max_connections")`...not 100% sure that'll work with the `su` though.

clokep:
@erikjohnston I think the changes look OK, not 100% sure about the `worker_replication_http_port` changes, but build still passes. 👍

scripts/synapse_sytest.sh
erikjohnston: Yup, we could, but I think just printing the values for now is Good Enough. If it breaks again at least we will have some debugging info

erikjohnston:
The `worker_replication_http_port` changes just brings all worker configs into line.

PR #829, opened 2020-03-11
Fix the CAS login test

Naugrimm:
see https://github.com/matrix-org/synapse/pull/6634

Signed-off-by: Erik <erik.gto@googlemail.com>

### DIFF
--- a/tests/12login/02cas.pl
+++ b/tests/12login/02cas.pl
@@ -1,3 +1,5 @@
+use URI::Escape;
+
 sub wait_for_cas_request
 {
    my ( $expected_path, %params ) = @_;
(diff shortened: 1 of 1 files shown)


tests/12login/02cas.pl
```diff
@@ -109,12 +111,12 @@ sub wait_for_cas_request
    do => sub {
       my ( $http, $homeserver_info ) = @_;
 
-      my $HS_URI = $homeserver_info->client_location;
-
       # the redirectUrl we send to /login/cas/redirect, which is where we
       # hope to get redirected back to
       my $REDIRECT_URL = "https://client?p=http%3A%2F%2Fserver";
 
+      my $HS_URI = $homeserver_info->client_location . "/_matrix/client/r0/login/cas/ticket?redirectUrl=" . uri_escape($REDIRECT_URL);
```
richvdh: should we not just check that the service provided to `cas/proxyValidate` matches that given to `cas/login`, rather than hardcoding what we expect it to be?

PR #830, opened 2020-03-11
Add tests where the session is abused for interactive auth.

clokep:
This attempts to make some (currently failing) test cases where the session is abused to change the item being acted upon between initiating ui-auth and completing it.
matrix-org/synapse#7068


clokep:
This will need matrix-org/synapse#7068 merged before passing. I failed to name the branches the same. Oops!

PR #834, opened 2020-03-18
Use r0 instead of 'unstable' for joined_members|rooms

Kegsay:
Dendrite is implementing these APIs and these APIs are already in
the r0 spec, so we should be testing against r0 and not unstable.


Kegsay:
LOOK AT HOW GREEN IT IS

Kegsay:
Thanks!

PR #844, opened 2020-03-25
Add test for updating room aliases on upgraded room join

anoadragon453:
Synapse PR: https://github.com/matrix-org/synapse/pull/6946

Having a `remote_room_alias_fixture` method seemed useful.

Test `/upgrade moves aliases to the new room` was edited as the `$server_name` variable was not used.

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -356,7 +356,7 @@ sub room_alias_name_fixture
 
 =head2 room_alias_fixture
 
-   $fixture = room_alias_fixture( prefix => $prefix, remote => false )
+   $fixture = room_alias_fixture( prefix => $prefix, remote => 0 )
 
 Returns a new Fixture, which when provisioned will allocate a name for a new
 room alias on the first homeserver, and return it as a string. Note that this
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -318,7 +318,7 @@ sub matrix_create_room
    });
 }
 
-push @EXPORT, qw( room_alias_name_fixture room_alias_fixture matrix_create_room_synced );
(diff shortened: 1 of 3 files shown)


tests/10apidoc/30room-create.pl
```diff
@@ -356,7 +356,7 @@ sub room_alias_name_fixture
 
 =head2 room_alias_fixture
 
-   $fixture = room_alias_fixture( prefix => $prefix )
+   $fixture = room_alias_fixture( prefix => $prefix, info => $info )
```
richvdh: should this be `remote` rather than `info`?

tests/10apidoc/30room-create.pl
```diff
@@ -374,7 +377,8 @@ sub room_alias_fixture
 
    return fixture(
       requires => [
-         room_alias_name_fixture( prefix => $args{prefix} ), $main::HOMESERVER_INFO[0],
+         room_alias_name_fixture( prefix => $args{prefix} ),
+         exists($args{remote}) ? $main::HOMESERVER_INFO[1] : $main::HOMESERVER_INFO[0],
```
richvdh: this will use `HOMESERVER_INFO[1]` even if `remote == 0`.

missing values in a map return `undef` which is falsy, so I think you can just do this to achieve what you want:

```suggestion
         $args{remote} ? $main::HOMESERVER_INFO[1] : $main::HOMESERVER_INFO[0],
```

tests/10apidoc/30room-create.pl
```diff
@@ -366,6 +366,9 @@ =head2 room_alias_fixture
 An optional prefix string can be provided, which will be prepended onto the
 alias name.
 
+An optional remote boolean can be supplied, which will generate an alias for
+the remote homeserver instead.
```
richvdh: I don't think we really talk about the "remote" homeserver anywhere else.

```suggestion
the second homeserver instead.
```

PR #845, opened 2020-03-31
Add tests for SSO + user interactive authentication.

clokep:
This adds tests for using SSO in the user interactive authentication flow. It requires matrix-org/synapse#7186.


tests/10apidoc/13ui-auth.pl
```diff
@@ -0,0 +1,271 @@
...
+            wait_for_cas_request(
+               "/cas/proxyValidate",
+               response => $CAS_SUCCESS,
+            ),
+            $http->do_request_json(
+               method   => "GET",
+               full_uri => $login_uri,
+               max_redirects => 0, # don't follow the redirect
+            ),
+         );
```
richvdh, line 117: since we've got two copies of this, it would be nice to factor it and the common parts of `CAS_SUCCESS` out to a separate function.

tests/10apidoc/13ui-auth.pl
```diff
@@ -0,0 +1,271 @@
...
+      })->then( sub {
+         # Repeat the device deletion, which should now complete.
```
richvdh, line 194: it shouldn't complete, it should give a 403.

tests/10apidoc/13ui-auth.pl
```diff
@@ -0,0 +1,271 @@
...
+         initial_device_display_name => "device display",
+      )->then( sub {
+         # Initiate the interactive authentication session with the first device.
+         matrix_delete_device( $user, $DEVICE_ID, {} );
+      })->main::expect_http_401->then( sub {
+         my ($resp) = @_;
+
+         my $body = decode_json $resp->content;
+
+         log_if_fail("Response to empty body", $body);
```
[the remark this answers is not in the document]
clokep: Interesting, this was copied-and-pasted from another test. 😢

tests/10apidoc/13ui-auth.pl
clokep: That makes sense! I'll need to figure out how to do that, but shouldn't be too hard.

tests/10apidoc/13ui-auth.pl
```diff
@@ -21,11 +24,35 @@ sub wait_for_cas_request
...
+         method   => "GET",
+         full_uri => $login_uri,
+         max_redirects => 0, # don't follow the redirect
+      ),
+   );
+}
+
 test "Interactive authentication types include SSO",
-   requires => [ local_user_fixture( with_events => 0 ), $main::API_CLIENTS[0] ],
+   requires => [ local_user_fixture( with_events => 0 ) ],
```
richvdh, line 52: 0 is the default for with_events.

```suggestion
   requires => [ local_user_fixture() ],
```

likewise elsewhere.

tests/10apidoc/13ui-auth.pl
```diff
@@ -21,11 +24,35 @@ sub wait_for_cas_request
    });
 }
 
+# Generate a ticket request from the client to the homeserver (and "validate"
+# it with the CAS server).
```
richvdh, line 28: it looks like this assumes the client already has the ticket (it's passed in `ticket`) and now it's submitting it.

Also we don't validate it with the CAS server, since we *are* the CAS server. 

```suggestion
# Generate a ticket-submission request from the client to the homeserver.
#
# Waits for the validation request from the homeserver, and returns the given response.
```

clokep: Right, that's why validate was in quotes, but definitely not clear enough! Thanks for the improved wording. 👍

clokep:
Looks like rebasing this got the other test passing. 👍 This is ready for review!

tests/10apidoc/13ui-auth.pl
```diff
@@ -0,0 +1,269 @@
+use JSON qw( decode_json );
+use URI::Escape;
+
+our @EXPORT = qw( wait_for_cas_request );
+
+# A convenience function which wraps await_http_request. It returns a successful
+# CAS response when queried for a particular path.
```
richvdh, line 7: it'd be nice to document what the `params` can be, but :man_shrugging:

clokep: I added some.

PR #851, opened 2020-04-10
Make message history pagination test less flaky

Cnly:
Dendrite has a [failing test](https://buildkite.com/matrix-dot-org/dendrite/builds/1145#annotation-SyTest%20-%20:go:%201.13%20/%20:postgres:%209.6) that gets fixed when some waiting is added before the `matrix_get_room_messages` call here.

Signed-off-by: Alex Chen <minecnly@gmail.com>

(Not actually a Perl user but thought this should be rather simple; please feel free to point out problems.)

### DIFF
--- a/tests/30rooms/04messages.pl
+++ b/tests/30rooms/04messages.pl
@@ -215,14 +215,14 @@ test "Message history can be paginated",
             body => "Message number $_[0]"
          )
       } foreach => [ 1 .. 20 ] )->then( sub {
-         await_sync( $user, check => sub {
-            sync_timeline_contains( $_[0], $room_id, sub {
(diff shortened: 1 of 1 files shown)


tests/30rooms/04messages.pl
```diff
@@ -215,6 +215,15 @@
             body => "Message number $_[0]"
          )
       } foreach => [ 1 .. 20 ] )->then( sub {
+         await_sync( $user, check => sub {
+            sync_timeline_contains( $_[0], $room_id, sub {
```
anoadragon453, line 219: You could probably use `await_sync_timeline_contains` instead of these two methods, right?

Otherwise looks fine to me.

PR #853, opened 2020-04-15
Modify login tests to be spec compliant

Kegsay:
The 'identifier' key is required as per:
  https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-login

Synapse likely falls back to previous logic if this key is missing
but Dendrite will just fail, correctly stating that the identifier
key is missing.

This is currently preventing the login tests from succeeding on Dendrite.

### DIFF
--- a/tests/10apidoc/02login.pl
+++ b/tests/10apidoc/02login.pl
@@ -82,7 +82,10 @@ test "POST /login can log in as a user",
 
          content => {
             type     => "m.login.password",
-            user     => $user_id,
+            identifier => {
+               type => "m.id.user",
+               user => $user_id,
+            },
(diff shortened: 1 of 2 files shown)


clokep:
Not necessarily a question for @Kegsay, but does this imply we should file an issue with Synapse to remove this fallback code or just remove it? Not sure how long it has been required.

Kegsay:
No idea, it's up to the Synapse guys to decide if they want to break backwards compat with older clients. I would probably remove the code personally, but it's not my call.

clokep:
I believe this is already covered by matrix-org/synapse#5665 actually.

PR #856, opened 2020-04-20
Test for sending to_device messages to multiple recipients at once

richvdh:
I thought we had a bug here, but it turns out we don't. Still, may as well keep
the test.

### DIFF
--- a/tests/46direct/01directmessage.pl
+++ b/tests/46direct/01directmessage.pl
@@ -72,7 +72,7 @@ sub matrix_recv_device_message
       return 1 if $f->failure;
       my $resp = $f->get;
       log_if_fail "Sync response", $resp;
-      scalar @{ $resp->{to_device}{events} };
+      return scalar @{ $resp->{to_device}{events} };
    };
 
    $f->then( sub {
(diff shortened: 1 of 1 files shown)


tests/46direct/01directmessage.pl
```diff
@@ -70,7 +70,9 @@ sub matrix_recv_device_message
    } until => sub {
       my ( $f ) = @_;
       return 1 if $f->failure;
-      scalar @{ $f->get->{to_device}{events} };
+      my $resp = $f->get;
+      log_if_fail "Sync response", $resp;
+      scalar @{ $resp->{to_device}{events} };
```
anoadragon453, line 75: I know `scalar` gets the length of an array in perl, but why are we doing that here without storing the value at all?

Or does the value here get returned?

Probably touches on the point that sytest could use more comments.

tests/46direct/01directmessage.pl
```diff
@@ -153,3 +155,63 @@ sub matrix_recv_and_ack_device_message
          Future->done(1);
       });
    };
+
+test "Can send to a to-device message to two users which both receive it using /sync",
```
anoadragon453, line 159:
```suggestion
test "Can send a to-device message to two users which both receive it using /sync",
```

tests/46direct/01directmessage.pl
```diff
@@ -153,3 +155,63 @@ sub matrix_recv_and_ack_device_message
...
+   requires => [ local_user_fixture(), local_user_fixture(), local_user_fixture(), qw( can_recv_device_message ) ],
+
+   check => sub {
+      my ( $sender, $recip1, $recip2 ) = @_;
+
+      # do intial syncs for each recipient
```
anoadragon453, line 165:
```suggestion
      # do initial syncs for each recipient
```

tests/46direct/01directmessage.pl
richvdh: yeah, in perl, the result of a block is the result of the last expression in that block, so there's an implicit `return` here.

> Probably touches on the point that sytest could use more comments.

possibly, although I think it's better to try to write code in a way that's clear. In this case I think an explicit `return` is preferable. In general I think `sytest` relies too much on implicit `return`.

richvdh, line 75:
```suggestion
      return scalar @{ $resp->{to_device}{events} };
```

PR #857, opened 2020-04-22
Add support for running worker mode against redis

erikjohnston:
This adds a `--redis-host` command.  c.f. https://github.com/matrix-org/synapse/pull/7040

### DIFF
--- a/scripts/synapse_sytest.sh
+++ b/scripts/synapse_sytest.sh
@@ -121,7 +121,7 @@ if [ -n "$OFFLINE" ]; then
 else
     # We've already created the virtualenv, but lets double check we have all
     # deps.
-    /venv/bin/pip install -q --upgrade --no-cache-dir /src
+    /venv/bin/pip install -q --upgrade --no-cache-dir /src[redis]
     /venv/bin/pip install -q --upgrade --no-cache-dir \
         lxml psycopg2 coverage codecov tap.py coverage_enable_subprocess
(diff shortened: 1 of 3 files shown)


scripts/synapse_sytest.sh
```diff
@@ -121,7 +121,7 @@ if [ -n "$OFFLINE" ]; then
 else
     # We've already created the virtualenv, but lets double check we have all
     # deps.
-    /venv/bin/pip install -q --upgrade --no-cache-dir /src
+    /venv/bin/pip install -q --upgrade --no-cache-dir /src[redis]
```
clokep, line 124: Do we want to be testing a setup with redis specified in addition?

erikjohnston: Basically the situation is that this happens before we call sytest and so don't have access to the config options. We could:

1. Just always install redis deps, the downside is that sytest won't catch if we inadvertently require redis deps when redis isn't configured.
2. Add an env var to configure whether we install redis deps, downside is that it adds more options.

Personally, I'm not sure sytest is the place to test whether optional deps really are optional, especially if we end up wanting to add tests for optional features in sytest we'd need to keep adding more sytest runs (though arguably we shouldn't be testing optional features in sytest).

clokep: That makes sense. Thanks! 👍

Issue #850, opened 2020-04-10
Possible missing dependency declaration: Test::Identity?


Cnly:
When running `./install-deps.pl` on Debian Testing, the script fails at:

```
<omitted>
Running Build test for PEVANS/IO-Async-SSL-0.22.tar.gz
t/00use.t ............... ok   
t/01upgrade.t ........... Can't locate Test/Identity.pm in @INC (you may need to install the Test::Identity module) (@INC contains: <omitted>
<omitted>
Stopping: 'install' failed for 'IO::Async::SSL'.
<omitted>
```

Running `cpan install Test::Identity` (as instructed above) and then `./install-deps.pl` again fixes the problem.

Does `Test::Identity` need to be in `cpanfile`?

clokep:
I don't see any use of `Test::Identity` in SyTest itself, it looks like it might be required by the `IO::Async::SSL` package though? Could it be a bug there?

anoadragon453:
I'm installing a local debian testing box to test.

richvdh:
yes, if anything, that sounds like a bug in `IO::Async::SSL` which should be raised on its bug tracker.

anoadragon453:
Interestingly, I didn't run into this problem on a fresh install of Debian testing. Problems I did run into though that are poorly documented:

* Installing `build-essential` is required on Debian/Ubuntu. Not sure exactly which package though but the error was related to `make`
* Installing postgres build dependencies are required: `sudo apt install libpq-dev`
* `cpan install File:Which` needed to be done manually

anoadragon453:
Yeah, `Test::Identity` was automatically installed for me as part of running `./install-deps.pl`.

richvdh:
sounds like we can close this?

PR #859, opened 2020-04-27
Mention required debian and cpan dependencies

anoadragon453:
This PR adds the missing dependencies that were necessary from a fresh Debian 9 install, listed in https://github.com/matrix-org/sytest/issues/850#issuecomment-614127796.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -28,6 +28,7 @@ requires 'Email::MIME';
 requires 'File::Basename';
 requires 'File::Path';
 requires 'File::Slurper';
+requires 'File::Which';
 requires 'Future', '>= 0.33';
 requires 'Getopt::Long';
 requires 'IO::Async', '>= 0.69';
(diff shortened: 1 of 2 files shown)


cpanfile
```diff
@@ -28,6 +28,7 @@ requires 'Email::MIME';
 requires 'File::Basename';
 requires 'File::Path';
 requires 'File::Slurper';
+requires 'File::Which';
```
clokep, line 31: Similar to #850 sytest doesn't seem to use this directly?

anoadragon453: Ah, good point. Looks like it's [Alien::Base](https://metacpan.org/pod/Alien::Base) that needs it.

But it's listed on cpan... so why isn't it installed? :thinking:

anoadragon453: It's not immediately obvious to the team, so I'll give this another run in a fresh VM, get some logs and see if we can track it down.

anoadragon453: Running this again in a fresh VM and installing the native dependencies mentioned in this PR, I received no dependency issues and was able to do a full successful sytest run.

I'll update the PR with only those.

PR #863, opened 2020-05-06
Add event ID to /v1/invite and /v2/invite on Sytest federation server

neilalexander:
According to [the spec](https://matrix.org/docs/spec/server_server/r0.1.3#put-matrix-federation-v2-invite-roomid-eventid), there should be an `:event_id` parameter here.


neilalexander:
Apparently this breaks test 595 so, yeah. Closing.

PR #866, opened 2020-05-07
Fix installing from a readonly copy of synapse

richvdh:
It seems that, as of pip 20.1, pip likes to do an in-tree build of synapse when
you install it, which is pretty nasty when pip is running in a docker image.

We can work around this by copying the bits we need into a tarball and
installing from that. Ultimately I'd like to change the build pipelines to
build a wheel upfront, and use that for the sytest builds, which will be a more
accurate representation of how most people install synapse, but this is a
prerequisite.

See also: https://discuss.python.org/t/installing-from-a-read-only-source/4115/3

### DIFF
--- a/scripts/synapse_sytest.sh
+++ b/scripts/synapse_sytest.sh
@@ -25,11 +25,6 @@ if [ -n "$MULTI_POSTGRES" ] || [ -n "$POSTGRES" ]; then
 
     # Start the database
     su -c 'eatmydata /usr/lib/postgresql/*/bin/pg_ctl -w -D $PGDATA start' postgres
-
-    #su -c psql postgres <<< "show config_file"
-    #su -c psql postgres <<< "show max_connections"
-    #su -c psql postgres <<< "show full_page_writes"
-    #su -c psql postgres <<< "show fsync"
 fi
 
 # Now create the databases
@@ -116,7 +111,7 @@ fi
 # if we're running against a source directory, turn it into a tarball.  pip
 # will then unpack it to a temporary location, and build it.  (As of pip 20.1,
 # it will otherwise try to build it in-tree, which means writing changes to the
-# source volume outside the container.
+# source volume outside the container.)
 #
 if [ -d "$SYNAPSE_SOURCE" ]; then
     echo "Creating tarball from synapse source"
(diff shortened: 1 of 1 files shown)


richvdh:
(worth noting that this only becomes a problem if/when we rebuild the sytest-synapse docker images)

scripts/synapse_sytest.sh
```diff
@@ -26,10 +26,10 @@ if [ -n "$MULTI_POSTGRES" ] || [ -n "$POSTGRES" ]; then
...
     su -c 'eatmydata /usr/lib/postgresql/*/bin/pg_ctl -w -D $PGDATA start' postgres
 
-    su -c psql postgres <<< "show config_file"
-    su -c psql postgres <<< "show max_connections"
-    su -c psql postgres <<< "show full_page_writes"
-    su -c psql postgres <<< "show fsync"
+    #su -c psql postgres <<< "show config_file"
+    #su -c psql postgres <<< "show max_connections"
+    #su -c psql postgres <<< "show full_page_writes"
+    #su -c psql postgres <<< "show fsync"
```
clokep, line 32: Should we remove these if they're not useful?

scripts/synapse_sytest.sh
```diff
@@ -110,6 +110,24 @@ elif [ -n "$POSTGRES" ]; then
 
 fi
 
+# default value for SYNAPSE_SOURCE
+: ${SYNAPSE_SOURCE:=/src}
+
+# if we're running against a source directory, turn it into a tarball.  pip
+# will then unpack it to a temporary location, and build it.  (As of pip 20.1,
+# it will otherwise try to build it in-tree, which means writing changes to the
+# source volume outside the container.
```
clokep, line 119:
```suggestion
# source volume outside the container.)
```

scripts/synapse_sytest.sh
```diff
@@ -110,6 +110,24 @@ elif [ -n "$POSTGRES" ]; then
 
 fi
 
+# default value for SYNAPSE_SOURCE
+: ${SYNAPSE_SOURCE:=/src}
```
clokep, line 114: Neat, did not know this about bash. (https://wiki.bash-hackers.org/syntax/pe#assign_a_default_value if anyone else doesn't know what this does)

scripts/synapse_sytest.sh
richvdh: done in 0165ef9b

PR #869, opened 2020-05-14
Room version 6 tests

clokep:
This adds some sytests for room version 6, mostly geared around the integer validation of [MSC2540](https://github.com/matrix-org/matrix-doc/pull/2540).

This requires matrix-org/synapse#7506

### To Do:

Bad event sent to:
* [x] `/_matrix/federation/v1/send`
* [x] `/_matrix/federation/vN/send_join`
* [x] `/_matrix/federation/vN/send_leave`
* [x] `/_matrix/federation/vN/invite`

Bad event returned from:
* [x] `/_matrix/federation/v1/backfill/{roomId}`
* [x] `/_matrix/federation/v1/get_missing_events/{roomId}`

### DIFF
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -16,8 +16,7 @@ use SyTest::Assertions qw( :all );
 use URI::Escape qw( uri_escape );
 
 use constant SUPPORTED_ROOM_VERSIONS => [qw(
-   1 2 3 4 5
-   org.matrix.msc2260
+   1 2 3 4 5 6
 )];
 
 sub configure
--- a/tests/30rooms/08levels.pl
+++ b/tests/30rooms/08levels.pl
@@ -7,6 +7,8 @@ my $user_fixture = local_user_fixture();
 
 sub lockeddown_room_fixture
 {
+   my ( %options ) = @_;
+
    fixture(
       requires => [ $creator_fixture, $user_fixture,
                     qw( can_change_power_levels ) ],
@@ -14,7 +16,7 @@ sub lockeddown_room_fixture
       setup => sub {
          my ( $creator, $test_user ) = @_;
 
-         matrix_create_and_join_room( [ $creator, $test_user ] )
+         matrix_create_and_join_room( [ $creator, $test_user ], %options )
          ->then( sub {
             my ( $room_id ) = @_;
 
@@ -157,3 +159,24 @@ foreach my $levelname (qw( ban kick redact )) {
          })->SyTest::pass_on_done( "Fails at setting 75" );
       };
 }
+
+multi_test "Users cannot set notifications powerlevel higher than their own",
+   requires => [ $creator_fixture, $user_fixture, lockeddown_room_fixture( room_version => "6" ),
+                 qw( can_change_power_levels )],
+
+   do => sub {
+      my ( $user, undef, $room_id ) = @_;
+
+      matrix_change_room_power_levels( $user, $room_id, sub {
+         my ( $levels ) = @_;
+
+         $levels->{notifications}{room} = 25;
+      })->SyTest::pass_on_done( "Succeeds at setting 25" )
+      ->then( sub {
+         matrix_change_room_power_levels( $user, $room_id, sub {
+            my ( $levels ) = @_;
+
+            $levels->{notifications}{room} = 10000000;
+         })->main::expect_http_403
+      })->SyTest::pass_on_done( "Fails at setting 75" );
+   };
--- a/tests/30rooms/08levels.pl
+++ b/tests/30rooms/08levels.pl
@@ -156,7 +156,7 @@ foreach my $levelname (qw( ban kick redact )) {
 
                $levels->{$levelname} = 10000000;
             })->main::expect_http_403
-         })->SyTest::pass_on_done( "Fails at setting 75" );
+         })->SyTest::pass_on_done( "Fails at setting 10000000" );
       };
 }
 
@@ -178,5 +178,5 @@ multi_test "Users cannot set notifications powerlevel higher than their own",
 
             $levels->{notifications}{room} = 10000000;
          })->main::expect_http_403
-      })->SyTest::pass_on_done( "Fails at setting 75" );
+      })->SyTest::pass_on_done( "Fails at setting 10000000" );
    };
--- a/tests/50federation/30room-join.pl
+++ b/tests/50federation/30room-join.pl
@@ -1061,3 +1061,85 @@ test "Event with an invalid signature in the send_join response should not cause
          }),
       )
    };
+
+test "Room version 6 rejects invalid JSON",
+   requires => [ $main::OUTBOUND_CLIENT, $main::INBOUND_SERVER,
+                 local_user_and_room_fixtures( room_opts => { room_version => "6" } ),
+                 federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $outbound_client, $inbound_server, $creator, $room_id, $user_id ) = @_;
+      my $first_home_server = $creator->server_name;
+
+      my $local_server_name = $outbound_client->server_name;
+      my $datastore         = $inbound_server->datastore;
+
+      $outbound_client->do_request_json(
+         method   => "GET",
+         hostname => $first_home_server,
+         uri      => "/v1/make_join/$room_id/$user_id",
+         params   => {
+            ver => ["6"],
+         },
+      )->then( sub {
+         my ( $body ) = @_;
+
+         log_if_fail "make_join body", $body;
+
+         assert_json_keys( $body, qw( event ));
+
+         my $protoevent = $body->{event};
+
+         assert_json_keys( $protoevent, qw(
+            auth_events content depth room_id sender state_key type
+         ));
+
+         assert_json_nonempty_list( my $auth_events = $protoevent->{auth_events} );
+
+         assert_json_nonempty_list( $protoevent->{prev_events} );
+
+         assert_json_number( $protoevent->{depth} );
+
+         $protoevent->{room_id} eq $room_id or
+            die "Expected 'room_id' to be $room_id";
+         $protoevent->{sender} eq $user_id or
+            die "Expected 'sender' to be $user_id";
+         $protoevent->{state_key} eq $user_id or
+            die "Expected 'state_key' to be $user_id";
+         $protoevent->{type} eq "m.room.member" or
+            die "Expected 'type' to be 'm.room.member'";
+
+         assert_json_keys( my $content = $protoevent->{content}, qw( membership ) );
+         $content->{membership} eq "join" or
+            die "Expected 'membership' to be 'join'";
+
+         my %event = (
+            ( map { $_ => $protoevent->{$_} } qw(
+               auth_events content depth prev_events room_id sender
+               state_key type ) ),
+
+            origin           => $local_server_name,
+            origin_server_ts => $inbound_server->time_ms,
+         );
+         # Insert a "bad" value into the send join, in this case a float.
+         ${event}{contents}{bad_val} = 1.1;
+
+         $datastore->sign_event( \%event );
+
+         $outbound_client->do_request_json(
+            method   => "PUT",
+            hostname => $first_home_server,
+            uri      => "/v2/send_join/$room_id/xxx",
+
+            content => \%event,
+         )
+      })->main::expect_http_400
+      ->then( sub {
+         my ( $response ) = @_;
+         my $body = decode_json( $response->content );
+         log_if_fail "Join error response", $body;
+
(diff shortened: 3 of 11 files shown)


clokep:
We should probably add some sytests to the other added features as well. I'll see how hard that is.

clokep:
Also -- these tests are almost definitely in the wrong file now. :) I'll see if I can find a better one...

clokep:
> We should probably add some sytests to the other added features as well. I'll see how hard that is.

* Looks like tests for MSC2209 should be added to `tests/30rooms/08levels.pl`
* It looks like there's already tests for the changes in alias behavior.

lib/SyTest/Federation/Client.pm
```diff
@@ -16,8 +16,7 @@ use SyTest::Assertions qw( :all );
 use URI::Escape qw( uri_escape );
 
 use constant SUPPORTED_ROOM_VERSIONS => [qw(
-   1 2 3 4 5
-   org.matrix.msc2260
```
clokep, line 20: This room version was previously removed from Synapse, it also seems a bit weird to bit unstable room versions in here.

tests/80torture/20json.pl
```diff
@@ -27,3 +27,51 @@
...
+               body    => "NaN",
+            },
+         )->followed_by( \&main::expect_http_400 ),
+
+         do_request_json_for( $user,
+            method  => "PUT",
+            uri     => "/r0/rooms/$room_id/send/sytest.dummy",
+            content => {
+               msgtype => "sytest.dummy",
+               body    => "inf",
```
anoadragon453, line 64: This produces a float instead of a string?

tests/30rooms/08levels.pl
```diff
@@ -157,3 +159,24 @@ sub test_powerlevel
...
+
+         $levels->{notifications}{room} = 25;
+      })->SyTest::pass_on_done( "Succeeds at setting 25" )
+      ->then( sub {
+         matrix_change_room_power_levels( $user, $room_id, sub {
+            my ( $levels ) = @_;
+
+            $levels->{notifications}{room} = 10000000;
+         })->main::expect_http_403
+      })->SyTest::pass_on_done( "Fails at setting 75" );
```
anoadragon453, line 181: 10000000 != 75 ?

tests/80torture/20json.pl
clokep: I'm fairly certain this boils down to a "perl is weird", but I'll double check that this is doing what I expect.

clokep: OK, so the tests were broken and not testing what I thought they were... so yeah this is wrong. 😢

clokep: I fixed this in 972e60bac09c1fc4088d4c14d778f1e987aceaa7, but it sends the following JSON now:

```json
{"body":-nan,"msgtype":"sytest.dummy"}
```

Which is invalid, but in a slightly different way than I expected. I think it is OK to leave these tests, but I'm not sure if they really add value beyond the standard float one. Any thoughts?

clokep:
@anoadragon453 Do you have any thoughts on where the JSON tests should go or if the current file is reasonable?

richvdh:
> @anoadragon453 Do you have any thoughts on where the JSON tests should go or if the current file is reasonable?

it seems reasonable to me, fwiw (though I do tend to forget that `80torture` exists).

anoadragon453:
@clokep Seems fine to me as well.

tests/80torture/20json.pl
anoadragon453: Probably nice to still keep, but I would leave a comment explaining that it doesn't produce valid JSON but still serves to test validity code?

clokep: I decided to rewrite these a bit and add some of the Python values which could cause issues. Going to ask you to look at this again, sorry!

richvdh:
(I should say: happy for this to land as-is and the additional tests to end up in a different PR)

clokep:
I'll take a look at those. 👍

clokep:
Before testing `/_matrix/federation/v1/send` we'll need to figure out if it makes sense to do matrix-org/synapse#7543, although the result of that was "it is server implementation dependent", so not sure if it belongs in sytest?

tests/50federation/35room-invite.pl
```diff
@@ -755,3 +754,34 @@ sub do_v2_invite_request
 
    # this test is a bit slooow
    timeout => 20;
+
+test "Inbound federation rejects invites which are not signed by the sender",
```
clokep, line 758: This is wrong, I have it fixed in a subsequent commit.

tests/50federation/33room-get-missing-events.pl
```diff
@@ -413,3 +411,125 @@ sub sytest_user_and_room_fixture {
          Future->done;
       });
    };
+
+test "Outbound federation will ignore a missing event with bad JSON for room version 6",
+   requires => [ $main::OUTBOUND_CLIENT, $main::INBOUND_SERVER,
+                 local_user_and_room_fixtures(
+                    user_opts => { with_events => 1 },
```
richvdh, line 418: unless you specifically want a call to `/events` during registration, you don't need this. I think whereever you cargo-culted it from is wrong...

tests/50federation/33room-get-missing-events.pl
```diff
@@ -413,3 +411,125 @@ sub sytest_user_and_room_fixture {
...
+      my $first_home_server = $creator->server_name;
+
+      my $datastore         = $inbound_server->datastore;
+
+      $outbound_client->join_room(
+         server_name => $first_home_server,
+         room_id     => $room_id,
+         user_id     => $user_id,
+         version     => "6",
+      )->then( sub {
```
richvdh, line 434: it's generally easier to use `federated_rooms_fixture` than separate calls to `local_user_and_room_fixtures`, `federation_user_id_fixture` and `join_room`.

tests/50federation/33room-get-missing-events.pl
```diff
@@ -413,3 +411,125 @@ sub sytest_user_and_room_fixture {
...
+                  limit       => $body->{limit},
+               );
+
+               log_if_fail "Backfilling", @events;
+
+               $req->respond_json( {
+                  events => \@events,
+               } );
+
+               log_if_fail "Done here";
```
richvdh, line 509: is this debug?

tests/50federation/33room-get-missing-events.pl
```diff
@@ -413,3 +411,125 @@ sub sytest_user_and_room_fixture {
...
+
+               Future->done(1);
```
richvdh, line 511: should probably decide if we're using `Future->done;` or `Future->done(1);`. I'd vote for the former, but ymmv

tests/50federation/33room-get-missing-events.pl
```diff
@@ -413,3 +411,125 @@ sub sytest_user_and_room_fixture {
...
+
+               log_if_fail "Send response", $body;
+
+               assert_json_keys( $body, 'pdus' );
+               # 'pdus' is a map from event id to error details.
+               my $pdus = $body->{pdus};
+
+               # Sending the event fails since fetching the event results in
+               # invalid JSON.
+               assert_json_keys( $pdus, $sent_event_id );
```
richvdh, line 529: sorry, are we expecting `sent_event_id` to have an error or not? either way this check doesn't seem to do what it sets out to, because we expect it to be in the result either way.

tests/50federation/34room-backfill.pl
```diff
@@ -447,3 +447,113 @@
...
+test "Outbound federation rejects backfill containing invalid JSON for events in room version 6",
+   requires => [ local_user_fixture(), $main::INBOUND_SERVER, federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $user, $inbound_server, $creator_id ) = @_;
+
+      my $local_server_name = $inbound_server->server_name;
+      my $datastore         = $inbound_server->datastore;
+
+      my $room_alias = "#50fed-31backfill:$local_server_name";
```
richvdh, line 460: why 31?

(this alias needs to be unique, so you might want to consider how that is arranged)

tests/50federation/34room-backfill.pl
```diff
@@ -447,3 +447,113 @@
...
+
+         do_request_json_for( $user,
+            method => "POST",
+            uri    => "/r0/join/$room_alias",
+
+            content => {},
+         )->then( sub {
+            my ( $body ) = @_;
+
+            my $room_id = $body->{room_id};
```
richvdh, line 518: I'd use `matrix_join_room` for this

tests/50federation/34room-backfill.pl
```diff
@@ -447,3 +447,113 @@
...
+
+            # It make take some time for the join to propagate, so we need to
+            # retry on failure.
```
richvdh, line 521: ... or possibly `matrix_join_room_synced`.

As a general rule, tests that just `retry_until_success` are horrible to debug: they never fail outright, they just time out, without any clues as to why they were failing.

tests/50federation/30room-join.pl
```diff
@@ -1061,3 +1061,94 @@ sub assert_is_valid_pdu {
...
+test "Inbound: room version 6 rejects invalid JSON",
+   requires => [ $main::OUTBOUND_CLIENT, $main::INBOUND_SERVER,
+                 local_user_and_room_fixtures( room_opts => { room_version => "6" } ),
+                 federation_user_id_fixture() ],
+
+   do => sub {
+      my ( $outbound_client, $inbound_server, $creator, $room_id, $user_id ) = @_;
+      my $first_home_server = $creator->server_name;
+
+      my $local_server_name = $outbound_client->server_name;
```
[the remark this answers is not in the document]
clokep: It is used to set the `origin` below in the send_join request.

tests/50federation/33room-get-missing-events.pl
clokep: Yes, I'll remove it. In general, I'm unsure how many `log_if_fail` calls are useful / if there's an overhead to them.

tests/50federation/33room-get-missing-events.pl
clokep: Ah, I forgot to ensure that it is an error, yes. I'll clarify what is happening in the comment too.

tests/50federation/33room-get-missing-events.pl
clokep: We seem very inconsistent on this throughout SyTest, but I'll make the new changes be `Future->done;`.

tests/50federation/30room-join.pl
```diff
@@ -1061,3 +1061,94 @@ sub assert_is_valid_pdu {
...
+         $protoevent->{sender} eq $user_id or
+            die "Expected 'sender' to be $user_id";
+         $protoevent->{state_key} eq $user_id or
+            die "Expected 'state_key' to be $user_id";
+         $protoevent->{type} eq "m.room.member" or
+            die "Expected 'type' to be 'm.room.member'";
+
+         assert_json_keys( my $content = $protoevent->{content}, qw( membership ) );
+         $content->{membership} eq "join" or
+            die "Expected 'membership' to be 'join'";
```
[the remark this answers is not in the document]
clokep: I'll remove it and add a note.

tests/50federation/30room-join.pl
```diff
@@ -1061,3 +1061,94 @@ sub assert_is_valid_pdu {
...
+      log_if_fail "Response", $response;
+      my $body = decode_json($response->content);
+      log_if_fail "Body", $body;
+
+      assert_eq($body->{errcode}, "M_BAD_JSON", 'responsecode');
+      Future->done(1);
+   });
+};
+
+test "Inbound: room version 6 rejects invalid JSON",
```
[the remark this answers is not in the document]
clokep: I added some comments describing the overall test, the important bits of configuration and tried to remove some extraneous asserts to show what we really want to test.

tests/50federation/33room-get-missing-events.pl
clokep: I think I caught the potential places this could be replaced...

clokep:
@richvdh I think I've hit all your comments. Thanks for pointing out the additional fixtures! 👍

tests/50federation/34room-backfill.pl
```diff
@@ -448,6 +448,15 @@
...
 
+# A homeserver receiving a response from `backfill` for a version 6 room with a
+# bad JSON value (e.g. a float) should discard the bad data.
+#
+# To test this we need to:
+# * Create a room.
+# * Add "bad" data into the room history.
+# * Join the room.
+# * Attempt to backfill the room history.
+# * Ensure that the "bad" event wil be discarded.
```
richvdh, line 459:
```suggestion
# * Ensure that the "bad" event will be discarded.
```

PR #870, opened 2020-05-15
Have frontend proxy persist events when using redis

erikjohnston:
Requires: https://github.com/matrix-org/synapse/pull/7517

I chose to use frontend proxy rather than the event creator to ensure we test the HTTP replication routing. (We could use a new worker, but that is surprisingly faffy)

### DIFF
--- a/tests/31sync/06state.pl
+++ b/tests/31sync/06state.pl
@@ -774,7 +774,7 @@ test "Current state appears in timeline in private history with many messages be
                  qw( can_sync ) ],
 
    # sending 50 messages can take a while
-   timeout => 20000,
+   timeout => 20,
 
    check => sub {
       my ( $creator, $syncer, $invitee ) = @_;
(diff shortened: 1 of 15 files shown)


erikjohnston:
There have been quite a few changes here...

tests/31sync/06state.pl
```diff
@@ -774,7 +774,7 @@
                  qw( can_sync ) ],
 
    # sending 50 messages can take a while
-   timeout => 20000,
+   timeout => 20,
```
anoadragon453, line 777: Has this made tests go faster or was this just an unnecessarily large number to begin with?

erikjohnston: Ah, I forgot about these. I think there was a confusion for the units here, they're in seconds not milliseconds (a timeout of 6 hours seems excessive).

PR #874, opened 2020-05-20
Fix race in 'Outbound federation can backfill events'

Kegsay:
This test will issue a `/join` for a remote room alias then attempt
to backpaginate in the room. The way the backpagination works
presents a race condition. Previously it:
 - Sent the `/join`, wait for 200 OK.
 - Hit `/sync` and remember `next_batch`
 - Hit `/messages` with `next_batch`.
 - If not enough messages, hit `/messages` again with `end` token from
   `/messages`.

In Dendrite, it's possible that `/sync` has not seen the join membership
event yet, resulting in a `next_batch` that represents the state *before*
the membership event. Backpagination then gets messages prior to the join
event, when the test expects it to include the join event. The fix is to
wait for the `m.room.member` event to come down `/sync` and THEN use the
`next_batch` token.

### DIFF
--- a/tests/50federation/34room-backfill.pl
+++ b/tests/50federation/34room-backfill.pl
@@ -17,6 +17,7 @@ test "Outbound federation can backfill events",
          creator => $creator_id,
          alias   => $room_alias,
       );
+      my $room_id;
 
       # Create some past messages to backfill from
       $room->create_and_insert_event(
@@ -65,7 +66,16 @@ test "Outbound federation can backfill events",
          )->then( sub {
             my ( $body ) = @_;
 
-            my $room_id = $body->{room_id};
+            $room_id = $body->{room_id};
+
+            # wait for it to arrive
+            await_sync_timeline_contains(
+               $user, $room_id,
+               check => sub {
+                  $_[0]->{type} eq "m.room.member"
+               },
+            );
+         })->then( sub {
 
             # It make take some time for the join to propagate, so we need to
             # retry on failure.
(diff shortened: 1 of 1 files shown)


tests/50federation/34room-backfill.pl
```diff
@@ -65,7 +66,16 @@
...
-            my $room_id = $body->{room_id};
+            $room_id = $body->{room_id};
+
+            # wait for it to arrive
+            await_sync_timeline_contains(
+               $user, $room_id,
+               check => sub {
+                  $_[0]->{type} eq "m.room.member"
+               },
+            );
```
richvdh, line 77: isn't this the point of the next bit happening in a loop? the comment even says so?

and if we need to make this change, can we remove the loop? Or at least update the comments?

Kegsay: I'll remove the loop as it shouldn't be needed anymore.

Kegsay: Done.

richvdh: I'm moderately interested to know why the loop was insufficent, but :man_shrugging:

Kegsay: The way this test and `matrix_get_room_messages` handles tokens breaks things. This function:
 - Hits `/sync`.
 - If the `from` parameter was set, uses it to call `/messages`, else uses `next_batch`. 

We pass an empty `from` at first, resulting in `next_batch` being used as the `?from=` param for `/messages`. The test then sets `from` to be `chunk.end` to backpaginate more results.

If the membership event has come down `/sync` then `next_batch` represents a point after the event has been sent, so using that token with `/messages` will return that event again which the test wants. If the membership event has not come down `/sync` yet then `next_batch` represents a point before the membership event, so using that token with `/messages` will not return the membership event, causing the test to fail.

The test assumes that `/sync` will have the membership event, but that's not always true until now.

PR #880, opened 2020-06-02
Remove room alias check for createRoom response

Jbbouille:
It seems in the spec that room creation response does not have a room_alias. 

Check this issue for more information: https://git.koesters.xyz/timo/conduit/issues/23.

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -23,7 +23,6 @@ test "POST /createRoom makes a public room",
 
          assert_json_keys( $body, qw( room_id room_alias ));
          assert_json_nonempty_string( $body->{room_id} );
-         assert_json_nonempty_string( $body->{room_alias} );
 
          Future->done(1);
       });


erikjohnston:
Hmm, actually, have you tested this change against a HS that doesn't return `room_alias` it looks like other parts of the code base still assume that it is returned so removing this check will still leave a lot of tests broken potentially?

Jbbouille:
I didn't test it against a HS that doesn't return `room_alias`. Actually I don't know what a `HS` is. How can I test this in order to safely merge it ?

erikjohnston:
Sorry, by `HS` I mean a home server like conduit.

clokep:
@Jbbouille Just wanted to check in and see if you've had a chance to look into this!

richvdh:
I'm going to merge this, since it's valid in itself.  I'm minded to agree with @erikjohnston that other parts of sytest are going to incorrectly have problems with server implementations which do not return a `room_alias` here, but they can be addressed separately.

PR #888, opened 2020-06-12
Switch to YAML::XS for YAML serialisation

richvdh:
it's better at understanding the difference between `1`, `true`, and
`"true"`. This leads to a few fixes around the codebase where it turned out we
were getting away with some sloppiness thanks to `YAML`'s idiosyncracies.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 .git/
+server-0/
+server-1/
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -18,7 +18,7 @@ use File::Path qw( remove_tree );
 use List::Util qw( any );
 use POSIX qw( strftime WIFEXITED WEXITSTATUS );
 
-use YAML ();
+use JSON;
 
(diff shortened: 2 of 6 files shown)


.dockerignore
```diff
@@ -1 +1,3 @@
 .git/
+server-0/
+server-1/
```
richvdh, line 3: this is kinda unrelated but was something I came across when trying to rebuild the docker images to include `YAML::XS`.

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -18,7 +18,7 @@ use File::Path qw( remove_tree );
 use List::Util qw( any );
 use POSIX qw( strftime WIFEXITED WEXITSTATUS );
 
-use YAML ();
```
richvdh, line 21: this was unused.

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -410,9 +410,8 @@ sub generate_listeners
          type         => "http",
          port         => $unsecure_port,
          bind_address => $bind_host,
-         tls          => 0,
```
richvdh, line 413: `tls` and `compress` are both disabled by default.

PR #896, opened 2020-06-17
Make typing tests use /sync not /events

Kegsay:
This adds the helper function `await_sync_ephemeral_contains` which behaves the same as `await_sync_timeline_contains`. Also, make typing send true/false not 1/0 which Dendrite dislikes (and so does the spec!).

https://github.com/matrix-org/sytest/pull/896/files?diff=split&w=1 gives a nicer diff.

### DIFF
--- a/scripts/dendrite_sytest.sh
+++ b/scripts/dendrite_sytest.sh
@@ -45,7 +45,6 @@ TEST_STATUS=0
 mkdir -p /logs
 ./run-tests.pl -I Dendrite::Monolith -d $GOBIN -W /src/sytest-whitelist -O tap --all \
     --work-directory="/work" \
-    tests/30rooms/20typing.pl \
     "$@" > /logs/results.tap || TEST_STATUS=$?
 
 if [ $TEST_STATUS -ne 0 ]; then
--- a/scripts/dendrite_sytest.sh
+++ b/scripts/dendrite_sytest.sh
@@ -45,6 +45,7 @@ TEST_STATUS=0
 mkdir -p /logs
 ./run-tests.pl -I Dendrite::Monolith -d $GOBIN -W /src/sytest-whitelist -O tap --all \
     --work-directory="/work" \
+    tests/30rooms/20typing.pl \
     "$@" > /logs/results.tap || TEST_STATUS=$?
 
 if [ $TEST_STATUS -ne 0 ]; then
--- a/tests/10apidoc/09synced.pl
+++ b/tests/10apidoc/09synced.pl
@@ -6,6 +6,7 @@ push our @EXPORT, qw(
    sync_timeline_contains
    await_sync
    await_sync_timeline_contains
+   await_sync_ephemeral_contains
    await_sync_timeline_or_state_contains
    await_sync_presence_contains
 );
@@ -92,6 +93,13 @@ sub sync_timeline_contains
    my ( $sync_body, $room_id, $check ) = @_;
 
    sync_room_contains( $sync_body, $room_id, "timeline", $check );
+}
+
+sub sync_ephemeral_contains
+{
+   my ( $sync_body, $room_id, $check ) = @_;
+
+   sync_room_contains( $sync_body, $room_id, "ephemeral", $check );
 }
 
 sub sync_presence_contains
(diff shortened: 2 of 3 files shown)


Kegsay:
For some reason Synapse isn't happy in worker mode...

tests/10apidoc/09synced.pl
```diff
@@ -189,6 +197,21 @@ sub await_sync_timeline_contains {
    )
 }
 
+sub await_sync_ephemeral_contains {
```
richvdh, line 200: can has perldoc please?

tests/30rooms/20typing.pl
```diff
@@ -109,27 +106,22 @@ sub matrix_typing
...
 
-               scalar @$users and
-                  die "Expected 0 members to be typing";
+                  assert_json_keys( $event, qw( type content ));
+                  assert_json_keys( my $content = $event->{content}, qw( user_ids ));
+                  assert_json_list( my $users = $content->{user_ids} );
 
-               return 1;
-            })
+                  return scalar @$users == 0;
```
richvdh, line 122: if this doesn't match, shouldn't we raise an error rather than continue waiting for another typing event?

Kegsay: It originally raised an error which caused Synapse in worker mode to fail, presumably because in worker mode it might redundantly send the typing:true event or something. Waiting a bit longer for the typing:false event makes this test pass in worker mode.

richvdh: It's surprising that this used to work reliably, but if what you say is correct...

The problem is that tests that just time out rather than failing when the code isn't doing the right thing are something of an antipattern (mostly: you can't tell if the server was slow or there is an actual bug). We already have too many tests in sytest that do this, and I am keen that we don't add any more.

It's hard to completely eliminate this, but to make it better, I suggest you refactor it so that you check each returned typing event, and more than one (?) event with the wrong number of users.

Most importantly though, you need to `log_if_fail` for each "rejected" typing event, saying why it is rejected. One way to do this is to wrap the whole thing in a `retry_until_success` loop with an `on_fail` handler (see https://github.com/matrix-org/sytest/blob/master/tests/30rooms/13guestaccess.pl#L414-L418 for example), but there are plenty of other solutions.

Kegsay: I've made the following changes:
 - We log each typing event received. We also log why the event is being ignored if it doesn't meet the criteria (number of users).
 - If we encounter more than 1 typing event with the wrong number of users, we bail out the test.

Produces output like: (ignore the fact that there are no users typing, I had to invert the checks to get this output)
```
not ok 3 (expected fail) Typing can be explicitly stopped # TODO expected fail
# Started: 2020-06-23 11:10:43.737
# Ended: 2020-06-23 11:10:43.756
# seen too many typing events with typing users at tests/30rooms/20typing.pl line 130.
# 0.017327: Received typing event: 
# {
#   content => { user_ids => [] },
#   event_id => "",
#   origin_server_ts => JSON::number(0),
#   sender => "",
#   type => "m.typing",
# }
# 0.017624: rejecting event because want zero users typing, but there are some
```

tests/30rooms/20typing.pl
```diff
@@ -109,17 +109,28 @@ sub matrix_typing
...
-                  return scalar @$users == 0;
+                  my $zero_users = scalar @$users == 0;
+                  if ( !$zero_users ) {
+                     log_if_fail "rejecting event because want zero users typing, but there are some";
+                     if ( $num_typing_events > 1) {
+                        # this is the second time we have seen a typing event with >0 typing users, bail out
+                        die "seen too many typing events with typing users";
+                     }
+                  }
+                  return $zero_users;
```
richvdh, line 133: suggest an early return here:

```suggestion
                  return 1 if scalar @$users == 0;

                  log_if_fail "rejecting event because want zero users typing, but there are some";

                  if ( $num_typing_events > 1 ) {
                     # this is the second time we have seen a typing event with >0 typing users, bail out
                     die "seen too many typing events with typing users";
                  }
                  return 0;
```

scripts/dendrite_sytest.sh
```diff
@@ -45,6 +45,7 @@ TEST_STATUS=0
 mkdir -p /logs
 ./run-tests.pl -I Dendrite::Monolith -d $GOBIN -W /src/sytest-whitelist -O tap --all \
     --work-directory="/work" \
+    tests/30rooms/20typing.pl \
```
richvdh, line 48: this looks like it shouldn't be here. (why not just pass `tests/30rooms/20typing.pl` on the `docker run` commandline?)

Kegsay: :D didn't know we passed all that through! Awesome thanks!

PR #898, opened 2020-06-23
Make invite tests use /sync not room /initialSync

Kegsay:
Dendrite doesn't implement deprecated endpoints which makes all these tests fail annoyingly, so make them use /sync.


Kegsay:
Failing test:
> New room members see first user's profile information in global initialSync

seems unrelated.

PR #909, opened 2020-07-08
Test that the client sees a leave event after invite rejection

richvdh:
Regression test for https://github.com/matrix-org/synapse/issues/2181.


tests/50federation/35room-invite.pl
```diff
@@ -250,11 +250,12 @@ sub do_v1_invite_request
    my ( $room, $first_home_server, $outbound_client, $invitation ) = @_;
 
    my $room_id = $room->room_id;
+   my $event_id = $room->id_for_event( $invitation );
```
richvdh, line 253: this is a slightly unrelated change, but the warning messages were confusing me.

PR #911, opened 2020-07-10
Correctly return the full auth chain in test

erikjohnston:
Synapse now needs the auth chain to be correct: https://github.com/matrix-org/synapse/pull/7817

### DIFF
--- a/tests/50federation/36state.pl
+++ b/tests/50federation/36state.pl
@@ -469,15 +469,16 @@ test "Outbound federation requests missing prev_events and then asks for /state_
                   $state{$k} = $event;
                }
 
+               my @auth_chain = $inbound_server->{datastore}->get_auth_chain_events(
+                  map { $_->{event_id} } values( %state )
+               );
+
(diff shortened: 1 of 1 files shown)


tests/50federation/36state.pl
```diff
@@ -469,15 +469,16 @@ sub get_state_ids_from_server {
...
                my $resp = {
                   pdu_ids => [
                      map { $_->{event_id} } values( %state ),
                   ],
                   auth_chain_ids => [
-                     # XXX we're supposed to return the whole auth chain here,
-                     # not just y's auth_events. It doesn't matter too much
-                     # here though.
-                     map { $_->[0] } @{ $missing_event_y->{auth_events} },
+                     map { $_->{event_id} } @auth_chain,
```
richvdh, line 481: ftr, this won't work in room v3 or later, though it looks like the whole test needs an update for that anyway. The right way to write this is `$room->id_for_event( $_ )`.

PR #912, opened 2020-07-10
Unflake "Outbound federation will ignore a missing event with bad JSON for room version 6"

### LINKED ISSUE #885: `Outbound federation will ignore a missing event with bad JSON for room version 6` is flakey
https://buildkite.com/matrix-dot-org/synapse/builds/9476#5143875e-134e-465c-b7d0-8e2d5884ae1e

clokep:
Fixes #885, I think see https://github.com/matrix-org/sytest/issues/885#issuecomment-647506143 and the subsequent comment for what's happening here, but this uses option 3 to solve it: allow the event we're expecting or previous events as the "earliest event".

### DIFF
--- a/tests/50federation/33room-get-missing-events.pl
+++ b/tests/50federation/33room-get-missing-events.pl
@@ -452,8 +452,8 @@ test "Outbound federation will ignore a missing event with bad JSON for room ver
 
       log_if_fail "Missing event", $missing_event;
 
-      # Generate another one and do send it so it will refer to the
-      # previous in its prev_events field
+      # Generate another event which will be sent. It will refer to the missing
+      # event in its prev_events field.
       my $sent_event = $room->create_and_insert_event(
          type => "m.room.message",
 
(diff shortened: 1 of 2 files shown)


tests/50federation/33room-get-missing-events.pl
```diff
@@ -484,8 +484,12 @@ sub sytest_user_and_room_fixture {
...
             @$earliest == 1 or
                die "Expected a single 'earliest_event' ID";
-            assert_eq( $earliest->[0], $room->id_for_event( $latest_event ),
-               'earliest_events[0]' );
+            # It is expected that the earliest event is the m.room.member event,
+            # but it is possible that the caches have not yet been invalidated
+            # so also allow any of that event's previous events.
+            my @expected = @{$latest_event->{prev_events}};
+            push( @expected, $room->id_for_event( $latest_event ) );
+            assert_in_list( $earliest->[0], @expected, 'earliest_events[0]' );
```
richvdh, line 492: I think this is going to use `'earliest_events[0]'` as one of the things that `$earliest->[0]` is allowed to match, which isn't quite what you want.

I'd just do: 

```perl
assert_ok( any { $earliest->[0] eq $_ } @expected, "'earliest_events' did not match" );
```

but if you prefer a utility function that's ok; just call it properly :)

clokep: Ah, that's a bit easier. :)

clokep:
I had to amend this commit cause...I totally messed it up. Just added the one on top of what was there though. Hope this looks better.

PR #920, opened 2020-07-24
Use /sync not /events for SYN-205 test

Kegsay:


Kegsay:
I've done enough of these to know how the `await_` form works, I just forgot to have a `return unless` or equiv.

richvdh:
sorry, I didn't mean to talk down to you; just trying to be helpful. The intended of the `check` param is that it just checks if it's the right sync or not, and the assertions then happen on the result of `await_sync`; doing the assertions inside the check was what made me think you weren't familiar with it.

I'll take another look.

tests/90jira/SYN-205.pl
```diff
@@ -11,14 +11,25 @@
...
-            my ( $event ) = @_;
-            return unless $event->{type} eq "m.room.member" and
-                          $event->{room_id} eq $room_id and
-                          $event->{state_key} eq $invitee->user_id and
-                          $event->{content}{membership} eq "invite";
+         await_sync($invitee, check => sub {
+            my ( $sync_body ) = @_;
+            log_if_fail $sync_body;
+            my $room =  $sync_body->{rooms}{invite}{$room_id};
+            return unless $room;
```
richvdh, line 18: it'd be nice to make the falsey return value explicit, rather than relying on undef to be falsey.

```suggestion
            return 0 unless $room;
```

tests/90jira/SYN-205.pl
```diff
@@ -11,14 +11,25 @@
...
 
-            return 1;
-         })->SyTest::pass_on_done( "Invitee received invite event" )
+            assert_json_keys( $invite, qw( sender content state_key type ));
+            $invite->{content}{membership} eq "invite"
+               or die "Expected an invite event";
+            $invite->{sender} eq $user->user_id
+               or die "Expected the invite to be from user A";
+
+            Future->done(1);
```
richvdh, line 32: `check` just needs a truthy value, not a future. Also it's nice to make the `return` explicit.

```suggestion
            return 1;
```

tests/90jira/SYN-205.pl
```diff
@@ -11,14 +11,25 @@
...
+                  and $_->{state_key} eq $invitee->user_id
+            } @{ $room->{invite_state}{events} };
 
-            return 1;
-         })->SyTest::pass_on_done( "Invitee received invite event" )
+            assert_json_keys( $invite, qw( sender content state_key type ));
+            $invite->{content}{membership} eq "invite"
+               or die "Expected an invite event";
+            $invite->{sender} eq $user->user_id
+               or die "Expected the invite to be from user A";
```
richvdh, line 30: why not use `assert_eq` here?

tests/90jira/SYN-205.pl
```diff
@@ -11,14 +11,25 @@
...
-            return unless $event->{type} eq "m.room.member" and
-                          $event->{room_id} eq $room_id and
-                          $event->{state_key} eq $invitee->user_id and
-                          $event->{content}{membership} eq "invite";
+         await_sync($invitee, check => sub {
+            my ( $sync_body ) = @_;
+            log_if_fail $sync_body;
+            my $room =  $sync_body->{rooms}{invite}{$room_id};
+            return unless $room;
+            assert_json_keys( $room, qw( invite_state ) );
```
richvdh, line 19: I'd *rather* these assertions happened outside the `check` sub, but I won't insist on it. Still, a bit of whitespace to separate the `check` part from the `assert` part would be nice:

```suggestion

            assert_json_keys( $room, qw( invite_state ) );
```

tests/90jira/SYN-205.pl
Kegsay: Mainly because the code before had it in this format. I don't really care which we end up using.

richvdh: which code?

Kegsay: I yanked `invite_state` code from invite tests when mapping this over to use `/sync`.

richvdh: I'm still unclear on where you've cargo-culted this from, but it doesn't really matter. The situation is:

 * yes, there exists code which uses the `$foo eq $bar or die` pattern
 * that code either pre-dates the existence of `assert_eq`, or has been cargo-culted from other code that does
 * I'd prefer it if we didn't proliferate the cargo-culting, but I'm not going to insist on it.

richvdh:
@Kegsay: please could you squash-merge things like this in future?

Kegsay:
Sure thing

PR #922, opened 2020-07-30
Wait for pushers to become active before proceeding

richvdh:
Hopefully, this will make the pusher tests less flaky when running in a
redis worker setup. Wait for the pusher to respond before we proceed.

### DIFF
--- a/lib/SyTest/Federation/Server.pm
+++ b/lib/SyTest/Federation/Server.pm
@@ -490,8 +490,6 @@ sub on_request_federation_v1_send
       }
 
       next if $self->on_event( $event );
-
-      warn "TODO: Unhandled incoming event of type '$event->{type}'";
    }
 
    Future->done( json => {} );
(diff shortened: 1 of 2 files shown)


lib/SyTest/Federation/Server.pm
```diff
@@ -490,8 +490,6 @@ sub on_request_federation_v1_send
       }
 
       next if $self->on_event( $event );
-
-      warn "TODO: Unhandled incoming event of type '$event->{type}'";
```
erikjohnston, line 494: Just checking this is intentional

richvdh: ah, yes it is. The updated code spits out more of these and I don't think we really need the warning.

erikjohnston: Fair, I've wondered about removing this myself!

PR #928, opened 2020-08-05
Add redis support to synapse docker image

richvdh:
... because I kept forgetting how to do the networking foo to get redis
running, and because it seems to make sense now that TCP replication is
deprecated.


richvdh:
(follow-up task is to change the buildkite pipeline to use it rather than spinning up a separate redis)

clokep:
@richvdh Looks like this has a test failure, looks unrelated, but not sure.

PR #929, opened 2020-08-05
Use ProcessManager in Synapse.pm

richvdh:
There is a bunch of magical code in `HomeServer::Synapse` for starting and monitoring processes. Replace this with the generic implementation in `ProcessManager`.

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -3,7 +3,7 @@ package SyTest::Homeserver;
 use strict;
 use warnings;
 use 5.010;
-use base qw( IO::Async::Notifier );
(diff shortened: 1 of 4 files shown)


lib/SyTest/Homeserver.pm
```diff
@@ -173,25 +173,6 @@ sub configure
...
-
-Kill any processes started for the homeserver, and wait for them to exit.
-
-Returns a Future when the proceses have exited.
-
-It is expected that this will be overridden by subclasses.
-
-=cut
-
-sub kill_and_await_finish
```
richvdh, line 190: this is now provided by SyTest::Homeserver::ProcessManager, so doesn't need a dummy implementation here.

richvdh:
yeah, I think so.

PR #932, opened 2020-08-06
Deflake 'Server correctly resyncs when server leaves and rejoins a room'

### LINKED ISSUE #918: "Server correctly resyncs when server leaves and rejoins a room" is flaky on redis
see https://buildkite.com/matrix-dot-org/synapse/builds/10133#97f168a4-b442-4d80-88bd-db3615aa36d1

richvdh:
Waiting for the leave event to turn up should make this less flaky.

fixes #918

### DIFF
--- a/tests/50federation/40devicelists.pl
+++ b/tests/50federation/40devicelists.pl
@@ -307,7 +307,7 @@ test "Server correctly resyncs when server leaves and rejoins a room",
       )->then( sub {
          my ( $first, $content ) = @_;
 
-         log_if_fail "first query response", $content;
+         log_if_fail "initial device query response", $content;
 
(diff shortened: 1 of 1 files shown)


tests/50federation/40devicelists.pl
```diff
@@ -317,19 +317,41 @@
...
+                  log_if_fail "received event over federation", $ev;
+                  return $ev->{state_key} eq $user->user_id &&
+                     $ev->{content}{membership} eq 'leave';
+               }
+            ),
+         );
       })->then( sub {
+         log_if_fail "left room; now rejoining";
+         my $iter = 0;
          retry_until_success {
```
richvdh, line 339: :thinking: I wonder if this loop is still required

PR #935, opened 2020-08-11
Fix /send server ACL test to match spec

neilalexander:
The test was looking for `errcode`, which Synapse returns but is not specced.

The test now looks for `error`, which is specced!

### DIFF
--- a/tests/50federation/50server-acl-endpoints.pl
+++ b/tests/50federation/50server-acl-endpoints.pl
@@ -53,9 +53,7 @@ sub can_send_event {
       my $pdu_result = $body->{pdus}->{$event_id};
 
       if( $params{expect_ban} ) {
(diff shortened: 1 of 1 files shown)


erikjohnston:
Although the annoying thing here is that this won't detect if we threw an "internal server error" while handling the ACL. Though I guess the net result would be the same...

neilalexander:
An internal server error should return a 500 and presumably no txn response body? (At least that's what dendrite does as I just had to fix a panic related to this and the test failed as expected.)

erikjohnston:
> An internal server error should return a 500 and presumably no txn response body? (At least that's what dendrite does as I just had to fix a panic related to this and the test failed as expected.)

Not if there was an internal error when processing an event. Instead it should return the error in the PDU results dict.

PR #941, opened 2020-08-26
Add plugin support.

valkum:
This PR adds support for plugins in sytest. The bootstrap script will download given plugins and sytest will use them.

You can now pass the env var $PLUGIN with the format `PLUGINS="<owner>/<repo>[:<owner>/<repo>]"`
bootstrap.sh (when downloading sytest) will load `<owner>/<repo>` from github to `/sytest/plugins/<owner>/<repo>`
and will look if a matching runner script is found in one of the plugins.

Example:
```sh
$ PLUGIN="valkum/conduit_sytest" bootstrap.sh "conduit"
$ # Downloads https://github.com/valkum/conduit_sytest to /sytest/plugins/valkum/sytest 
$ # and executes /sytest/plugins/valkum/sytest/scrypts/conduit_sytest.sh
```
Sytest will look for Output and ServerImplementation plugins in the plugin dir.
Plugin dir is `/sytest/plugin` by default but can be overwritten by the env var `$SYTEST_PLUGINS`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,12 +146,17 @@ synapse) and place it within the homeserver configuration directory
 
 .. _dictConfig: https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig
 
+Plugins
+~~~~~~~
+
 Sytest supports plugins. Plugins follow the same project structure as sytest and can be placed
 in the ``plugins`` directory. They should contain the ``lib/SyTest/HomeserverFactory`` and
 ``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories, similar to the root of the sytest repository.
 The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS`` environment variable.
 
 Currently only ``Homeserver`` and ``Output`` implementations are supported in plugins.
+
+See https://github.com/valkum/sytest_conduit for an example of a plugin.
 
 Developing
 ----------
--- a/README.rst
+++ b/README.rst
@@ -147,10 +147,11 @@ synapse) and place it within the homeserver configuration directory
 .. _dictConfig: https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig
 
 Sytest supports plugins. Plugins follow the same project structure as sytest and can be placed
-in the `plugins` directory.
-To override the path of the plugins directory set the env var `SYTEST_PLUGINS` to another directory.
+in the ``plugins`` directory. They should contain the ``lib/SyTest/HomeserverFactory`` and
+``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories. Similar to the sytest repository.
+The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS``environment variable.
 
-Currently only Homeserver and Output Implementations are supported in plugins.
+Currently only ``Homeserver`` and ``Output`` implementations are supported in plugins.
 
 Developing
 ----------
--- a/README.rst
+++ b/README.rst
@@ -146,6 +146,12 @@ synapse) and place it within the homeserver configuration directory
 
 .. _dictConfig: https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig
 
+Sytest supports plugins. Plugins follow the same project structure as sytest and can be placed
+in the `plugins` directory.
+To override the path of the plugins directory set the env var `SYTEST_PLUGINS` to another directory.
+
+Currently only Homeserver and Output Implementations are supported in plugins.
+
 Developing
 ----------
 
--- a/README.rst
+++ b/README.rst
@@ -148,8 +148,8 @@ synapse) and place it within the homeserver configuration directory
 
 Sytest supports plugins. Plugins follow the same project structure as sytest and can be placed
 in the ``plugins`` directory. They should contain the ``lib/SyTest/HomeserverFactory`` and
-``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories. Similar to the sytest repository.
-The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS``environment variable.
+``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories, similar to the root of the sytest repository.
+The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS`` environment variable.
 
 Currently only ``Homeserver`` and ``Output`` implementations are supported in plugins.
 
--- a/docker/bootstrap.sh
+++ b/docker/bootstrap.sh
@@ -29,10 +29,10 @@ else
     fi
 
     # Try and fetch the branch
-    wget -q https://github.com/valkum/sytest/archive/$branch_name.tar.gz -O sytest.tar.gz || {
+    wget -q https://github.com/matrix-org/sytest/archive/$branch_name.tar.gz -O sytest.tar.gz || {
         # Probably a 404, fall back to develop
         echo "Using develop instead..."
-        wget -q https://github.com/valkum/sytest/archive/develop.tar.gz -O sytest.tar.gz
+        wget -q https://github.com/matrix-org/sytest/archive/develop.tar.gz -O sytest.tar.gz
     }
 
     mkdir -p /sytest
@@ -61,10 +61,12 @@ if [ -x "/sytest/scripts/${SYTEST_TARGET}_sytest.sh" ]; then
 elif [ -x "/sytest/docker/${SYTEST_TARGET}_sytest.sh" ]; then
     # old branches of sytest used to put the sytest running script in the "/docker" directory
     exec "/sytest/docker/${SYTEST_TARGET}_sytest.sh" "$@"
+
 else
     PLUGIN_RUNNER=$(find /sytest/plugins/ -type f -exec test -x {} \; -name "${SYTEST_TARGET}_sytest.sh" -print)
     if [ -n PLUGIN_RUNNER ]; then
         exec ${PLUGIN_RUNNER} "$@"
+        
     else
         echo "sytest runner script for ${SYTEST_TARGET} not found" >&2
         exit 1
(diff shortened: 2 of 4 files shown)


clokep:
@valkum Were you hoping for feedback on this or does it need more work (noting that this is a draft PR)?

valkum:
> @valkum Were you hoping for feedback on this or does it need more work (noting that this is a draft PR)?

The SyTest part is tested, but I could not test the bootstrap.sh part in a container yet because of failing builds for the base image. Mentioned that in the homerserver-dev matrix room.
Wanted to get some feedback before I invest more time into testing.

run-tests.pl
```diff
@@ -34,17 +35,22 @@
       : undef;
 });
 
+my $plugin_dir = $ENV{"SYTEST_PLUGINS"} || "plugins"; # Read plugin dir from env var SYTEST_PLUGINS or fallback to plugins
+my @plugins = grep { -d } glob(catdir($plugin_dir, "*"));
```
valkum, line 39: Needs some rework, because bootstrap.sh will download into `plugins/<owner>/<plugin>` 
glob probably won't list all dirs 2 layers deep? I do not write perl fluently.

valkum:
I finished the missing work and tested it locally. Ready for review

anoadragon453:
Could you give us an understanding of what the usecase for plugins in Sytest would be? I took a look at https://github.com/valkum/sytest_conduit, and it seems add files to allow SyTest to spin up and run against a Conduit instance.

Could you not instead simply PR those files to this repo to add support instead?

valkum:
Submitting a PR for Conduit would be another way. But depending on how many Homeserver implementations will get created over time I thought this would be a better way.
With plugin support different projects are able to add their own Homeserver implementations and Output formatters.

If you are more interested in simply adding Conduit support to SyTest its fine. I would submit a PR with our Homeserver implementation then.

run-tests.pl
```diff
@@ -34,17 +35,22 @@
       : undef;
 });
 
+my $plugin_dir = $ENV{"SYTEST_PLUGINS"} || "plugins"; # Read plugin dir from env var SYTEST_PLUGINS or fallback to plugins
+my @plugins = grep { -d } glob(catdir($plugin_dir, "*", "*")); # Read all plugins/<author>/<plugin>
```
richvdh, line 39: I'm curious about this structure: why so many directories?

run-tests.pl
```diff
@@ -34,17 +35,22 @@
       : undef;
 });
 
+my $plugin_dir = $ENV{"SYTEST_PLUGINS"} || "plugins"; # Read plugin dir from env var SYTEST_PLUGINS or fallback to plugins
```
richvdh, line 38: please could you document this environment variable in https://github.com/matrix-org/sytest/#running ?

docker/bootstrap.sh
```diff
@@ -51,6 +63,12 @@ elif [ -x "/sytest/docker/${SYTEST_TARGET}_sytest.sh" ]; then
     exec "/sytest/docker/${SYTEST_TARGET}_sytest.sh" "$@"
 
 else
-    echo "sytest runner script for ${SYTEST_TARGET} not found" >&2
-    exit 1
+    PLUGIN_RUNNER=$(find /sytest/plugins/ -type f -exec test -x {} \; -name "${SYTEST_TARGET}_sytest.sh" -print)
+    if [ -n PLUGIN_RUNNER ]; then
+        exec ${PLUGIN_RUNNER} "$@"
+        
+    else
```
richvdh, line 70:
```suggestion
        exec "${PLUGIN_RUNNER}" "$@"
    else
```

docker/bootstrap.sh
```diff
@@ -51,6 +63,12 @@ elif [ -x "/sytest/docker/${SYTEST_TARGET}_sytest.sh" ]; then
     exec "/sytest/docker/${SYTEST_TARGET}_sytest.sh" "$@"
 
 else
-    echo "sytest runner script for ${SYTEST_TARGET} not found" >&2
-    exit 1
+    PLUGIN_RUNNER=$(find /sytest/plugins/ -type f -exec test -x {} \; -name "${SYTEST_TARGET}_sytest.sh" -print)
```
richvdh, line 66: no need to shell out to `test` here

```suggestion
    PLUGIN_RUNNER=$(find /sytest/plugins/ -type f -executable -name "${SYTEST_TARGET}_sytest.sh" -print)
```

docker/bootstrap.sh
```diff
@@ -37,6 +37,18 @@ else
 
     mkdir -p /sytest
     tar -C /sytest --strip-components=1 -xf sytest.tar.gz
+
+    if [ -n "$PLUGIN" ]; then
+        mkdir /sytest/plugins
+        echo "--- Downloading plugins for sytest"
+        IFS=':'; for plugin in $PLUGIN; do
```
richvdh, line 44: this should probably be called `PLUGINS`, if it's going to support more than one plugin?

docker/bootstrap.sh
```diff
@@ -37,6 +37,18 @@ else
...
+            wget -q https://github.com/$plugin/archive/master.tar.gz -O plugin.tar.gz || {
```
richvdh, line 45: I'm not entirely convinced that tying this to github is the right thing to do. Can we make it more generic?

docker/bootstrap.sh
```diff
@@ -37,6 +37,18 @@ else
...
+                echo "Failed to download plugin: $plugin"
```
richvdh, line 46: we write this (to stdout rather than stderr) and then carry on with mkdir and tar anyway? seems like we should exit instead? or just let `set -e` handle it?

richvdh (changes requested):
I think this is generally a nice approach. I've left a few comments on details of it.

Generally: as part of the work, please make sure the documentation is updated to match. In particular, I think the top-level README and https://github.com/matrix-org/sytest/blob/develop/docker/README.md will need an update to document the new features, and the expected layout of the plugins directory/archives.

run-tests.pl
valkum: So you could use github as a the plugin repository. We could omit the <author> part and keep the sturcture more flat. That would need some changes to the downloading of plugins.

docker/bootstrap.sh
valkum: -executable is not very portable to my understanding. e.g. FreeBSD find does not support it.

docker/bootstrap.sh
valkum: Yeah. I thought of the PATH var while writing it not sure why that is not called PATHS, but PLUGINS makes more sense. Will change that.

docker/bootstrap.sh
valkum: It becomes difficult to select a seperator to pass multple plugins if we want to be able to pass whole URLs.
Something like # or : (with stripped https://) comes into my mind. But that would need good documentation.
Would are your thought on this?

docker/bootstrap.sh
valkum: Thanks for finding this. Yeah we should exit there with sending the msg to stderr.

docker/bootstrap.sh
ara4n: macOS doesn't have it either.

valkum:
I answered some of your feedback. I will add documentation when we decided on the exact format of dirs and variables.

docker/bootstrap.sh
richvdh: fair enough. Maybe we can just omit it? if it exists but isn't executable, that suggests an error rather than a file we should ignore?

run-tests.pl
richvdh: I think I'd prefer that.

docker/bootstrap.sh
richvdh: for the separator: can we just use space? PLUGINS="https://github.com/valkum/conduit_sytest/archive/master.tar.gz https://gitlab.com/project/plugin/-/archive/master/plugin-master.tar.gz"

for the directory: can we just make up a `mktemp`-style name? it doesn't need to be related to the URL of the plugin, does it?

richvdh:
@valkum are you still interested in working on this?

valkum:
Yeah, wrote kegsay in the complement room that I am currently at crunchtime and can pick up work on this (and the one in complement) after the 26th.

richvdh (changes requested):
looks good, though the docs could do with a few tweaks.

thanks!

README.rst
```diff
@@ -146,6 +146,13 @@ synapse) and place it within the homeserver configuration directory
 
 .. _dictConfig: https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig
 
+Sytest supports plugins. Plugins follow the same project structure as sytest and can be placed
+in the ``plugins`` directory. They should contain the ``lib/SyTest/HomeserverFactory`` and
+``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories. Similar to the sytest repository.
```
richvdh, line 151:
```suggestion
``lib/SyTest/Homeserver``, or ``lib/SyTest/Output`` directories, similar to the root of the sytest repository.
```

README.rst
```diff
@@ -146,6 +146,13 @@ synapse) and place it within the homeserver configuration directory
...
+The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS``environment variable.
```
richvdh, line 152:
```suggestion
The path of the plugins directory can be overridden via the ``SYTEST_PLUGINS`` environment variable.
```

richvdh (approved):
lgtm! thanks very much!

richvdh:
right, one remaining thing to do: please could you read https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off and add a comment with a "signed-off-by" line below, to confirm you agree to the statement there?

richvdh:
[CI failed due to matrix-org/synapse#8785, which is fixed in #984)

valkum:
signed-off-by: Rudi Floren <rudi.floren@gmail.com>

PR #943, opened 2020-08-28
Wait for synapse to start by using sd_notify.

erikjohnston:
Currently SyTest waits for synapse process to fully start by repeatedly
checking if it can connect to the HTTP socket. This won't work for
starting worker processest that do not listen on any HTTP socket.

Instead, we use the `sd_notify` mechanism. This works by SyTest binding
to a unix socket, passing the socket address to Synapse via env, and
waiting to receive a `READY=1` notification on the socket.

### DIFF
--- a/lib/SyTest/Homeserver/ProcessManager.pm
+++ b/lib/SyTest/Homeserver/ProcessManager.pm
@@ -283,15 +283,7 @@ sub _start_process_and_await_notify
       $finished_future->without_cancel()->then_fail(
          "Process died without becoming connectable",
       ),
-   )->else_with_f( sub {
-      my ( $f ) = @_;
-
-      # We need to manually kill child procs here as we don't seem to have
-      # registered the on finish handler yet.
-      $self->kill_and_await_finish()->then( sub {
-         $f
-      })
-   } );
+   );
    return $fut;
 }
 
@@ -355,11 +347,7 @@ sub _await_ready_notification
       socktype => "dgram",
       path     => $path,
    } )->then( sub {
-      # We add a timeout so that we don't wait for ever if process wedges.
-      Future->wait_any(
-         $poke_fut,
-         $loop->timeout_future( after => 15 )
-      )
+      $poke_fut
    })
 }
 
--- a/lib/SyTest/Homeserver/ProcessManager.pm
+++ b/lib/SyTest/Homeserver/ProcessManager.pm
@@ -267,12 +267,12 @@ sub _start_process_and_await_notify
    my %setup_map = @$setup;  # Copy to a hash so we can pull out the env entry.
    my $env = $setup_map{env} // {};
    if ( not defined $setup_map{env} ) {
-      # There was no env entry, so we me need to add it to the setup array.
+      # There was no env entry, so we need to add it to the setup array.
       push @$setup, env => $env;
    }
 
    # We need to set this up *before* we start the process as we need to bind the
-   # socket before startin the process.
+   # socket before starting the process.
    my $await_fut = $self->_await_ready_notification( $env );
 
    my $proc = $self -> _start_process( %params );
--- a/lib/SyTest/Homeserver/ProcessManager.pm
+++ b/lib/SyTest/Homeserver/ProcessManager.pm
@@ -315,7 +315,9 @@ sub _await_ready_notification
 
    # We replace null byte with '@' to allow us to pass it in via env. (This is
    # as per the sd_notify spec).
-   $env->{"NOTIFY_SOCKET"} = $path =~ s/\0/\@/rg;
+   my $path_env = $path;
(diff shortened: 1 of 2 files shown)


erikjohnston:
Oh, this does't work with pydron of course. Whoops

erikjohnston:
Ok, so I've made nothing use it yet, and instead will use it when I rejig things to remove `pydron.py`.

lib/SyTest/Homeserver/ProcessManager.pm
```diff
@@ -237,5 +238,129 @@ sub _kill_process
...
+   # The env is passed in as part of the `setup` param, which is an ordered map
+   # represented as an array.
+   $params{setup} //= [];
+
+   my $setup = $params{setup};
+
+   my %setup_map = @$setup;  # Copy to a hash so we can pull out the env entry.
+   my $env = $setup_map{env} // {};
+   if ( not defined $setup_map{env} ) {
+      # There was no env entry, so we me need to add it to the setup array.
```
richvdh, line 270:
```suggestion
      # There was no env entry, so we need to add it to the setup array.
```

lib/SyTest/Homeserver/ProcessManager.pm
```diff
@@ -237,5 +238,129 @@ sub _kill_process
...
+      push @$setup, env => $env;
+   }
+
+   # We need to set this up *before* we start the process as we need to bind the
+   # socket before startin the process.
```
richvdh, line 275:
```suggestion
   # socket before starting the process.
```

lib/SyTest/Homeserver/ProcessManager.pm
```diff
@@ -237,5 +238,129 @@ sub _kill_process
...
+      ),
+   )->else_with_f( sub {
+      my ( $f ) = @_;
+
+      # We need to manually kill child procs here as we don't seem to have
+      # registered the on finish handler yet.
+      $self->kill_and_await_finish()->then( sub {
+         $f
+      })
+   } );
```
richvdh, line 294: this feels wrong, for a couple of reasons:
* failure to start one subprocess should not necessarily mean that all other subprocesses get killed (at least at the abstract level of a ProcessManager class)
* it's asymmetric with `_start_process_and_await_connectable`.

what is the "finish handler" of which the comment speaks?

lib/SyTest/Homeserver/ProcessManager.pm
```diff
@@ -237,5 +238,129 @@ sub _kill_process
...
+   my $output = $self->{output};
+
+   # Create a random abstract socket name. Abstract sockets start with a null
+   # byte.
+   my $random_id = join "", map { chr 65 + rand 25 } 1 .. 20;
+   my $path = "\0sytest-$random_id.sock";
+
+   # We replace null byte with '@' to allow us to pass it in via env. (This is
+   # as per the sd_notify spec).
+   $env->{"NOTIFY_SOCKET"} = $path =~ s/\0/\@/rg;
```
richvdh, line 326: cleaner to do this outside `_await_ready_notification` (possibly in another helper function) and just pass in the `$path` to `_await_ready_notification` ?

richvdh, line 326: this looks like it modifies $path itself, which isn't what you want?

erikjohnston: Aahaha. Hahaha. Ha.

:sob: 

The `r` flag there means don't overwrite the variable. Apparently that is the way that you do non-destructive replace or something. Would `my $path_env = $path; $path_env =~ s/\0/\@/g;` work, do you know?

richvdh: oh right. Yes, that should work. As, I think, should `($env->{"NOTIFY_SOCKET"} = $path) =~ s/\0/\@/g;`

lib/SyTest/Homeserver/ProcessManager.pm
erikjohnston: Ah, turns out that there is a timeout several layers up. Its set to 60s so I probably got bored waiting before it fired :/

lib/SyTest/Homeserver/ProcessManager.pm
erikjohnston: I had a look at this, and basically either a) you move it to a function that accepts `$env` which I don't think buys you very much, or b) move this block of code up to the `_start_process_and_await..`  function which just makes it a bit messy.

I kind of like have all the sd notify bits in one place tbh. Maybe it'd be better to rename it `_create_socket_and_await_notification` or something?

richvdh: well, maybe. whatever.

richvdh (approved):
lgtm

Issue #942, opened 2020-08-28
you can no longer set up sytest, and the docker images no longer build


richvdh:
https://github.com/genio/alien-sodium/issues/2

richvdh:
options include:

* pin Alien::Sodium to an old version
* use a different curve25519 impl

PR #944, opened 2020-08-31
test peeking via MSC2753 on dendrite

ara4n:


README.rst
```diff
@@ -90,6 +90,11 @@ some configuration output, rather than an error::
     $ pkg-config --libs libsodium
     -L/usr/local/Cellar/libsodium/1.0.8/lib -lsodium
 
+It's possible that Crypt::NaCL::Sodium has got out of sync with brew's libsodium,
+causing errors about aes128ctr, which was removed in 1.0.8.1 of libsodium.
+In which case you may want to strip out the aes128ctr stuff from Sodium.xs to
+get it to install... :/
```
clokep, line 96: This sounds like #942?

ara4n: it is indeed.

tests/31sync/17peeking.pl
```diff
@@ -0,0 +1,175 @@
...
+      matrix_set_room_history_visibility( $user, $room_id, "world_readable" )->then(sub {
+         matrix_send_room_text_message_synced( $user, $room_id, body => "something to peek");
+      })->then(sub {
+         do_request_json_for( $peeking_user,
+            method => "POST",
+            uri    => "/r0/peek/$room_id",
+            content => {},
+         )
+      })->then(sub {
+         matrix_sync( $peeking_user );
```
Kegsay, line 20: This is racey. There's no guarantee that a `/sync` immediately after a 200 OK from `/peek` will result in the `peek` block existing, because it takes time for `syncapi` to consume Kafka events.

ara4n: hum, true. what's the right soln then? to retry until it turns up?

ara4n: i'm going to assume so; done. i'm a bit confused on why just setting a timeout on the sync isn't good enough though.

Kegsay: > hum, true. what's the right soln then? to retry until it turns up?

Yes, there are helper functions to do this.

README.rst
richvdh: #942 is fixed now.

tests/10apidoc/30room-create.pl
```diff
@@ -16,7 +16,7 @@
          content => {
             visibility      => "public",
             # This is just the localpart
-            room_alias_name => "30room-create",
+            room_alias_name => "30room-create-".time,
```
Kegsay, line 19: Why was this added?

Kegsay (approved):

tests/10apidoc/30room-create.pl
ara4n: it was causing test fails when run on an existing db if that alias already existed.

PR #945, opened 2020-09-01
Start 'dendron' processes directly

erikjohnston:
Instead of using a python script to run multiple synapse workers we may as well do that in sytest. This just makes things easier to extend.

Built on top of #943


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -937,13 +958,43 @@ sub wrap_synapse_command
...
    }
 
-   return @command;
+   my @base_synapse_command = $self->_generate_base_synapse_command();
+   my $idx = $self->{hs_index};
+
+   $self->_start_process_and_await_notify(
+      setup => [ env => $env ],
+      command => \@base_synapse_command,
+      name => "synaps-$idx-master",
```
clokep, line 972: Is this purposefully missing the "e" on "synapse"?

clokep (commented):
After this PR, does `pydron.py` need to exist anymore? (It's also referenced in `scripts/synapse_sytest.sh`)

lib/SyTest/Homeserver/ProcessManager.pm
```diff
@@ -75,11 +75,13 @@ sub _start_process
          if exists $params{$_};
    }
 
+   my $proc_name = $params{name} // "server";
```
richvdh, line 78: could you document this in the Pod?

richvdh (approved):
there's quite a lot changing at once here, but I think it's ok.

tests/05homeserver.pl
```diff
@@ -137,7 +136,7 @@
 
                # If we just exit then we need to call the AT_END functions
                # manually (if we don't we'll leak child processes).
-               $_->() for @main::AT_END;
+               run_AT_END;
```
erikjohnston, line 139: So, err, I messed up yesterday and thought I tested that doing `$_->() for @main::AT_END;` worked, but it seems that `@main::AT_END` refers to the function `sub AT_END`? I created `run_AT_END` to work around it, but I am a little bit confused how perl was deciding to pick the function rather than the array here, while it does "the right thing" when we do the same thing in `run-tests.pl`. Is it because the variable isn't exposed to other files, only functions are?

richvdh: > Is it because the variable isn't exposed to other files, only functions are?

Yes, this. It is declared as `my @AT_END`, which makes it scoped to the current file. `@main::AT_END` is then a non-existent variable, and in perl it's valid to read from a non-existent array and they will spring into existence on demand.

Normally perl warns about reading from a variable that isn't referenced anywhere else, but the the presence of `sub AT_END` means that the name `main::AT_END` exists in some form, hence suppressing the usual warning. Note that that doesn't mean that `@main::AT_END` refers to the sub: it is still an undefined array.

To do what you wanted, you can replace `my @AT_END` with `our @AT_END`, which will add the definition to the current namespace (`main` in this case). But tbh `run_AT_END` might not be a bad solution anyway. I'd probably spell it with parens for clarity though:

```suggestion
               run_AT_END();
```

richvdh: Other things I discovered/remembered while poking at this (writing them here for want of a better place):

 * a "namespace" is just a hash, hence `print keys(%main::)` tells you what's defined in the `main` namespace. The values of the hash can be either code refs (if only a sub is defined), or a typeglob.
* `sub foo` adds `foo` to the current namespace, and function lookup looks in the current namespace, hence `main::run_AT_END` is not required. Scalar/array/hash lookup does not look in the namespace by default, hence `@main::AT_END` is required to access a var in a different file.

erikjohnston: Aha! That all makes a lot of sense, thanks!

PR #949, opened 2020-09-04
Pin Alien::Sodium to an old version, to fix install errors

richvdh:
Workaround for https://github.com/genio/alien-sodium/issues/2. Fixes
https://github.com/matrix-org/sytest/issues/942.


erikjohnston (approved):
I think this works, but perl???

richvdh:
it's a joy, isn't it.

PR #950, opened 2020-09-05
Test public rooms list with unicode chars

erdnaxeli:
Related to https://github.com/matrix-org/synapse/pull/8262.

Signed-off-by: Alexandre Morignot <erdnaxeli@cervoi.se>


babolivier (approved):
lgtm, thanks!

PR #955, opened 2020-09-16
s/do_request_json_for/do_request_json/

### LINKED ISSUE #954: "After changing password, can log in with new password" tries to login with an access_token
https://github.com/matrix-org/sytest/blob/c0c8009b2a736694823a1a92dadccaa41f61e638/tests/14account/01change-password.pl#L54-L72 

I don't think it should be trying to do that :(

Logs can be found on https://buildkite.com/matrix-dot-org/synapse/builds/11896#

Half-Shot:
Fixes #954 

Fix an issue where /login would be provided a stale access token


anoadragon453 (approved):
LGTM, assuming tests pass.

PR #958, opened 2020-09-22
Run the background worker during worker mode.

clokep:
This is a corresponding PR to matrix-org/synapse#8369 which runs an additional worker for background tasks. I suspect this won't really have impact on sytest since theoretically these shouldn't interfere with tests.


clokep:
I think we would need to manually rebuild the docker image, CI doesn't seem to do that automatically?

clokep:
This will conflict with #945, but hopefully that merges first.

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -253,6 +253,7 @@ sub start
         send_federation       => ( not $self->{dendron} ),
         update_user_directory => ( not $self->{dendron} ),
         enable_media_repo     => ( not $self->{dendron} ),
+        run_background_tasks_on  => ( $self->{dendron} ? "background_worker1" : "master" ),
```
clokep, line 256: I'm not fully confident on what `self->{dendron}` is, but it seems to be only exist when workers are being used.

erikjohnston: Yeah, its basically the thing that enables `pydron` and stuff

clokep:
@erikjohnston Merged in develop, I think this is ready again.

erikjohnston (approved):

PR #960, opened 2020-09-24
Update state idempotency tests to not use initialsync

neilalexander:
This updates the following tests so that they no longer use initialsync:
```
Setting state twice is idempotent
Joining room twice is idempotent
```

Closes matrix-org/dendrite#1311.

### DIFF
--- a/tests/30rooms/01state.pl
+++ b/tests/30rooms/01state.pl
@@ -242,14 +242,24 @@ sub get_test_state
 {
    my ( $user, $room_id, $type ) = @_;
 
-   matrix_initialsync_room( $user, $room_id, limit => 0 )->then( sub {
-      my ( $body ) = @_;
+   matrix_sync( $user,
+      filter            => '{"room":{"rooms":["'.$room_id.'"]},"account_data":{},"presence":{"types":[]}}',
+      update_next_batch => 0,
+      set_presence      => "offline",
+   )->then( sub {
+        my ( $body ) = @_;
 
-      my %state_by_type = partition_by { $_->{type} } @{ $body->{state} };
+        # We expect the event to be in the initial sync
+        sync_timeline_contains( $body, $room_id, sub {
+            my ( $event ) = @_;
+            return unless $event->{type} eq $type and $event->{state_key} eq $user->user_id;
 
-      my $event_id = $state_by_type{ $type }[0]{event_id};
+            log_if_fail "Event for ${\$user->user_id}", $event;
+            Future->done( $event->{event_id} );
+            return 1;
+        }) or die "Expected event not in initial sync";
 
-      Future->done( $event_id );
+        return 0;
    });
 }
 
@@ -259,7 +269,7 @@ sub set_test_state
 
    matrix_put_room_state( $user, $room_id,
          type      => "a.test.state.type",
-         state_key => "",
+         state_key => $user->user_id,
          content   => { "a_key" => "a_value" },
    )->then( sub {
       get_test_state( $user, $room_id, "a.test.state.type" );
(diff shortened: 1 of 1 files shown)


tests/30rooms/01state.pl
```diff
@@ -242,14 +242,24 @@ sub get_test_state
...
    my ( $user, $room_id, $type ) = @_;
 
-   matrix_initialsync_room( $user, $room_id, limit => 0 )->then( sub {
-      my ( $body ) = @_;
+   matrix_sync( $user,
+      filter            => '{"room":{"rooms":["'.$room_id.'"]},"account_data":{},"presence":{"types":[]}}',
+      update_next_batch => 0,
+      set_presence      => "offline",
+   )->then( sub {
+        my ( $body ) = @_;
```
richvdh, line 250: I'm no fan of the 3-space indentation convention, but 5 spaces doesn't seem an improvement...

tests/30rooms/01state.pl
```diff
@@ -242,14 +242,24 @@ sub get_test_state
...
 
-      my %state_by_type = partition_by { $_->{type} } @{ $body->{state} };
+        # We expect the event to be in the initial sync
+        sync_timeline_contains( $body, $room_id, sub {
+            my ( $event ) = @_;
+            return unless $event->{type} eq $type and $event->{state_key} eq $user->user_id;
 
-      my $event_id = $state_by_type{ $type }[0]{event_id};
+            log_if_fail "Event for ${\$user->user_id}", $event;
+            Future->done( $event->{event_id} );
```
richvdh, line 258: this creates a Future which is discarded?

tests/30rooms/01state.pl
```diff
@@ -242,14 +242,24 @@ sub get_test_state
...
+   matrix_sync( $user,
+      filter            => '{"room":{"rooms":["'.$room_id.'"]},"account_data":{},"presence":{"types":[]}}',
+      update_next_batch => 0,
+      set_presence      => "offline",
+   )->then( sub {
+        my ( $body ) = @_;
 
-      my %state_by_type = partition_by { $_->{type} } @{ $body->{state} };
+        # We expect the event to be in the initial sync
+        sync_timeline_contains( $body, $room_id, sub {
```
richvdh, line 253: I think `sync_timeline_contains` returns a truthy result? shouldn't we be doing something with that result (otherwise the call is redundant?)

tests/30rooms/01state.pl
```diff
@@ -259,15 +269,14 @@ sub set_test_state
 
    matrix_put_room_state( $user, $room_id,
          type      => "a.test.state.type",
-         state_key => "",
+         state_key => $user->user_id,
```
richvdh, line 272: why this change, ooi?

tests/30rooms/01state.pl
richvdh: ah I think this should be at the higher level?

richvdh (changes requested):
I think this is a bit broken, as reflected in the fact the tests are failing.

tests/30rooms/01state.pl
neilalexander: VS Code yet again :/

tests/30rooms/01state.pl
neilalexander: Honestly I have no idea. I still don't think I grok how this works looking at it again.

tests/30rooms/01state.pl
neilalexander: I just copy-pasta'd from elsewhere, but I figured that if the timeline didn't hit the `sub` and match the event we want, that the `Future` wouldn't be done?

tests/30rooms/01state.pl
neilalexander: Mostly just to match the filter `return unless $event->{type} eq $type and $event->{state_key} eq $user->user_id;` in `sync_timeline_contains`.

richvdh:
@neilalexander is this still on your radar?

PR #961, opened 2020-09-25
Run worker apps as the generic worker.

clokep:
Synapse workers are now supposed to be run as `synapse.app.generic_worker` instead of the specific app.

This is a smaller bit of #958, but merging it will let me keep working on that locally until I'm sure that things work. 😄 

I think that the Docker container needs to be rebuilt to actually test this?

### DIFF
--- a/docker/pydron.py
+++ b/docker/pydron.py
@@ -64,8 +64,7 @@ try:
 
         base = (
             args["--synapse-python"]
-            + " -m synapse.app."
-            + appname
+            + " -m synapse.app.generic_worker"
             + " -D --config-path="
             + args["--synapse-config"]
             + " --config-path="


erikjohnston:
Have you tried this locally with e.g. federation sender? I know that some of the apps still require using a named app, e.g. federation sender will only send federation if the worker app is configured to be `synapse.app.federation_sender`. I can't remember if it works if you start a worker with a different module than that of the `worker_app` config?

clokep:
> Have you tried this locally with e.g. federation sender? I know that some of the apps still require using a named app, e.g. federation sender will only send federation if the worker app is configured to be `synapse.app.federation_sender`. I can't remember if it works if you start a worker with a different module than that of the `worker_app` config?

Hmm...I tried it locally to start a background worker. I didn't realize that some of the apps have custom code. I'll take another look.

clokep:
Looks like you *can* run a generic worker already, just in a really confusing fashion. I'll update #958 and close this then.

PR #962, opened 2020-09-30
Remove tls_fingerprints

neilalexander:
This removes `tls_fingerprints` since they are no longer in the spec.


anoadragon453 (approved):
lgtm

PR #970, opened 2020-10-06
Fix signal handling in scripts/synapse_sytest.sh

richvdh:
make sure that we pass SIGINTs and SIGTERMs through to the perl process, so
that `docker stop` and ctrl-c actually stop the test runs.


anoadragon453 (approved):
:heart:

PR #972, opened 2020-10-08
Run multiple event persisters when using Redis

erikjohnston:
This makes SyTest test Synapse's sharded event persister set up.

Requires https://github.com/matrix-org/synapse/pull/8499

### DIFF
--- a/tests/30rooms/60version_upgrade.pl
+++ b/tests/30rooms/60version_upgrade.pl
@@ -1070,46 +1070,51 @@ test "Local and remote users' homeservers remove a room from their public direct
             $remote_joiner, $new_room_id, ( server_name => $creator->server_name, )
          );
       })->then(sub {
-         do_request_json_for( $creator,
-            method => "GET",
-            uri    => "/r0/publicRooms",
-         );
-      })->then( sub {
-         # Check public rooms list for local user
-         my ( $body ) = @_;
+         retry_until_success {
+            do_request_json_for( $creator,
+               method => "GET",
+               uri    => "/r0/publicRooms",
+            )->then( sub {
+               # Check public rooms list for local user
+               my ( $body ) = @_;
 
-         log_if_fail "Public rooms list for local user", $body;
+               log_if_fail "Public rooms list for local user", $body;
 
-         assert_json_keys( $body, qw( chunk ) );
+               assert_json_keys( $body, qw( chunk ) );
 
-         # Check that the room list contains new room id
-         any { $new_room_id eq $_->{room_id} } @{ $body->{chunk} }
-            or die "Local room list did not include expected room id $new_room_id";
+               # Check that the room list contains new room id
+               any { $new_room_id eq $_->{room_id} } @{ $body->{chunk} }
+                  or die "Local room list did not include expected room id $new_room_id";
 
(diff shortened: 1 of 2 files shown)


clokep:
@erikjohnston Looks like this has some failing tests!

tests/30rooms/60version_upgrade.pl
```diff
@@ -1070,46 +1070,51 @@ sub upgrade_room_synced {
...
          );
       })->then(sub {
-         do_request_json_for( $creator,
-            method => "GET",
-            uri    => "/r0/publicRooms",
-         );
-      })->then( sub {
-         # Check public rooms list for local user
-         my ( $body ) = @_;
+         retry_until_success {
```
richvdh, line 1073: why are these changes necessary? And if they are really necessary, can we limit the number of times we retry, rather than going on forever?

richvdh (changes requested):

tests/30rooms/60version_upgrade.pl
erikjohnston: This may no longer be necessary, but I don't think the public room lists in general get updated immediately (as they can be populated via background processes).

The `retry_until_success` will get killed by the test timeout, which is how we use `retry_until_success`/`repeat_until_true` elsewhere?

richvdh: could you add a comment to say that we're retrying because the directory is updated asynchonously, or sth.

My main gripe with `retry_until_success` is that it turns test failures into test timeouts, which we tend to ignore as "oh we had a slow VM for random reasons", so a limit on the number of retries is preferable if possible.

The other problem with `retry_until_success` is that it swallows the failure, so you can't see what's going wrong. https://github.com/matrix-org/sytest/blob/develop/tests/30rooms/04messages.pl#L387-L408 includes a bunch of code to log what's going on, but on closer inspection maybe we should put all that boilerplate into `retry_until_success` (which could maybe also benefit from a "try N times" param which defaults to something other than "infinity").

erikjohnston: Have added a comment.

Yeah, I think that sounds like a general issue with `retry_until_success`. We use this pattern a lot so it feels like something to investigate and improve separately.

tests/30rooms/60version_upgrade.pl
```diff
@@ -1070,6 +1070,8 @@ sub upgrade_room_synced {
             $remote_joiner, $new_room_id, ( server_name => $creator->server_name, )
          );
       })->then(sub {
+         # The room list can be updated asynchrounously, so we retry if it
```
richvdh, line 1073:
```suggestion
         # The room list can be updated asynchronously, so we retry if it
```

tests/30rooms/60version_upgrade.pl
```diff
@@ -1093,6 +1095,8 @@ sub upgrade_room_synced {
             })
          }
       })->then(sub {
+         # The room list can be updated asynchrounously, so we retry if it
```
richvdh, line 1098:
```suggestion
         # The room list can be updated asynchronously, so we retry if it
```

richvdh (approved):

PR #974, opened 2020-10-15
Fix a missing Future->done

richvdh:
Seems to have broken in #972. What I don't understand is why it is still working in the CI.


richvdh:
oh, it turns out this is allowed in Future 0.45: https://metacpan.org/source/PEVANS/Future-0.45/Changes#L5. If we're dropping support for earlier versions we need to note this in `cpanfile`.

erikjohnston (approved):

PR #976, opened 2020-10-16
Mark "Read receipts are sent as events" as deprecated

S7evinK:
As per the [Spec](https://matrix.org/docs/spec/client_server/r0.6.1#deprecated-get-matrix-client-r0-events) the endpoint `/_matrix/client/r0/events` is deprecated.

### DIFF
--- a/tests/30rooms/21receipts.pl
+++ b/tests/30rooms/21receipts.pl
@@ -118,6 +118,7 @@ multi_test "Read receipts are visible to /initialSync",
    };
 
 test "Read receipts are sent as events",
+   deprecated_endpoints => 1,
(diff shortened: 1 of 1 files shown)


Kegsay:
Shouldn't we be replacing this to use `/sync` instead of `/events`?

S7evinK:
Isn't that already tested in https://github.com/matrix-org/sytest/blob/cab0f6b3a0ae8c8324f5c8ab609ac6286da178ca/tests/31sync/12receipts.pl#L1
and
https://github.com/matrix-org/sytest/blob/cab0f6b3a0ae8c8324f5c8ab609ac6286da178ca/tests/31sync/12receipts.pl#L55
?

neilalexander (approved):

PR #977, opened 2020-10-16
Mark "Read receipts are sent as events" as deprecated

neilalexander:
The `Read receipts are sent as events` test uses `/events`, which is deprecated, and there are already other tests which check the same functionality using non-deprecated endpoints:

- `Read receipts appear in initial v2 /sync`
- `New read receipts appear in incremental v2 /sync`


richvdh:
dup of #976?

neilalexander:
Yes, looks like it is.

PR #981, opened 2020-11-16
Implementation-specific tests

neilalexander:
This PR adds a test parameter `implementation_specific`, which instructs Sytest to run a test for a specific implementation only. It also marks the admin tests as Synapse-specific.

### DIFF
--- a/lib/SyTest/HomeserverFactory/Dendrite.pm
+++ b/lib/SyTest/HomeserverFactory/Dendrite.pm
@@ -31,6 +31,11 @@ sub _init
    $self->SUPER::_init( @_ );
 }
 
+sub implementation_name
+{
+   return "dendrite";
+}
+
 sub get_options
 {
    my $self = shift;
--- a/lib/SyTest/HomeserverFactory/Dendrite.pm
+++ b/lib/SyTest/HomeserverFactory/Dendrite.pm
@@ -70,6 +70,11 @@ sub _init
    $self->SUPER::_init( @_ );
 }
 
+sub implementation_name
+{
+   return "dendrite";
+}
+
 sub create_server
 {
    my $self = shift;
--- a/lib/SyTest/HomeserverFactory/Dendrite.pm
+++ b/lib/SyTest/HomeserverFactory/Dendrite.pm
@@ -70,11 +70,6 @@ sub _init
    $self->SUPER::_init( @_ );
 }
 
-sub implementation_name
-{
-   return "dendrite";
-}
-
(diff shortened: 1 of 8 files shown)


anoadragon453:
This generally looks fine to me. I'm a little wary of adding *yet another* way of excluding tests (on top of the test whitelist/blacklist), and generally the right solution to me sounds like we should just rewrite these tests in Synapse's unit test suite.

Then again, I don't particularly want to do the latter... so this seems like a good alternative. Though I'm not entirely sure if it could be useful in any other case.

@richvdh curious for your thoughts, but otherwise lgtm.

run-tests.pl
```diff
@@ -622,6 +622,13 @@ sub _push_test
        }
    }
 
+   if( exists $params{implementation_specific} ) {
+       my @tokens = split /::/, $SERVER_IMPL;
+       if ( $tokens[0] ne $params{implementation_specific} ) {
```
richvdh, line 627: This feels slightly too magical and fragile; could we give the Homeserver classes an `implementation_name` method (see `server_name` as an existing example) or something instead?

richvdh (commented):
looks nice otherwise.

run-tests.pl
neilalexander: I have added the `implementation_name` method to the `HomeserverFactory`, since that seems to be easily accessible from `run_tests.pl`.

lib/SyTest/HomeserverFactory/Synapse.pm
```diff
@@ -103,6 +108,11 @@ sub _init
    $self->{args}{redis_host} = "";
 }
 
+sub implementation_name
+{
+   return "synapse";
+}
+
```
richvdh, line 115: this should be unnecessary, as it will be inherited from `SyTest::HomeserverFactory::Synapse` ?

lib/SyTest/HomeserverFactory/Dendrite.pm
```diff
@@ -65,6 +70,11 @@ sub _init
    $self->SUPER::_init( @_ );
 }
 
+sub implementation_name
+{
+   return "dendrite";
+}
```
richvdh, line 76: ditto

richvdh (changes requested):
can you add a stub impl to `HomeserverFactory` itself, with some comments to document the interface?

(ideally HomeserverFactory would have better documentation in general)

richvdh (approved):
lgtm!

PR #982, opened 2020-11-17
Timeout enlarge

bodqhrohro:
With 60 seconds timeout some tests fail to finish in time on my weak laptop, so I suggest to increase it, if that wouldn't have any negative impact on CI or whatever.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 .vscode
 .coverage*
 .idea
+/signingkeyserver.db
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -758,7 +758,7 @@ sub _run_test0
 
       Future->wait_any(
          $f_setup,
-         $loop->delay_future( after => 60 )
+         $loop->delay_future( after => 300 )
             ->then_fail( "Timed out waiting for setup" )
       )->get;
 
--- a/changelog.d/982.bugfix
+++ b/changelog.d/982.bugfix
@@ -1,0 +1,2 @@
+- Add signingkeyserver.db to .gitignore. Thanks to @bodqhrohro
+- Increase the test timeout to 300 seconds. Thanks to @bodqhrohro
--- a/changelog.d/982.bugfix
+++ b/changelog.d/982.bugfix
@@ -1,3 +1,0 @@
-- Add signingkeyserver.db to .gitignore. Thanks to @bodqhrohro
-- Increase the test timeout to 300 seconds. Thanks to @bodqhrohro
-- Increase the setup timeout to 300 seconds. Thanks to @bodqhrohro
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@
 .vscode
 .coverage*
 .idea
+/signingkeyserver.db
```
anoadragon453, line 16: Mmm, what generates this file? Also ideally this change would be in a separate PR :slightly_smiling_face:

run-tests.pl
```diff
@@ -758,7 +758,7 @@ sub _run_test0
 
       Future->wait_any(
          $f_setup,
-         $loop->delay_future( after => 60 )
+         $loop->delay_future( after => 300 )
```
anoadragon453, line 761: 300 seconds is a bit of a jump from 60. Perhaps we could start with x2 = 120?

What is the maximum amount of time a test takes on your machine?

anoadragon453 (commented):
Hello, and thank you for your PR :slightly_smiling_face: 

Could you tell us whether there is a particular test that's timing out, or are you having trouble with lots of them?

Also, an alternative solution (though a more complex one) would be to just make the test timeout configurable.

anoadragon453:
Hmm, not entirely sure where the CI failures are coming from. The error line is:

```
# fixture failed - TODO: Missing fixture dependency can_register_with_secret
```

You may need to merge latest develop to your branch.

bodqhrohro:
Well, after some investigation, I find this PR useless. There are lots of other timeouts in the code.

Instead it should probably be implemented other way: as an environment variable specifying a factor that **all** hardcoded timeouts should be multiplied to. This way, only those having a weak machine will have to tune it, in a much more flexible way.

Should I close it for now and post an issue about the new approach, until me or someone else implements it?

.gitignore
bodqhrohro: Dendrite does.

Does it happen because this file is not yet overridden in `Dendrite.pm` to Postgres like other files are? Should `*.db` be in `.gitignore` instead for the case when the tests are run without the `POSTGRES` variable?

run-tests.pl
bodqhrohro: How do I measure it? TAP doesn't seem to print start/end time for successful tests by default.

Actually, there's no upper limit, as the machine is loaded with other tasks and may easily go into the thrashing state.

bodqhrohro: Experiments with #996 show that with ×2 scaling, there still are 73 tests timing out; with ×3 and more, there still remain about 22–24 tests timing out, which all seem to be expected to fail because of timeouts though.

bodqhrohro:
Replaced by #996.

PR #983, opened 2020-11-18
Add tests for allowed symbols in user identifiers

bodqhrohro:
https://matrix.org/docs/spec/appendices#user-identifiers

Signed-off-by: Bohdan Horbeshko bodqhrohro@gmail.com

### DIFF
--- a/changelog.d/983.feature
+++ b/changelog.d/983.feature
@@ -1,0 +1,1 @@
+- Add tests for allowed symbols in user identifiers. Thanks to @bodqhrohro
--- a/changelog.d/983.feature
+++ b/changelog.d/983.feature
@@ -1,1 +1,0 @@
-- Add tests for allowed symbols in user identifiers. Thanks to @bodqhrohro
--- a/tests/10apidoc/01register.pl
+++ b/tests/10apidoc/01register.pl
@@ -183,6 +183,11 @@ foreach my $chr (split '', '!":?\@[]{|}£é' . "\n'" ) {
 
 foreach my $chr (split '', 'q3._=-/' ) {
    test "POST /register allows registration of usernames with '$chr'",
+      # Note: this test relies on somewhat implementation-specific details,
+      # in that it assumes that the `username` presented to `/register` maps
+      # directly onto an MXID. In reality, a server is free to accept or reject
+      # any `usernames` it wants, provided it maps those it accepts onto
+      # valid MXIDs.
       requires => [ $main::API_CLIENTS[0],
                     qw( can_register_dummy_flow ) ],
 
--- a/tests/10apidoc/01register.pl
+++ b/tests/10apidoc/01register.pl
@@ -180,6 +180,32 @@ foreach my $chr (split '', '!":?\@[]{|}£é' . "\n'" ) {
       };
 }
 
+
+foreach my $chr (split '', 'q3._=-/' ) {
+   test "POST /register allows registration of usernames with '$chr'",
+      requires => [ $main::API_CLIENTS[0],
+                    qw( can_register_dummy_flow ) ],
+
+      do => sub {
+         my ( $http ) = @_;
+
+         my $reqbody = {
+            auth => {
+               type => "m.login.dummy",
+            },
+            username => 'chrtestuser'.$chr,
+            password => "sUp3rs3kr1t",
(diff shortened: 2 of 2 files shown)


erikjohnston:
Hi, sorry its taken a while to get back to you! It's great to see people diving into SyTest :slightly_smiling_face: 

I've been trying to think if this is something that we want in SyTest, namely there is nothing in the spec that *requires* servers to accept these characters in names (indeed, synapse will refuse to allow you to register usernames starting with a number or underscore). 

On the other hand, it is somewhat useful to test whether a server accepts the full range of characters in usernames or not. I guess ideally we'd somehow mark this test as "informational, but not required for spec compliance", we do have a bunch of tests already that sort of ish fit into similar categories (mainly Synapse specific tests), but I don't think we currently have a way of actually marking tests.

I think I am probably mildly in favour of merging this PR, and if other server implementations run into issues we can cross that bridge down the line. I'll wait a bit before merging in case anyone else has opinions.

richvdh:
it's also worth noting that there is nothing that requires the username passed to `register` to be the same as the mxid that the server allocates that user: the server is free to allocate mxids sequentially if it so chooses, for example.

We should probably at least have a comment on the test to say that it's making assumptions about the server implementation?

bodqhrohro:
>there is nothing that requires the username passed to register to be the same as the mxid that the server allocates that user

Uh oh, I missed that point.

Does it mean that the tests for rejected characters (which I based these tests on) are making this assumption too? Can a home server, for example, automatically escape the prohibited characters in MXID instead of returning `M_INVALID_USERNAME`?

richvdh:
> Does it mean that the tests for rejected characters (which I based these tests on) are making this assumption too?

yes, indeed they are. 

>  Can a home server, for example, automatically escape the prohibited characters in MXID instead of returning `M_INVALID_USERNAME`?

yes.

but as erik says: we don't have a great way of testing this stuff otherwise, and given that all the server implementations we know about handle these scenarios in the way sytest assumes, it's probably ok - but I do think that we should include comments in the tests where we are making new assumptions about implementation.

bodqhrohro:
>but I don't think we currently have a way of actually marking tests

So the `implementation_specific` argument is not exactly for denoting the Synapse-like behaviour, right?

Then I guess it should be split into two arguments: one for strictly implementation-specific tests, and one for behaviour-specific tests, which may be used for HS that strive to mimick the behaviour of other HS beyond the specification. So it would become possible to run something like `-I Dendrite::Monolith -M Synapse`, where `-M` is for "mimick".

Should I post a separate issue about it and mark the tests with `implementation_specific => "synapse"` for now?

richvdh:
> Then I guess it should be split into two arguments: one for strictly implementation-specific tests, and one for behaviour-specific tests, 

I'm not sure, to be honest. That sounds a bit overcomplicated to me, and I don't think it's a big enough problem to make that worthwhile.

changelog.d/983.feature
```diff
@@ -0,0 +1 @@
+- Add tests for allowed symbols in user identifiers. Thanks to @bodqhrohro
```
richvdh, line 1: we don't keep a changelog for sytest, and we don't use newsfiles like this for it. Thank you for your diligence in adding one, but it's redundant here. Please could you remove it?

tests/10apidoc/01register.pl
```diff
@@ -180,6 +180,32 @@
       };
 }
 
+
+foreach my $chr (split '', 'q3._=-/' ) {
+   test "POST /register allows registration of usernames with '$chr'",
```
richvdh, line 185:
```suggestion
   test "POST /register allows registration of usernames with '$chr'",
      # Note: this test relies on somewhat implementation-specific details,
      # in that it assumes that the `username` presented to `/register` maps
      # directly onto an MXID. In reality, a server is free to accept or reject
      # any `usernames` it wants, provided it maps those it accepts onto
      # valid MXIDs.
```

richvdh (changes requested):
I suggest you just add a comment like this

richvdh:
Thanks. I'm pretty sure the CI is failing for unrelated reasons, so I'm going to go ahead and merge this

richvdh (approved):

PR #986, opened 2020-12-02
Fix push URLs to match the spec.

clokep:
This matches matrix-org/synapse#8865 to ensure the push URLs match the spec.

### DIFF
--- a/tests/61push/08_rejected_pushers.pl
+++ b/tests/61push/08_rejected_pushers.pl
@@ -1,3 +1,6 @@
+# Per the specification HTTP pushers must point to the following location.
+my $PUSH_LOCATION = "/_matrix/push/v1/notify";
+
 sub create_pusher
 {
    my ( $user, $app_id, $push_key, $url ) = @_;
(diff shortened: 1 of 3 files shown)


richvdh (commented):
in principle this looks good, but it seems like there are more instances of non-compliance.

clokep:
Sorry about that! I thought I caught all the instances. 😢

tests/61push/08_rejected_pushers.pl
```diff
@@ -20,17 +23,22 @@ sub create_pusher
...
+
+      # Ensure this is the expected pusher.
+      return unless $body->{notification}{devices};
+      return unless $body->{notification}{devices}[0]{pushkey} eq $pushkey;
+
+      # Respond to expected request.
+      $request->respond_json( $response // {} );
+
       return 1;
    });
```
clokep, line 43: This is attempting to check one of the fields of the request instead of differentiating on URL.

erikjohnston (approved):

PR #987, opened 2020-12-09
Use an empty ip_range_blacklist to allow connecting to localhost.

clokep:
Corresponds to matrix-org/synapse#8870. Empties out the new `ip_range_blacklist` setting (similar to `federation_ip_range_blacklist`) so that connections to private IPs work OK.


anoadragon453 (approved):
LGTM.

PR #990, opened 2020-12-10
Ensure that SSO users are registered before trying to UI auth them.

clokep:
We have some tests for UI auth + SSO, but they're performed with users who aren't actually logged in via SSO, so the resulting authentication flows don't really make sense.

This modifies the UI auth + SSO tests to log the users in via SSO, not via normal login. (This allows a homeserver to know that they're an SSO user, not user that logs in via a password.)

The diff here is pretty awful, but I pretty much abstracted the "Can login with new user via CAS" to a sub (`matrix_login_with_cas`) that then gets used in multiple places. It also pulls the `cas_login_fixture` into the UI auth tests as well and adds a new sub `generate_cas_response` instead of hard-coding the same XML in many places.

### DIFF
--- a/tests/10apidoc/13ui-auth.pl
+++ b/tests/10apidoc/13ui-auth.pl
@@ -1,7 +1,7 @@
 use JSON qw( decode_json );
 use URI::Escape;
 
-our @EXPORT = qw( wait_for_cas_request );
+our @EXPORT = qw( wait_for_cas_request generate_cas_response matrix_login_with_cas );
 
 my $cas_login_fixture = fixture(
    requires => [ $main::API_CLIENTS[0] ],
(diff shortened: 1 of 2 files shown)


clokep:
This is necessary for matrix-org/synapse#8856

tests/10apidoc/13ui-auth.pl
```diff
@@ -29,17 +82,132 @@ sub wait_for_cas_request
...
+
+      # step 3: client sends credentials to CAS server
+      # step 4: CAS redirects, with a ticket number, back to synapse.
+      #
+      # For this test, we skip this bit, as it's nothing to do with synapse,
+      # really.
+      #
+      # The URI that CAS redirects back to is the value of the 'service'
+      # param we gave it, with an additional "ticket" query-param:
+      my $login_uri = $service . "&ticket=$CAS_TICKET";
```
erikjohnston, line 135:
```suggestion
      my $login_uri = "$service&ticket=$CAS_TICKET";
```

no?

tests/10apidoc/13ui-auth.pl
```diff
@@ -29,17 +82,132 @@ sub wait_for_cas_request
...
+         method => "GET",
+         uri    => "/r0/login/sso/redirect",
+         params => {
+            redirectUrl => $REDIRECT_URL,
+         },
+      ),
+   )->then( sub {
+      my ( $cas_request, $_cas_response ) = @_;
+      log_if_fail( "Initial CAS request query:",
+                   $cas_request->query_string );
```
erikjohnston, line 121: We tend not to put parens around `log_if_fail`. Do we need them here?

erikjohnston (approved):
I think this looks broadly sane?

tests/10apidoc/13ui-auth.pl
clokep: Nope. I've made this change in the entire file.

PR #993, opened 2020-12-29
Add signing_key_server entry to _get_config.

perj:
Without this, tests still work, unless you run with a readonly /sytest
directory. A default signingkeyserver.db will be created there.

That seems unfortunate however, better to add an entry for it. Also, the
docker/README.md instructions do say to mount /sytest readonly so they
do not work without this fix.

Signed-off-by: Per Johansson <per@morth.org>


perj:
Obsoleted by #1000

PR #995, opened 2020-12-29
Ignore generated files when running on macOS

MadLittleMods:
Ignore generated files when running on macOS

 - Ignore `.DS_Store` from macOS
 - Ignore `*.db` files that can get generated from running (like `signingkeyserver.db`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,9 @@
 /synapse
 /test-server
 /var
+*.db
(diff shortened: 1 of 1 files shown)


erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston 🐕

PR #996, opened 2020-12-29
Add TIMEOUT_FACTOR environment variable

bodqhrohro:
Signed-off-by: Bohdan Horbeshko <bodqhrohro@gmail.com>

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -39,6 +39,7 @@ The following environment variables can be set with `-e` to control the test run
  * `OFFLINE`: set non-empty to avoid updating the python or perl dependencies.
  * `BLACKLIST`: set non-empty to change the default blacklist file to the
    specified path relative to the Synapse directory
+ * `TIMEOUT_FACTOR`: sets a number that test timeouts are multiplied to
 
 Some examples of running Synapse in different configurations:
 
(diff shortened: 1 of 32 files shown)


docker/README.md
```diff
@@ -39,6 +39,7 @@ The following environment variables can be set with `-e` to control the test run
  * `OFFLINE`: set non-empty to avoid updating the python or perl dependencies.
  * `BLACKLIST`: set non-empty to change the default blacklist file to the
    specified path relative to the Synapse directory
+ * `TIMEOUT_FACTOR`: sets a number that test timeouts are multiplied to
```
anoadragon453, line 42:
```suggestion
 * `TIMEOUT_FACTOR`: sets a number that test timeouts are multiplied by.
```

anoadragon453 (commented):
The code looks like it'll work. I'll leave that to someone with a bit more perl experience to give the final sign-off to though.

I'm happy to see this be a configurable option rather than a new default! However I am a bit concerned about the design. I could imagine new code being written without remembering to add `$TIMEOUT` in.

Perhaps at least the `matrix_*` methods could multiply the timeout by `$TIMEOUT` internally?

bodqhrohro:
> Perhaps at least the `matrix_*` methods could multiply the timeout by `$TIMEOUT` internally?

I intentionally left a possibility to avoid the factor, for a case if some tests require (or will require) a strict non-flexible delay. Though if this is irrellevant, the factor may be enforced more deeply; after all, such fastidious tests can just *divide* a delay by the factor to compensate it :)

erikjohnston (approved):
I think this makes sense, thanks!

PR #998, opened 2020-12-29
Use Dendrite docs as source of truth for Sytest

MadLittleMods:
Use Dendrite docs as source of truth for how use Sytest.

The Dendrite docs seem more up to date and actually worked for me, https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md#using-the-sytest-docker-image

The current docs in this repo for Dendrite didn't work for me and wasted a lot of time before I found the other one, https://github.com/matrix-org/sytest/blob/11240a80c42994a3de87f6e16b9452dacff2f2c4/docs/dendrite-setup.md


Kegsay (approved):

MadLittleMods:
Thanks for the review @Kegsay 🦈

PR #1001, opened 2021-01-07
Test that handling to device messages off master works

erikjohnston:
Test for https://github.com/matrix-org/synapse/pull/9044

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -307,7 +307,7 @@ sub start
               host => "$bind_host",
               port => $self->{ports}{event_persister2},
            },
-	   "client_reader" => {
+           "client_reader" => {
               host => "$bind_host",
               port => $self->{ports}{client_reader},
            },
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -307,10 +307,15 @@ sub start
...
            },
+	   "client_reader" => {
+              host => "$bind_host",
+              port => $self->{ports}{client_reader},
+           },
         },
 
         stream_writers => {
            events => $self->{redis_host} ne '' ? [ "event_persister1", "event_persister2" ] : "master",
+	   to_device => $self->{redis_host} ne '' ? [ "client_reader" ] : "master",
```
erikjohnston, line 318: There's no particular reason to choose `client_reader`, but I couldn't think of a better place and I'm not sure we want to add more workers at this point

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -307,10 +307,15 @@ sub start
               host => "$bind_host",
               port => $self->{ports}{event_persister2},
            },
+	   "client_reader" => {
```
richvdh, line 310:
```suggestion
	    "client_reader" => {
```

lib/SyTest/Homeserver/Synapse.pm
richvdh, line 318:
```suggestion
	     to_device => $self->{redis_host} ne '' ? [ "client_reader" ] : "master",
```

richvdh: maybe stick those words in a comment

richvdh (approved):
looks fine otherwise

lib/SyTest/Homeserver/Synapse.pm
erikjohnston: Done :)

PR #1002, opened 2021-01-07
Rename sytest-synapse:py38 to :py39

callahad:
These are based on Debian Bullseye.

As of last month, Bullseye ships Python 3.9:

  https://packages.debian.org/bullseye/python3

  https://tracker.debian.org/news/1199188/python3-defaults-390-4-migrated-to-testing/

Signed-off-by: Dan Callahan <danc@element.io>


richvdh (approved):
lgtm.

PR #1007, opened 2021-01-13
Test moving of account_data and receipts off master

erikjohnston:
Tests https://github.com/matrix-org/synapse/pull/9104

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -319,7 +319,9 @@ sub start
            # There's no particular reason to choose client_reader, but I
            # couldn't think of a better place and I'm not sure we want to add
            # more workers at this point
-           to_device => $self->{redis_host} ne '' ? [ "client_reader" ] : "master",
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -1269,6 +1271,10 @@ EOCONFIG
       $haproxy_map .= <<'EOCONFIG';
 
 ^/_matrix/client/(api/v1|r0|unstable)/sendToDevice/          client_reader
+^/_matrix/client/(api/v1|r0|unstable)/rooms/.*/tag           client_reader
```
clokep, line 1274: Isn't this `tags`, not `tag`?

clokep (dismissed):

lib/SyTest/Homeserver/Synapse.pm
erikjohnston: It's `/tag/:type` and `/tags$`, and this (rather cheekily) covers both

clokep: Ah, right. 👍

clokep (approved):

PR #1011, opened 2021-02-08
Use a consistent public baseurl.

clokep:
This sets `public_baseurl` in the Synapse config to match what is provided via `ServerInfo`.

Note that the `cas_config.service_url` was deprecated in matrix-org/synapse#9199 so we don't need to provide it anymore.

Hopefully will fix the build issues in matrix-org/synapse#9313.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -152,7 +152,12 @@ sub start
 
    my $listeners = [ $self->generate_listeners ];
    my $bind_host = $self->{bind_host};
-   my $secure_port = $self->secure_port;
+
+   # run-tests.pl defines whether TLS should be used or not.
+   our $WANT_TLS;
+   my $public_baseurl = $WANT_TLS ?
+      "https://${bind_host}:" . $self->secure_port() :
+      "http://${bind_host}:" . $self->unsecure_port();
 
    my $macaroon_secret_key = "secret_$port";
    my $registration_shared_secret = "reg_secret";
@@ -172,7 +177,7 @@ sub start
    my $config_path = $self->{paths}{config} = $self->write_yaml_file( "config.yaml" => {
         server_name => $self->server_name,
         log_config => $log_config_file,
-        public_baseurl => "https://${bind_host}:$secure_port",
+        public_baseurl => $public_baseurl,
 
         # We configure synapse to use a TLS cert which is signed by our dummy CA...
         tls_certificate_path => $self->{paths}{cert_file},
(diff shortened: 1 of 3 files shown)


clokep:
The failing logs in this case seems to have an additional:

```
# localhost:8800 - connect: Cannot assign requested address failed [Cannot assign requested address] at /sytest/run-tests.pl line 775.
```

But not sure why that's happening from this change.

clokep:
> But not sure why that's happening from this change.

This was because `$self->{ports}{synapse}` is only the public port when *not* using haproxy. When using haproxy this should be `$self->{ports}{haproxy}`, that logic is already encapsulated in the `secure_port` helper.

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -172,7 +172,7 @@ sub start
    my $config_path = $self->{paths}{config} = $self->write_yaml_file( "config.yaml" => {
         server_name => $self->server_name,
         log_config => $log_config_file,
-        public_baseurl => "http://${bind_host}:$unsecure_port",
+        public_baseurl => "https://${bind_host}:$secure_port",
```
clokep, line 175: I do wonder if this has to take into account the `$WANT_TLS` option somewhere? 🤷

erikjohnston: Does it work with `--no-tls` still with this? Can we use the unsecure port if its enabled or does that break stuff?

clokep: TIL that was a flag. 😄 It did break with that, I pushed a commit fixing it.

erikjohnston: Me too! I was like "WTF is WANT_TLS??"

erikjohnston (approved):

neilalexander:
Looks like this PR broke Sytest for Dendrite:
```
fixture failed - Can't locate object method "public_baseurl" via package "SyTest::Homeserver::Dendrite::Monolith" at tests/05homeserver.pl line 56.
```
Apparently it wasn’t highlighted in the checks here because the Dendrite pipelines are soft-failed — we may want to do something about that too.

clokep:
Oh, sorry about that! I saw all green ✅ and assumed it was OK. 😢 I agree that we should update the pipelines to make Dendrite pass.

richvdh:
I'm really not thrilled about this change: it's very confusing, and lacks documentation.

`Homeserver` now has both a `$hs->public_baseurl` method, and a `$hs->{public_baseurl}` member, which are subtly different, and neither are documented.

Adding required methods to a base class without checking the implementations of that class is pretty poor form too.

Finally, @neilalexander: the commit comment on https://github.com/matrix-org/sytest/commit/eed37c2a728314f4f39c2920d5992b95c77e0f73 is confusing: it looks like it was part of this PR, whereas it wasn't. Please don't commit anything direct to develop that isn't completely trivial, and please don't add PR numbers to the subject unless it's actually part of the PR in question.

I think I'm going to have a go at cleaning up public_baseurl.

richvdh:
#1014 sets out to improve this.

PR #1013, opened 2021-02-22
Avoid swallowing stdout on errors

richvdh:
Make sure that stderr and stdout get correctly caught and logged, to help diagnose mysterious "failed to start server" errors. Things written direct to stderr tend to get overwritten by other diagnostic output, so use the proper `Output::diag` method.

Previously, stderr was being caught and returned in the failure, which wasn't terribly clear, while stdout went to /dev/null.

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -556,12 +556,18 @@ sub _run_command
 
    my $on_output = sub {
       my ( $stream, $buffref, $eof) = @_;
+
+      # write each complete line in the buffer
       while ( $$buffref =~ s/^(.*)\n// ) {
          &$diag($1);
       }
+
+      # if this is the end of the output, and there is
+      # anything left in the buffer, print out the remainder.
       if( $eof && $$buffref ) {
          &$diag($$buffref);
       }
+
       return 0;
    };
 
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -536,7 +536,9 @@ sub await_connectable
 This method runs a specified command and returns a future which will complete
 when the process exits.
 
-The parameters are passed to C<IO::Loop->run_child>.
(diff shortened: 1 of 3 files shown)


lib/SyTest/Homeserver.pm
```diff
@@ -566,13 +586,11 @@ sub _run_command
...
 
-         if( $stderr ) {
-            $failure .= ": $stderr";
-         }
          $fut->fail( $failure );
-      }
+      },
    );
 
+   $self->loop->add($proc);
```
anoadragon453, line 593: I'm not sure what the purpose of this line is?

lib/SyTest/Homeserver.pm
```diff
@@ -546,16 +548,34 @@ sub _run_command
...
+   };
+
+   my $on_output = sub {
+      my ( $stream, $buffref, $eof) = @_;
+      while ( $$buffref =~ s/^(.*)\n// ) {
+         &$diag($1);
+      }
+      if( $eof && $$buffref ) {
+         &$diag($$buffref);
+      }
```
anoadragon453, line 564: Is this bit for printing out the final line without a `\n` at the end?

anoadragon453 (approved):
Looks like perl.

I just had some questions about some bits of the code that could maybe do with some comments to explain them. On the whole looks good though (and yay for not having to dig into results.tap!).

lib/SyTest/Homeserver.pm
richvdh: heh, nor am I really. It's just some incantation you have to do when you use `IO::Async::Process`. See the examples at https://metacpan.org/pod/IO::Async::Process#SYNOPSIS.

lib/SyTest/Homeserver.pm
richvdh: yeah, exactly. I'll add a comment.

PR #1014, opened 2021-02-23
Drop --no-tls and clean up Homeserver object API

richvdh:
This was getting to be a bit of a mess, with unused/duplicate functions, and undocumented methods named the same as undocumented fields with different functions. Clean it all up.

Commits should be reviewable independently.


clokep:
Looks like a nice clean-up! 👍 Thanks!

The dendrite builds are soft-failing on this branch, it looked like only a few tests were failing though. Maybe @neilalexander knows what's going on?

richvdh:
It looks to me like the same tests are failing here as on develop (see https://buildkite.com/matrix-dot-org/sytest/builds/1297), so I think those are separate problems.

PR #1015, opened 2021-02-24
Fix the headers on HTTP requests for synapse worker mode

richvdh:
A few different problems here, all needed to fix the requests being sent to Synapse. Commits should be reviewable independently.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -1170,6 +1170,7 @@ defaults
 
 frontend http-in
     bind ${bind_host}:$ports->{haproxy} ssl crt $self->{paths}{pem_file}
+    http-request set-header X-Forwarded-Proto https if { ssl_fc }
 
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -1095,6 +1084,23 @@ sub _start_synapse
...
+   my $self = shift;
+
+   return
+      {
+         type         => "http",
+         port         => $self->{ports}{synapse_unsecure},
+         bind_address => "127.0.0.1",
+         x_forwarded  => JSON::true,
+         resources    => [{
+            names => [ "client", "federation", "replication" ]
```
clokep, line 1098: This drops "metrics" since it has a separate listener?

clokep (approved):
LGTM!

lib/SyTest/Homeserver/Synapse.pm
richvdh: that, and the fact that we just don't use the metrics in sytest.

PR #1017, opened 2021-03-04
Fix username field in AS tests

neilalexander:
Apparently the `"user"` field is old non-spec-compliant behaviour and we should be using `"username"` now: https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-register


Kegsay (approved):

richvdh:
when you find stuff like this, please can you open an issue against synapse?

neilalexander:
Sure thing, I'll make one now.

PR #1021, opened 2021-04-01
Remove Debian Stretch containers

### LINKED ISSUE #1003: Tags for sytest-synapse images should be renamed
Currently, our `sytest-synapse` images are built with different Debian base images to yield images with an assortment of Python and PostgreSQL versions:

https://github.com/matrix-org/sytest/blob/6b1ae9efdd6d7128daeb31260755b2bdcc8a7348/docker/build.sh#L9-L11

This is misleading and error-prone, as `bullseye` is currently Debian's testing release, and thus subject to version upgrades.

For example, the image tagged `py38` and described as "Python 3.8 / Postgres 12 / Monolith" in the [Synapse pipeline](https://github.com/matrix-org/pipelines/blob/8cdf2a59660b83e5a0366928afec9387a5593b6c/synapse/pipeline.yml#L367) would currently build with Python 3.9 and PostgreSQL 13. See #1002

## Proposal

- Use Debian release names for the tags of stable Debian releases
- Use `testing` (and `unstable`, etc.) as the names of other releases

This would yield:

- `sytest-synapse:py35` → `sytest-synapse:stretch`
- `sytest-synapse:py37` → `sytest-synapse:buster`
- `sytest-synapse:py38` → `sytest-synapse:testing`

callahad:
Renames remaining containers based on underlying Debian distro to be a
bit more explicit about the testing environment.

Fixes #1003

Signed-off-by: Dan Callahan <danc@element.io>

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -6,14 +6,12 @@ are.
 
 Included currently is:
 
-- `matrixdotorg/sytest:stretch` and `matrixdotorg/sytest:buster`: base
-  containers with SyTest dependencies installed
-- `matrixdotorg/sytest-synapse:py35`: a container which will run SyTest against
-  Synapse on Python 3.5 + Stretch
-- `matrixdotorg/sytest-synapse:py37`: a container which will run SyTest against
-  Synapse on Python 3.7 + Buster
-- `matrixdotorg/sytest-dendrite:go113`: a container which will run SyTest
-  against Dendrite on Go 1.13 + Buster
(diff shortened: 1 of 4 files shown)


richvdh (approved):
lgtm.

we'll probably need to remove the old images from dockerhub to stop them bitrotting and confusing us.

callahad:
> we'll probably need to remove the old images from dockerhub to stop them bitrotting and confusing us.

Filed #1022 to track that

PR #1025, opened 2021-04-15
Test that moving presence off master works

erikjohnston:
Tests https://github.com/matrix-org/synapse/pull/9820

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -89,6 +89,10 @@ sub _init
       event_persister2         => main::alloc_port( "event_persister2[$idx]" ),
       event_persister2_metrics => main::alloc_port( "event_persister2[$idx].metrics" ),
       event_persister2_manhole => main::alloc_port( "event_persister2[$idx].manhole" ),
+
+      writer         => main::alloc_port( "writer[$idx]" ),
+      writer_metrics => main::alloc_port( "writer[$idx].metrics" ),
+      writer_manhole => main::alloc_port( "writer[$idx].manhole" ),
    };
 }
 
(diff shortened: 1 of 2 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -89,6 +89,10 @@ sub _init
       event_persister2         => main::alloc_port( "event_persister2[$idx]" ),
       event_persister2_metrics => main::alloc_port( "event_persister2[$idx].metrics" ),
       event_persister2_manhole => main::alloc_port( "event_persister2[$idx].manhole" ),
+
+      writer         => main::alloc_port( "writer[$idx]" ),
+      writer_metrics => main::alloc_port( "writer[$idx].metrics" ),
+      writer_manhole => main::alloc_port( "writer[$idx].manhole" ),
```
erikjohnston, line 95: I couldn't think of a better name than "writer"

erikjohnston:
Oh I've messed this up by the looks of it.

erikjohnston:
(that should fix it)

lib/SyTest/Homeserver/Synapse.pm
anoadragon453: Similarly unimaginative, but simply `stream_writer`? Or `non_events_stream_writer`? I don't think just seeing "writer" in the logs would tell me much.
<sub>or `write-o-tron 6000`?</sub>

tests/10apidoc/20presence.pl
```diff
@@ -71,23 +71,23 @@ sub matrix_set_presence_status
...
-         my ( $body ) = @_;
-
-         ( $body->{status_msg} // "" ) eq $status_msg or
-            die "Incorrect status_msg";
-
-         Future->done(1);
-      });
+      )->then( sub {
+         # If presence is on a different worker it may take a while for it to
+         # propagate.
```
anoadragon453, line 81: Users always receive their own presence down sync: https://github.com/matrix-org/sytest/blob/0d59907cc6c9de33b8aa37c8dbfd75d9fa9c755c/tests/31sync/05presence.pl#L3

In which case, it might be better to write a `matrix_set_presence_status_synced` method.

...but also: meh, this works.

anoadragon453 (approved):
I think this is good in it's current state, though I left some suggestions for small improvements.

PR #1039, opened 2021-04-25
Add sqlite pusher_database to Homeserver Dendrite.pm

danpe:
Added new Pusher Database as part of https://github.com/matrix-org/dendrite/pull/1842 to resolve https://github.com/matrix-org/dendrite/issues/611

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -234,6 +234,11 @@ sub _get_config
                ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
                "file:$self->{hs_dir}/devices.db" : $db_uri,
          },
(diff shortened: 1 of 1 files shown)


richvdh (approved):
sorry, I'm not sure how this slipped through the net.

It looks sensible to me, though presumably we shouldn't merge it until https://github.com/matrix-org/dendrite/pull/1842 is ready.

danpe:
Thanks @richvdh, true no need to merge yet as we are refactoring the dendrite implementation :)

richvdh:
@neilalexander I see you did a bunch of work on https://github.com/matrix-org/dendrite/pull/1842. Could you have another quick look at this and check it is still relevant?

neilalexander:
@richvdh This is no longer needed since we refactored the push code to use the existing user API database instead, so will close. Thanks for flagging!

Issue #957, opened 2020-09-17
Don't expect optional keys in `/sync` responses


ShadowJonathan:
**Describe the bug**
The [spec for `/sync` responses](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-sync) says that all keys except `next_batch` are optional, yet in many spots in sytest (illustrated by test failures down below) these keys are implicitly expected in the response.

**Additional context**
This is a blocker for https://github.com/matrix-org/synapse/pull/8336.

[Here](https://buildkite.com/matrix-dot-org/synapse/builds/11929) is a buildkite build that details this problem.

ShadowJonathan:
#913 looks relevant to this.

ShadowJonathan:
[Some additional context from Conduit](https://git.koesters.xyz/timo/conduit/issues/29) about failing tests, some of these are possible to be fixed by this.

clokep:
#1040 did some work on this, but from there:

> It would partially fix #957, but I haven't gone into rooms and stripped those keys, so there may still be some dependencies there. Like room specific account data, summary, etc.

PR #1040, opened 2021-05-01
Fix test failures when optional keys in /sync are left out

deepbluev7:
This makes all the sytests pass, if my patched synapse just sends the minimum amount of keys necessary in /sync (apart from 3 email related ones, which is probably my system).

Related: https://github.com/matrix-org/synapse/pull/9919

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>

### DIFF
--- a/tests/30rooms/06invite.pl
+++ b/tests/30rooms/06invite.pl
@@ -159,8 +159,11 @@ sub invited_user_can_reject_invite
 
       # Check that invitee no longer sees the invite
 
-      assert_json_object( $body->{rooms}{invite} );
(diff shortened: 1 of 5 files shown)


ShadowJonathan:
Does this fix https://github.com/matrix-org/sytest/issues/957?

deepbluev7:
It would partially fix #957, but I haven't gone into rooms and stripped those keys, so there may still be some dependencies there. Like room specific account data, summary, etc.

richvdh (approved):
lgtm. Thanks!

PR #1046, opened 2021-05-23
Mention the blacklist file option in the README

### LINKED ISSUE #800: Documentation doesn't mention the blacklist
Now that synapse ships with a blacklist of tests, the sytest readme should point to it

govynnus:
This would close #800


babolivier (approved):
Looks good to me, thanks!

PR #1052, opened 2021-05-27
Script to convert TAP to GHA workflow commands

richvdh:
I plan to wire this into the github actions workflow, to make it easier to see
what broke.

### DIFF
--- a/scripts/tap_to_text.pl
+++ b/scripts/tap_to_text.pl
@@ -1,0 +1,52 @@
+#!/usr/bin/perl
+#
+# Write a textual summary of a TAP file
+
+use strict;
+use warnings FATAL => 'all';
+
+use TAP::Parser;
+
+my $RED = "\e[31m";
+my $RESET_FG = "\e[39m";
+
+my $tap_file = $ARGV[0];
+
+my $parser = TAP::Parser->new( { source => $tap_file } );
+my $in_error = 0;
+my $expected_fail = 0;
+
+while ( my $result = $parser->next ) {
+   if ( $result->is_test ) {
+      # conclude any previous error block
+      if( $in_error ) {
+         print "\n";
+      }
+
+      $in_error = 0;
+
+      if ( not $result->is_ok ) {
+         $in_error = 1;
+
+         my $number = $result->number;
+         my $description = $result->description;
+
+         print "${RED}FAILURE:$RESET_FG #$number: $description\n";
+      } elsif ( $result->directive and not $result->is_actual_ok ) {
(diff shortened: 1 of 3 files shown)


scripts/tap_to_text.pl
```diff
@@ -0,0 +1,52 @@
...
+
+      $in_error = 0;
+
+      if ( not $result->is_ok ) {
+         $in_error = 1;
+
+         my $number = $result->number;
+         my $description = $result->description;
+
+         print "${RED}FAILURE:$RESET_FG #$number: $description\n";
```
callahad, line 34: I wonder if we should look for `GITHUB_ACTIONS` being [set in the environment](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables), and if so, log the error in a [format that it will detect](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message)?

E.g., `print ("::error::#$number: $description\n");`

richvdh: Ahh I looked for a document like that, and failed to find one.

I think it makes sense for this to be GHA-specific, rather that having magic based on the environment. I'll have a go at this.

callahad:
Output looks good to me. Wonder if this should live here or in Synapse where it's easier to iterate on, but 🤷.

Wonder if the grouping is strictly necessary, but it doesn't hurt.

I have a stripped down PR running at matrix-org/synapse#10090 which I'll re-start until it naturally hits a failure just so we can have a "real" example of how this script behaves.

callahad (approved):

anoadragon453:
> Wonder if this should live here or in Synapse where it's easier to iterate on, but shrug.

It can be useful for other homeserver implementations, so should probably live in the Sytest repo.

callahad:
Actually, one requested change: could this return a non-0 exit code when failures are detected?

callahad:
For context, this is what this looks like right now 🎉 

Screen Shot 2021-05-28 at 13 51 57

Screen Shot 2021-05-28 at 13 51 36

richvdh:
> > Wonder if this should live here or in Synapse where it's easier to iterate on, but shrug.
> 
> It can be useful for other homeserver implementations, so should probably live in the Sytest repo.

Yup absolutely this. Likewise, I can't see other test frameworks being silly enough to write TAP output.

richvdh:
> Actually, one requested change: could this return a non-0 exit code when failures are detected?

your wish is my command. If you're happy with this can you mash the merge button?

Issue #1054, opened 2021-05-28
Socket error when starting Synapse - Cannot bind - No such file or directory


MadLittleMods:
When trying to run Sytest against Synapse, I am running into the following error socket bind error, `Error starting server-0: Cannot bind - No such file or directory @details=bind IO::Socket::UNIX=GLOB(0x7feab5d1df20) #sytest-JXOAUKDFIBVQMRPNAJEO.sock No such file or directory at tests/05homeserver.pl line 128.`.

Which probably corresponds to [`lib/SyTest/Homeserver/ProcessManager.pm#L355-L361`](https://github.com/matrix-org/sytest/blob/722b20fa8d0ec8f021784dea24ebe4c362b2d302/lib/SyTest/Homeserver/ProcessManager.pm#L355-L361)

```sh
$ ./run-tests.pl tests/31sync/01filter.pl
Running tests/31sync/01filter.pl...
  Testing if: Can create filter...
Tests: 0 / 2 | OKSignature ok
subject=/CN=localhost
 # Started test HTTPS Server at localhost:56633
 # Started test SMTP Server at localhost:56634
 # Starting Homeserver using SyTest::HomeserverFactory::Synapse=HASH(0x7fdf0680a990)
 # Starting server-0
 # Using DB config from /Users/eric/Documents/github/element/sytest/server-0/database.yaml
 # Clearing SQLite database at :memory:
  Testing if: Can create filter...
Tests: 0 / 2 | OKSignature ok
subject=/CN=localhost
 # Generating config for port 8800
 # Creating config for server 0 with command python -m synapse.app.homeserver --config-path /Users/eric/Documents/github/element/sytest/server-0/config.yaml --generate-config --report-stats=no --server-name localhost:8800
 # [python]: Config file '/Users/eric/Documents/github/element/sytest/server-0/config.yaml' already exists. Generating any missing config files.
 # Starting server 0 for port 8800
  Testing if: Can create filter...
Tests: 0 / 2 | OK
 ** Error starting server-0: Cannot bind - No such file or directory @details=bind IO::Socket::UNIX=GLOB(0x7feab5d1df20) #sytest-JXOAUKDFIBVQMRPNAJEO.sock No such file or directory at tests/05homeserver.pl line 128.
  Testing if: Can create filter... FAIL:
  Testing if: Can download filter... SKIP due to lack of can_create_filter
 # Killing process server
 ** server process failed - code=2
 
1 tests FAILED
```

[...]

PR #1058, opened 2021-06-18
Fix socket binding on macOS

MadLittleMods:
Fix socket binding on macOS. Abstract namespaced sockets are only supported on Linux so moving the socket to a real temporary file.

Fix https://github.com/matrix-org/sytest/issues/1054

## Dev notes

 - https://apple.stackexchange.com/a/157509/164173
 - "abstract socket name, supported only on Linux", https://docs.racket-lang.org/unix-socket/index.html
 
 
```
 # Clean up the socket file after we're done
   $poke_fut->on_done( sub {
      unlink($path) or die "$0: can't unlink $path: $!\n";
   });
```

### DIFF
--- a/lib/SyTest/Homeserver/ProcessManager.pm
+++ b/lib/SyTest/Homeserver/ProcessManager.pm
@@ -317,16 +317,11 @@ sub _await_ready_notification
 
    my $loop = $self->loop;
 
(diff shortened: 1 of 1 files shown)


richvdh (approved):
lgtm

MadLittleMods:
Thanks @richvdh 🦗

PR #1060, opened 2021-06-30
Add support for testing Conduit

kroeckx:
This add initial support for testing Conduit.

I currently know about 3 issues:
- When starting, it shows: `# Starting Homeserver using SyTest::HomeserverFactory::Conduit=HASH(0x55de4cf83420)`. I assume that HASH part is my very limited knowledge of perl.
- After the test is run, the server-0 and server-1 directory still exist. This results in a 2nd run getting additional errors, for instance it attempts the make the same user as the previous run. I'm not sure who is responsible for cleaning up.
- As far as I know, Conduit doesn't have any https server support, but some tests seem to use it.

### DIFF
--- a/lib/SyTest/Homeserver/Conduit.pm
+++ b/lib/SyTest/Homeserver/Conduit.pm
@@ -1,0 +1,98 @@
+package SyTest::Homeserver::Conduit;
+
+use strict;
+use warnings;
+use 5.010;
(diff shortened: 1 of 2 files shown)


kroeckx:
I've been pointed to https://github.com/valkum/sytest_conduit, so it looks like this is not needed.

PR #1061, opened 2021-07-01
Fix 'Outbound federation will ignore a missing event with bad JSON for room version 6'

erikjohnston:
We can't assume that homeservers will process inbound federation events synchronously.

In this particular case we replace the check for errors in `/send` response with a check that if we send a new event that references the expected rejected event we'll see a request for that event come in.

c.f. https://github.com/matrix-org/synapse/issues/10275

### DIFF
--- a/tests/50federation/33room-get-missing-events.pl
+++ b/tests/50federation/33room-get-missing-events.pl
@@ -493,43 +493,16 @@ test "Outbound federation will ignore a missing event with bad JSON for room ver
          $inbound_server->await_request_get_missing_events( $room_id )
          ->then( sub {
             my ( $req ) = @_;
-            my $body = $req->body_from_json;
-
-            log_if_fail "First /get_missing_events body", $body;
-
-            assert_json_keys( $body, qw( earliest_events latest_events limit ));
-            # TODO: min_depth but I have no idea what it does
-
-            assert_json_list( my $earliest = $body->{earliest_events} );
-            @$earliest == 1 or
-               die "Expected a single 'earliest_event' ID";
-            # It is expected that the earliest event is the m.room.member event,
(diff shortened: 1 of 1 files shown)


tests/50federation/33room-get-missing-events.pl
```diff
@@ -418,7 +418,10 @@ sub sytest_user_and_room_fixture {
 # * Add an event with "bad" data into the room history, but don't send it.
 # * Add a "good" event into the room history and send it.
 # * The homeserver attempts to get the missing event (with the bad data).
-# * Ensure that fetching the event results in an error.
+# * The homeserver should reject the "good" event.
+# * To check this we send some another valid event pointing at the "good" event,
```
richvdh, line 422:
```suggestion
# * To check this we send another valid event pointing at the "good" event,
```

tests/50federation/33room-get-missing-events.pl
```diff
@@ -517,21 +536,46 @@ sub sytest_user_and_room_fixture {
...
+                  'latest_events[0]' );
 
-            Future->done;
-         }),
-      );
+               $req->respond_json( {
+                  events => [ $sent_event ],
+               } );
+
+               Future->done;
```
richvdh, line 573: if this is a cut and paste of the earlier code, let's factor it out to a separate `sub` ?

richvdh (changes requested):

richvdh:
do we even still need this test, given it has been ported to complement?

erikjohnston:
> do we even still need this test, given it has been ported to complement?

FTR we decided we *do* want to keep the test here, at least until we can add worker mode to complement and add it to CI

richvdh (approved):
looks generally plausible, modulo CI being borked

PR #1062, opened 2021-07-04
Fix invalid UTF-8 test

kroeckx:


anoadragon453 (approved):
Although not explicitly written under [`/register` in the spec](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0register), according to [the error code definitions](https://spec.matrix.org/unstable/client-server-api/#api-standards), I agree that both `M_BAD_JSON` and `M_NOT_JSON` are valid error codes to return in this case.

anoadragon453:
CI is failing for an unrelated reason.

richvdh:
to be clear, `M_NOT_JSON` means "I couldn't decode your JSON". `M_BAD_JSON` means "I could decode your json, but the parameters therein were bad".

If there's an 0x81 in the JSON, that means that it's not valid JSON. Why is it correct to return `M_BAD_JSON` here?

kroeckx:
> to be clear, `M_NOT_JSON` means "I couldn't decode your JSON". `M_BAD_JSON` means "I could decode your json, but the parameters therein were bad".
> 
> If there's an 0x81 in the JSON, that means that it's not valid JSON. Why is it correct to return `M_BAD_JSON` here?

The documentation does say that M_NOT_JSON would be a valid response
to having the 0x81 in it. But I think it's a mistake in the spec
to distinguish between the 2 cases. The input is invalid, whether
it happens to be valid json or not doesn't really matter. So maybe
that should be an issue against the spec.

richvdh:
> But I think it's a mistake in the spec
to distinguish between the 2 cases. The input is invalid, whether
it happens to be valid json or not doesn't really matter.

well maybe, but as you say that's a matter for the spec, not here. For now, the situation is that the spec gives two different error codes and sytest should check that implementations follow the spec.

anoadragon453:
Ah, sorry I misunderstood the difference between the two. It may be worthwhile for clarity's sake to explicitly state that both `M_BAD_JSON` and `M_NOT_JSON` can never both be acceptable.

> The input is invalid, whether
it happens to be valid json or not doesn't really matter.

The spec states that `M_BAD_JSON` is for valid json, but an incompatible payload for that endpoint. It can be useful to distibguish between the two from the client's perspective.

Either way yes, this is a spec issue and until anything changes there, we should back out this commit.

@kroeckx Out of interest, did you happen to notice this while looking through Sytest's code, or are you implementing a Matrix homeserver and would find this change helpful?

kroeckx:
I was running it against Conduit, which uses Ruma to parse it. I
think Ruma does indicate the specific error, so in theory it would
be possible to try to map some of those to M_BAD_JSON and others to
M_NOT_JSON, but that looks like it will break in the future.

kroeckx:
Note that this PR had 2 commits but where combined into 1 on merge.
I think the first commit that changes the field from test to
username makes sense.

anoadragon453:
> Note that this PR had 2 commits but where combined into 1 on merge.

Thanks for the heads up. I agree, as otherwise the server may potentially return a different error entirely. Best to eliminate all error variables besides the one we're testing for.

Note to readers that `username` is not a required field on `/register` [according to the spec](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0register), but still should be parsed by the homeserver.

I've opened a PR with only the first commit for review at https://github.com/matrix-org/sytest/pull/1075.

PR #1065, opened 2021-07-04
Don't check for presence of `room_alias` in `createRoom` response

kroeckx:
The specification only has a room_id as return value, there is no room_alias.

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -21,7 +21,7 @@ test "POST /createRoom makes a public room",
       )->then( sub {
          my ( $body ) = @_;
 
-         assert_json_keys( $body, qw( room_id room_alias ));
+         assert_json_keys( $body, qw( room_id ));
          assert_json_nonempty_string( $body->{room_id} );
 
          Future->done(1);


kroeckx:
It would be nice that the check was all keys are present, and that no other keys are present.

anoadragon453:
This is an instance of Synapse differing from the spec - as Synapse [will return `room_alias`](https://github.com/matrix-org/synapse/blob/a683028d81606708f686b890c0a44f5a20b54798/synapse/handlers/room.py#L853-L854) if one was requested in the `/createRoom` call.

I suggest we merge this PR to remove the requirement from other servers for now, and opt to add this return field to the spec if we think it's useful?

My assumption for the origin of this field is that it was a convenience for the client, as the client must only send the **localpart** of the alias, but would get back the full alias from the server rather than needing to construct it itself.

richvdh:
this should have been done as part of https://github.com/matrix-org/sytest/pull/880 really. The same thing still applies: this is a step in the right direction, but there are probably other bits of sytest that rely on its presence and will need fixing.

see also https://github.com/matrix-org/matrix-doc/issues/1243 for some context on its non-inclusion in the spec.

richvdh (approved):

PR #1067, opened 2021-07-09
Fix 'Federation rejects inbound events where the prev_events cannot be found'

erikjohnston:
We can't assume that homeservers will process inbound federation events synchronously.

In this particular case we replace the check for errors in /send response with a check that if we send a new event that references the expected rejected event we'll see a request for that event come in.

c.f. matrix-org/synapse#10275


richvdh (approved):
lgtm

PR #1068, opened 2021-07-09
Make implementation_specific accept an array

kegsay:
Since Dendrite now implements some shared secret registration bits, we want to test it.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -645,8 +645,8 @@ sub _push_test
    }
 
    if( exists $params{implementation_specific} ) {
-       if (grep { $_ ne $HS_FACTORY->implementation_name() } $params{implementation_specific}) {
-           return;
+       unless (grep { $_ eq $HS_FACTORY->implementation_name() } @{$params{implementation_specific}}) {
+          return;
        }
    }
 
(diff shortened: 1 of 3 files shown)


richvdh (approved):

neilalexander (approved):

richvdh:
wait, it's failing CI

neilalexander:
Probably because there are other tests marked as Synapse-only with `'synapse'` rather than `['synapse']`?

run-tests.pl
```diff
@@ -645,8 +645,8 @@ sub _push_test
    }
 
    if( exists $params{implementation_specific} ) {
-       if ( $HS_FACTORY->implementation_name() ne $params{implementation_specific} ) {
-           return;
+       unless (grep { $_ eq $HS_FACTORY->implementation_name() } @{$params{implementation_specific}}) {
```
richvdh, line 648: looks like you either need to update *all* the `implementation_specific` declarations, or make this cope with strings as well as lists.

richvdh, line 648: also:

```suggestion
       unless (grep { $_ eq $HS_FACTORY->implementation_name() } @{ $params{implementation_specific} }) {
```

because leo.

kegsay: Bleh, will update them all because like hell I'm modifying this line.

richvdh: hehe

could you update `DEVELOP.rst`, which mentions this, too?

PR #1075, opened 2021-07-12
Use an existing field to test invalid UTF-8

anoadragon453:
This change is the first commit of https://github.com/matrix-org/sytest/pull/1062, which was reverted after being merged via https://github.com/matrix-org/sytest/pull/1066. The invalid commit was the second one in the changeset, but the first commit is potentially useful.

The test attempts to check for an error from the homeserver indicating that the JSON is invalid by including an invalid character under the key `test` in the request. However, since `test` is not a known key a request to [/register](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0register), the homeserver may ignore it entirely.

This change attempts to allow homeservers to pass this test even if they only selectively parse request bodies. Originally written by @kroeckx.

### DIFF
--- a/tests/10apidoc/01request-encoding.pl
+++ b/tests/10apidoc/01request-encoding.pl
@@ -6,7 +6,7 @@ test "POST rejects invalid utf-8 in JSON",
    do => sub {
       my ( $http ) = @_;
 
-      my $reqbody = '{ "test": "a' . chr(0x81) . '" }';
+      my $reqbody = '{ "username": "a' . chr(0x81) . '" }';
 
       $http->do_request(
          method => "POST",


clokep:
I don't see how this matters -- if the payload includes an invalid character than the entire payload is invalid, no?

anoadragon453:
> I don't see how this matters -- if the payload includes an invalid character than the entire payload is invalid, no?

I suppose so - but there may be some mythical JSON parser out there that would happen to ignore the key. 

I guess my response is: why not use a valid key here?

clokep:
> I guess my response is: why not use a valid key here?

Because all keys are "valid". You can put arbitrary info into the payload -- it is JSON afterall.

anoadragon453:
Ah, sorry I'm using confusing terminology.

I meant "why not use a key that the server is expected to read given this endpoint"? Such as "username" or "password".

clokep:
I'm not sure it really matters, but it seems that another test would be needed then since the original PR (#1062) talks about how this test is failing in Ruma/Conduit since it ignores that field, but works in Synapse. I'm not convinced this is a sytest bug, but a Conduit bug. 🤷

kroeckx:
Conduit doesn't fail this test currently, but would fail it with this change.

Conduit makes use of Ruma to do most of the parsing, but also
parses itself when it involves authentication. Both Conduit and
Ruma make use of serde to parse it, but use it differently. The
effect is that in Ruma's case the fields are known and unknown
fields are just skipped. It doesn't try turn something into a string
just to throw it away. But in Conduit's case the whole json is turned
into a flexible structure with everything parsed.

So in the case of an unknown field with invalid UTF-8, Ruma doesn't
return an error, but for some APIs Conduit sees an error instead and
returns M_NOT_JSON. If it's any other JSON parsing error, or
invalid UTF-8 in a known field, Ruma returns an error, and that
error is currently always turned into an M_BAD_JSON.

I think not wasting time trying to parse a field you know you're
not going to use makes perfect sense. So just send a field that
it will parse.

richvdh:
> I think not wasting time trying to parse a field you know you're
not going to use makes perfect sense. So just send a field that
it will parse.

I'm not familiar with JSON parsing in the Rust ecosystem, but I don't understand how you can even get as far as *attempting* to parse that field. The input as it currently stands is *invalid JSON*. Once any JSON parser hits that invalid character, it should stop trying to parse the input and bail out. You can't just ignore invalid input because it happens to be between two `"` characters.

To draw an analogy, suppose you are trying to parse this "JSON": 

```
{
   "unused_field": {
        nonsense_here
   },
   "important_field": "value"
}
```

The content of `unused_field` is invalid, but we only care about `important_field`. Are we correct to ignore `unused_field`? I would argue heavily that would not be correct.

erikjohnston:
> I'm not familiar with JSON parsing in the Rust ecosystem, but I don't understand how you can even get as far as attempting to parse that field. The input as it currently stands is invalid JSON. Once any JSON parser hits that invalid character, it should stop trying to parse the input and bail out. You can't just ignore invalid input because it happens to be between two " characters.

"JSON must be UTF-8" is basically saying that "all JSON strings must be UTF-8 encoded", and so if you're writing a parser then operating purely on bytes to parse the structure of the data is perfectly fine. The difference then between Rust and Python is that Rust doesn't bother parsing the string fields unless they're used, whereas Python does (though if you give the python JSON parser an invalid utf8 string, rather than bytes, it will happily parse it).

You can argue that parsers should still be validating the string data to be UTF-8, but that is slightly less clear cut to me.

(Plus, technically can't you encode JSON in other UTF variants according to the specs if its in a closed system?)

kroeckx:
> I don't understand how you can even get as far as *attempting* to parse that field.

I guess I don't understand how you write parsers either.

richvdh:
I guess it's fair that you can just look at the byte values.

Anyway I think this has gone round the houses enough. It's still a valid test of bad encoding.

richvdh (approved):

PR #1077, opened 2021-07-12
Type-cast origin_server_ts as JSON::number to avoid sending it as a string

kegsay:
For some reason this test was sending origin_server_ts as a string.
Type-casting this makes the test pass on Dendrite.


neilalexander (approved):
Good catch!

PR #1078, opened 2021-07-13
Remove DINUM-specific tests from sytest

babolivier:
These are tests testing a DINUM-specific feature. As per comments in https://github.com/matrix-org/synapse-dinsic/pull/99, this feature will be moved to a Synapse module, and these tests will live in this newly created module.


babolivier:
Required to unblock https://github.com/matrix-org/synapse-dinsic/pull/99

babolivier:
Tests failing are because SyTest is using Synapse's mainline instead of its `dinsic` branch

anoadragon453 (approved):
Yep, these don't really belong in SyTest.

Issue #1064, opened 2021-07-04
m.room.test event missing a txnId


kroeckx:
tests/10apidoc/30room-create.pl tries to send an m.room.test event to wait for the room creation to be complete. But it doesn't send a txnId, which is a required parameter.

PR #1080, opened 2021-07-14
Use a non-deprecated API to send m.room.test

### LINKED ISSUE #878: Sytest uses removed POST /rooms/{roomId}/send/{eventType}
This endpoint was removed [5 years ago](https://github.com/matrix-org/matrix-doc/pull/181) and I'd argue servers shouldn't need support that endpoint just to have unrelated tests succeed.

Conduit ran into this: https://git.koesters.xyz/timo/conduit/issues/35
And it seems like Dendrite did too 3 years ago: https://github.com/matrix-org/dendrite/pull/265

At least the usage to finalize tests should be removed, see here for example: https://github.com/matrix-org/sytest/blob/e14ee8315bd6efbdcda79b0d9412bc5dd8ff61a3/tests/10apidoc/30room-create.pl#L440

The tests for this endpoint should also be made optional imo: https://github.com/matrix-org/sytest/blob/a3396d3766a156debbb287eaa57f58ac4f728d80/tests/10apidoc/34room-messages.pl#L9

kroeckx:
Fixes: #1064

I've tried to use matrix_send_room_message_synced, but I have no perl skills.

This also partially fixes #878

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -424,6 +424,8 @@ The resultant future completes with the room_id.
 
 =cut
 
+my $txn_id = 1000000;
+
 sub matrix_create_room_synced
 {
    my ( $user, %params ) = @_;
@@ -436,11 +438,12 @@ sub matrix_create_room_synced
 
       matrix_do_and_wait_for_sync( $user,
          do => sub {
-            my $uri = "/r0/rooms/$room_id/send/m.room.test";
+            my $uri = "/r0/rooms/$room_id/send/m.room.test/$txn_id";
+            $txn_id++;
 
             do_request_json_for(
                $user,
-               method => "POST",
+               method => "PUT",
                uri    => $uri,
                content => {},
             );


tests/10apidoc/30room-create.pl
```diff
@@ -436,11 +438,12 @@ sub matrix_create_room_synced
 
       matrix_do_and_wait_for_sync( $user,
          do => sub {
-            my $uri = "/r0/rooms/$room_id/send/m.room.test";
+            my $uri = "/r0/rooms/$room_id/send/m.room.test/$txn_id";
+	    $txn_id++;
```
babolivier, line 442: Indentation looks wrong here?

babolivier (changes requested):
Looks good otherwise, thanks!

babolivier (approved):
Thanks! Also in the future, please avoid force-pushing on your branch once someone has started reviewing it, as it makes it more difficult to figure out what's changed since the previous review. We squash-merge so it'll appear as a single commit once merged anyway :slightly_smiling_face:

PR #1082, opened 2021-07-16
Mark groups tests as using deprecated endpoints

neilalexander:
Since we've replaced Groups with Spaces and new implementations aren't likely to implement Groups, we should mark them as `deprecated_endpoints`.


clokep:
Note that [MSC1772](https://github.com/matrix-org/matrix-doc/blob/master/proposals/1772-groups-as-rooms.md#background-and-objectives) includes the following:

> The existing `/r0/groups` API would be deprecated in Synapse and remain unspecified.

So I think this is safe to say that other homeservers shouldn't be implementing this.

kegsay (approved):

PR #1083, opened 2021-07-17
Use correct start and stop to get the message

kroeckx:
The output of /sync without since parameter has the most recent
message events, and so should include at least the message we
just sent.

The tests used prev_batch, but this should point before the message
that was returned in /sync. Starting backward from an event before
the test message should not have received any message.

We now test between next_batch and prev_batch, we should get the
message.

### DIFF
--- a/tests/10apidoc/34room-messages.pl
+++ b/tests/10apidoc/34room-messages.pl
@@ -142,7 +142,8 @@ test "GET /rooms/:room_id/messages returns a message",
          matrix_sync( $user )
       })->then( sub {
          my ( $sync_body ) = @_;
-         my $token = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{prev_batch};
+         my $next_batch = $sync_body->{next_batch};
+         my $prev_batch = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{prev_batch};
 
          do_request_json_for( $user,
             method => "GET",
@@ -151,7 +152,8 @@ test "GET /rooms/:room_id/messages returns a message",
             # With no params this does "forwards from END"; i.e. nothing useful
             params => {
                 dir => "b",
-                from => $token,
+                from => $next_batch,
+                to => $prev_batch,
             },
          )
       })->then( sub {
@@ -180,7 +182,8 @@ test "GET /rooms/:room_id/messages lazy loads members correctly",
          matrix_sync( $user )
       })->then( sub {
          my ( $sync_body ) = @_;
-         my $token = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{prev_batch};
+         my $next_batch = $sync_body->{next_batch};
+         my $prev_batch = $sync_body->{rooms}->{join}->{$room_id}->{timeline}->{prev_batch};
 
          do_request_json_for( $user,
             method => "GET",
@@ -189,7 +192,8 @@ test "GET /rooms/:room_id/messages lazy loads members correctly",
             params => {
                dir => "b",
                filter => '{ "lazy_load_members" : true }',
-               from => $token,
+               from => $next_batch,
+               to => $prev_batch,
             },
          )
       })->then( sub {


kroeckx:
I believe that the Synapse is not following the specification and so the test failure is expected. I'm unsure why Dendrite passes with both the old and new test, I suspect there is a different bug in Dendrite.

tests/10apidoc/34room-messages.pl
```diff
@@ -151,7 +152,8 @@ sub matrix_send_room_text_message
             # With no params this does "forwards from END"; i.e. nothing useful
             params => {
                 dir => "b",
-                from => $token,
+                from => $next_batch,
```
richvdh, line 155: the spec could be clearer about this, but it's not intended that you can pass a `next_batch` token in `from`. Per https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-rooms-roomid-messages:

> `from`: ... this token can be obtained from a `prev_batch` token returned for each room by the sync API, or from a `start` or `end` token returned by a previous request to this endpoint.

I think that to do what this test wants, we need to do a *second* sync, and then paginate backwards from the `prev_batch` of the second sync. Which, incidentally, will also work around the synapse bug.

richvdh (changes requested):

kroeckx:
I think what you're saying is that Sytest should act more like a
client should act. So what I would expect is:
/sync
/send/m.room.message
/sync&since=next_batch

And get the message from the 2nd /sync call.

But maybe the intention of this test is to use the /messages API,
and see if that works. From your comment, I understand that if we
add a since to the 2nd sync call, calling the messages API with
a from set to the 2nd sync's next_batch should work around
Synapse's bug, but would also not really follow the spec.

My understanding is that a client will use that API
in case the /sync didn't return all the messages. So I think to
properly test that, we need to to send enough messages that they
don't fit in a /sync. I'm not sure if there is a test like that.

I'm currently also not sure what other tests expect this test to
test.

I think your comment about using the prev_batch from the 2nd sync
is just the same as the current test, it would still ask for
messages before the test message.

richvdh:
> I think what you're saying is that Sytest should act more like a client should act. So what I would expect is: /sync /send/m.room.message /sync&since=next_batch And get the message from the 2nd /sync call.

No, this is a test of the `/message` API.

I would expect:

* `/send`
* `/sync`  <- results include test message
* `/sync?since=next_batch`  <- empty result
* `/messages?from=prev_batch` <- includes test message

> I think your comment about using the prev_batch from the 2nd sync
is just the same as the current test, it would still ask for
messages before the test message.

No, I don't think so. If the test message is in the 1st sync, then the 2nd sync will be empty. The `prev_batch` from the 2nd sync will be *after* the test message.

kroeckx:
> I would expect:
> 
> * `/send`
> * `/sync`  <- results include test message
> * `/sync?since=next_batch`  <- empty result
> * `/messages?from=prev_batch` <- includes test message

That 2nd sync doesn't return a prev_batch since it's an empty
result. It just contains a next_batch.

richvdh:
> That 2nd sync doesn't return a prev_batch since it's an empty result. It just contains a next_batch.

Ugh, I see, yes. Perhaps you'll have to send another message to force an entry for that room in the second /sync result.

kroeckx:
I've changed things and updated the commit message.

kroeckx:
The test suite failure does not look related to this PR

tests/10apidoc/34room-messages.pl
```diff
@@ -138,12 +144,18 @@ sub matrix_send_room_text_message
       my ( $user, $room_id ) = @_;
 
       matrix_send_room_text_message( $user, $room_id,
-         body => "Here is the message content",
+         body => "Here is the first message",
       )->then( sub {
-         matrix_sync( $user )
+         matrix_sync( $user );
+      })->then( sub {
+	 matrix_send_room_text_message( $user, $room_id,
```
richvdh, line 151:
```suggestion
         matrix_send_room_text_message( $user, $room_id,
```

tests/10apidoc/34room-messages.pl
```diff
@@ -176,12 +192,18 @@ sub matrix_send_room_text_message
...
```
richvdh, line 199:
```suggestion
         matrix_send_room_text_message( $user, $room_id,
```

richvdh (changes requested):
the change looks generally fine now. Please could you update the summary and description in the PR?

PR #1084, opened 2021-07-18
Mark POST /rooms/:room_id/send/:event_type as deprecated

kroeckx:
Fixes: #878

The other issue is fixed in #1080


babolivier (approved):
Thanks!

PR #1085, opened 2021-07-21
Fix `/aliases` test to use stable endpoint

neilalexander:
The test was previously referring to an unstable endpoint, but it's listed in the latest spec: https://matrix.org/docs/spec/client_server/r0.6.1#id284


clokep (approved):
Seems reasonable to me.

PR #1086, opened 2021-07-21
Don't check for home_server when registering

kroeckx:
home_server is deprecated


richvdh (changes requested):
@kroeckx it looks like the main reason this and your other PRs, are failing CI, is that you've branched from an old version of Sytest which fails on Synapse 1.37.1 and later.

Please could you merge develop into your branches?

kroeckx:
I think I've rebased them all

richvdh (approved):
lgtm

PR #1087, opened 2021-07-21
Don't check usernames starting with an underscore

kroeckx:
The grammar allows such usernames. Even if it wasn't allowed
the server can return a different username. But some servers
might reserve the use of them.


richvdh:

richvdh (approved):

PR #1088, opened 2021-07-21
Fix usage of typing API

kroeckx:
- Timeout is a required parameter when starting typing.
- The typing parameter takes a bool not an integer.


richvdh:

kroeckx:
All the test suite errors look like bugs in Synapse and Dendrite.

tests/10apidoc/35room-typing.pl
```diff
@@ -19,3 +24,50 @@
          Future->done(1);
       });
    };
+
+test "PUT /rooms/:room_id/typing/:user_id without timeout fails",
```
richvdh, line 28: what makes you think `timeout` is mandatory? https://spec.matrix.org/unstable/client-server-api/#put_matrixclientr0roomsroomidtypinguserid doesn't mark it as required.

kroeckx:
> what makes you think `timeout` is mandatory? https://spec.matrix.org/unstable/client-server-api/#put_matrixclientr0roomsroomidtypinguserid doesn't mark it as required.

My interpretation is that it's required when typing is true, but not
when it's false.

tests/10apidoc/35room-typing.pl
richvdh: re https://github.com/matrix-org/sytest/pull/1088#issuecomment-888461770: I don't agree with your interpretation.

kroeckx: The reason I think it's required when typing is true are:
- It says: `If false, the timeout key can be omitted.`, which to me implies that it needs to be present when it's true.
- It says `This tells the server that the user is typing for the next N milliseconds where N is the value specified in the timeout key`. That implies to me that we need to have a value.

Note that I'm not the only one interpreting it that way. If you think this is wrong, please fix the specifications. Adding a default value on the server side is easy enough.

kroeckx: So to clarify it more, those are the options:
typing | timeout
-- | --
true | present
true | not present
false | present
false | not present

The **can** be omitted to me means that both the 3rd and 4th option are valid. Without the sentence, option 3 and 4 are also valid. But since the sentence is there, to me it says that if it's true you **can not** omit it.

richvdh:
blocked by https://github.com/matrix-org/synapse/issues/10497

PR #1092, opened 2021-07-21
Create the certificate with a subjectAltName

kroeckx:

### DIFF
--- a/lib/SyTest/SSL.pm
+++ b/lib/SyTest/SSL.pm
@@ -61,7 +61,7 @@ sub create_ssl_cert
 
    # Create extension file
    my $ext_file = "$cert_file.ext";
-   open(my $fh, '>', $ext_file) or die "Could not open file '$filename' $!";
+   open(my $fh, '>', $ext_file) or die "Could not open file '$ext_file': $!";
    print $fh "subjectAltName=DNS:$server_name\n";
    close $fh;
 
@@ -72,4 +72,3 @@ sub create_ssl_cert
       "-out", $cert_file, "-extfile", $ext_file,
    ) == 0 or die "openssl x509 failed $?";
 }
-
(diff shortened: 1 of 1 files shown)


richvdh:

clokep:
Curious -- what's the reasoning for this?

kroeckx:
> Curious -- what's the reasoning for this?

The subjectAltName is a mandatory field for such a certificate,
and there are implementations that rejects certificates without
it.

In this case, Conduit is using rust webpki when acting as client
to validate the certificate, and it rejects the certificate as
invalid.

lib/SyTest/SSL.pm
```diff
@@ -59,11 +59,17 @@ sub create_ssl_cert
       "-subj", "/CN=$server_name",
    ) == 0 or die "openssl req failed $?";
 
+   # Create extenion file
```
richvdh, line 62:
```suggestion
   # Create extension file
```

lib/SyTest/SSL.pm
```diff
@@ -59,11 +59,17 @@ sub create_ssl_cert
...
+   my $ext_file = "$cert_file.ext";
+   open(my $fh, '>', $ext_file) or die "Could not open file '$filename' $!";
+   print $fh "subjectAltName=\@alts\n[alts]\nDNS=$server_name\n";
```
richvdh, line 65: can we not just write this:

```suggestion
   print $fh "subjectAltName=DNS:$server_name\n";
```

richvdh (commented):
seems sensible. question below

lib/SyTest/SSL.pm
```diff
@@ -59,11 +59,17 @@ sub create_ssl_cert
       "-subj", "/CN=$server_name",
    ) == 0 or die "openssl req failed $?";
 
+   # Create extension file
+   my $ext_file = "$cert_file.ext";
+   open(my $fh, '>', $ext_file) or die "Could not open file '$filename' $!";
```
richvdh, line 64:
```suggestion
   open(my $fh, '>', $ext_file) or die "Could not open file '$ext_file' $!";
```

richvdh (approved):
thanks!

PR #1093, opened 2021-07-21
Allow a JSON null value to indicate the value is not present

kroeckx:


richvdh:

PR #1094, opened 2021-07-21
If authentication failed, no steps completed

kroeckx:


richvdh:

kroeckx:
It seems Synapse claims `m.login.dummy` completed, which is not in the flows.

tests/14account/02deactivate.pl
```diff
@@ -43,13 +43,15 @@ sub matrix_deactivate_account
...
-         assert_json_keys( $body, qw( error errcode params completed flows ));
+         assert_json_keys( $body, qw( error errcode params flows ));
 
          my $errcode = $body->{errcode};
 
          $errcode eq "M_FORBIDDEN" or
             die "Expected errcode to be M_FORBIDDEN but was $errcode";
 
+         exists $body->{completed} and die "Got an unexpected 'completed' key";
+
```
richvdh, line 54: I'm not entirely convinced that the presence of `completed` should be a failure here.

I agree that `completed` should not be *required*.

richvdh (changes requested):

tests/14account/02deactivate.pl
kroeckx: It seems the test I did against synapse was something else. It seems synapse returns: `"completed":[]`, which makes sense.

I guess this is an other case where the specification isn't really clear.

PR #1099, opened 2021-07-28
Put media into a hopefully writable location

neilalexander:
It looks like the sytest pipeline mounts `".:/sytest:ro"` which the dendrite pipeline doesn't, and the relative media path was meaning we were hitting errors during CI builds:

`Failed to create temp dir: Failed to create base temp dir: mkdir /sytest/media_store: read-only file system`


richvdh (approved):
looks reasonable

neilalexander:
The couple of sytest account failures are related to e1e6064cf347595b5cf04e763e017943ddbe4db4 so will look into those tomorrow, but will merge this for now since it solves the original problem with the media path.

PR #1101, opened 2021-07-31
Add support for a presence db in Dendrite

S7evinK:
This adds a presence database for Dendrite.
Also deprecates `/r0/presence/list/:user_id`, as it seems to be removed in the latest CS spec.


kegsay (approved):

kegsay:
We don't yet have a presence DB in master though.

kegsay:
We plan to remove manual configs in Sytest in the near future, so new features can directly add their new config options to the [generate-config](https://github.com/matrix-org/dendrite/blob/master/cmd/generate-config/main.go) script. For now though, this PR is fine, it's just a heads up that this will change in the future.

Issue #1063, opened 2021-07-04
DELETE /device/{deviceId} test doesn't send all required parameters


kroeckx:
The tests/10apidoc/12device_management.pl test tries to log in using a password, but doesn't send all the required parameters.

According to the documentation, it should send:
```
{
  "type": "m.login.password",
  "identifier": {
    ...
  },
  "password": "<password>",
  "session": "<session ID>"
}
```

But what it sends is:
```
{"auth":{"user":"@anon-20210704_143726-8:localhost:8800","password":"cashewnuts","type":"m.login.password"}}
```

So it's seems to be missing an identifier and a session, and instead sends a user.

Conduit rightfully sends a 400 Bad Request, at which point Sytest says the test failed.

richvdh:
similar to #1091, though a different test case I think

PR #1102, opened 2021-07-31
Use identifier when doing authentication

kroeckx:
Fixes: #1063


kroeckx:
I think I have them all now.

richvdh (approved):
lgtm, thanks!

PR #1103, opened 2021-07-31
Use v2 send_join in tests

kroeckx:
The v1 is still used in other tests


richvdh (approved):
lgtm, thanks!

PR #1106, opened 2021-08-04
Try and fix flakey device list update after leave

### LINKED ISSUE #1036: `If remote user leaves room we no longer receive device updates` is flakey
See https://github.com/matrix-org/synapse/pull/9855/checks?check_run_id=2412069509

erikjohnston:
Hopefully fixes #1036, but I was unable to reproduce it locally, so :shrug: 

The only race I could think of here is that the creator's HS receives the device list update for the left user while its still processing the leave of that user, but that seems fairly unlikely.

### DIFF
--- a/tests/41end-to-end-keys/06-device-lists.pl
+++ b/tests/41end-to-end-keys/06-device-lists.pl
@@ -397,9 +397,27 @@ test "If remote user leaves room we no longer receive device updates",
       })->then( sub {
 
          # now one of the remote users leaves the room...
-         matrix_leave_room_synced( $remote_leaver, $room_id );
+         matrix_leave_room( $remote_leaver, $room_id );
       })->then( sub {
          log_if_fail "Remote_leaver " . $remote_leaver->user_id . " left room";
+
+         # wait for the leave to propagate to the creators homeserver
+         await_sync_timeline_contains( $creator, $room_id, check => sub {
+            my ( $event ) = @_;
+
+            assert_json_keys( $event, qw( type content sender ));
+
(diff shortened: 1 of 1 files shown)


tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -397,10 +397,28 @@ sub sync_until_user_in_device_list
...
+         # wait for the leave to propagate to the creators homeserver
+         await_sync_timeline_contains( $creator, $room_id, check => sub {
+            my ( $event ) = @_;
+
+            assert_json_keys( $event, qw( type content sender ));
+
+            return unless $event->{type} eq "m.room.member";
+            return unless $event->{sender} eq $remote_leaver->user_id;
+
+
```
richvdh, line 413:
```suggestion
```

tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -397,10 +397,28 @@ sub sync_until_user_in_device_list
...
+            assert_json_keys( my $content = $event->{content}, qw( membership ));
+
+            return unless $content->{membership} eq "leave";
+
+            return 1;
+         });
+      })->then( sub {
+
```
richvdh, line 421:
```suggestion
```

richvdh (approved):
Don't the logs give any clues as to the problem?

Seems ok, anyway.

erikjohnston:
> Don't the logs give any clues as to the problem?

Not that I saw, alas, though I might easily have missed something given the sheer number of workers

PR #1107, opened 2021-08-04
Don't send garbage to homeserver in end-to-end-keys tests

neilalexander:
This updates the `end-to-end-keys` tests to send real base64 and things that look actually like key IDs.


kegsay (approved):
💯

PR #1111, opened 2021-08-07
Use a valid formatted key to test upload of a key

kroeckx:
The algorithms, keys and signatures parameters are required


tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -35,6 +47,35 @@
...
+   proves => [qw( can_upload_e2e_keys )],
+
+   do => sub {
+      my ( $user ) = @_;
+
+      do_request_json_for( $user,
+         method  => "POST",
+         uri     => "/r0/keys/upload",
+         content => {
+            device_keys => {
```
neilalexander, line 62: Can you please add an inline comment here to state that the reason that this is invalid is because it is missing required fields, for the next person looking at this?

neilalexander (approved):

kegsay (approved):
Can we merge this?

richvdh:
well, hence https://github.com/matrix-org/synapse/issues/10554 I guess. We could add the new test to the synapse blacklist :shrug:

kegsay:
That feels like a reasonable thing to do, given we do the same on the dendrite side for edge cases like this.

erikjohnston:
Synapse PR to add this to the blacklist: https://github.com/matrix-org/synapse/pull/11270

erikjohnston:
Synapse should now have blacklisted this test, so if we merge in `develop` CI should pass

erikjohnston:
Is the dendrite faulures expected?

erikjohnston:
> I think we have seen `Inbound federation can query room alias directory` failing on the Dendrite side sometimes, so I wouldn't worry too much about that.

It'd be good to try and hunt that down at some point, failing CI is sad, etc.

erikjohnston (approved):

Issue #1071, opened 2021-07-10
"GET /directory/room/:room_alias yields room ID" doesn't actaully send the room alias


kroeckx:
The test "GET /directory/room/:room_alias yields room ID" has this code:
```
         uri    => "/r0/directory/room/$room_alias",
```

But $room_alias seems to be empty. The request that is sends has: `GET /_matrix/client/r0/directory/room/?access_token=...`

kroeckx:
The log contains:
```
** Use of uninitialized value $room_alias in concatenation (.) or string at tests/10apidoc/31room-state.pl line 211
```

`POST /join/:room_alias can join a room` fails with a related error:
```
** Use of uninitialized value $room_alias in concatenation (.) or string at tests/10apidoc/33room-members.pl line 120
```

`POST /join/:room_alias can join a room with custom content` with:
```
** Use of uninitialized value $room_alias in concatenation (.) or string at tests/10apidoc/33room-members.pl line 235
```
`Can register using an email address` with:
```
** Use of uninitialized value $name in concatenation (.) or string at lib/SyTest/Assertions.pm line 64
```

`Remote users may not join unfederated rooms` with:
```
** Use of uninitialized value $room in concatenation (.) or string at tests/10apidoc/33room-members.pl line 82
```

Just after the `m.room.canonical_alias accepts present aliases` there is:
```
** Use of uninitialized value $room_alias in substitution (s///) at tests/30rooms/05aliases.pl line 94
```

`Guest users can join guest_access rooms` has:
```
** Use of uninitialized value $user in concatenation (.) or string at tests/10apidoc/31room-state.pl line 476
```

Around `Backfill works correctly with history visibility set to joined` there is:
```
** Use of uninitialized value $room in concatenation (.) or string at tests/10apidoc/33room-members.pl line 82
```

`Can recv device messages until they are acknowledged` and `Device messages with the same txn_id are deduplicated` and 2 others have:
```
** Use of uninitialized value $name in concatenation (.) or string at lib/SyTest/Assertions.pm line 95
```

PR #1112, opened 2021-08-08
Make matrix_create_room return the alias when the server doesn't supp…

kroeckx:
…ort it

The reply is non-standard. Check that it's correct in case it's
returned. Return the requested alias.

Fixes: #1071

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -287,8 +287,7 @@ The following options have defaults:
 commandline.
 
 The resultant future completes with two values: the room_id from the
-/createRoom response; the room_alias from the /createRoom response (which is
-non-standard and its use is deprecated).
+/createRoom response; the room_alias.
 
 =cut
 
@@ -313,7 +312,14 @@ sub matrix_create_room
    )->then( sub {
       my ( $body ) = @_;
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      # The /createRoom response can contain a non-standard room
+      # alias reply, which use has been deprecated.
+      if (defined $body->{room_alias}) {
+         assert_eq( $body->{room_alias}, $opts{rom_alias_name},
+            "room alias" );
+      }
+
+      Future->done( $body->{room_id}, $opts{room_alias_name} );
    });
 }
 


tests/10apidoc/30room-create.pl
```diff
@@ -313,7 +312,14 @@ sub matrix_create_room
    )->then( sub {
       my ( $body ) = @_;
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      # The /createRoom response can contain a non-standard room
+      # alias reply, which use has been deprecated.
+      if (defined $body->{room_alias}) {
+         assert_eq( $body->{room_alias}, $opts{rom_alias_name},
```
richvdh, line 318:
```suggestion
         assert_eq( $body->{room_alias}, $opts{room_alias_name},
```

tests/10apidoc/30room-create.pl
```diff
@@ -313,7 +312,14 @@ sub matrix_create_room
    )->then( sub {
       my ( $body ) = @_;
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      # The /createRoom response can contain a non-standard room
+      # alias reply, which use has been deprecated.
```
richvdh, line 316:
```suggestion
      # alias reply, which has been deprecated.
      # If it is returned, however, let's check it is correct.
```

tests/10apidoc/30room-create.pl
```diff
@@ -287,8 +287,7 @@ =head2 matrix_create_room
 commandline.
 
 The resultant future completes with two values: the room_id from the
-/createRoom response; the room_alias from the /createRoom response (which is
-non-standard and its use is deprecated).
+/createRoom response; the room_alias.
```
richvdh, line 290: please document the behaviour if no `room_alias_name` was given in `%opts`.

tests/10apidoc/30room-create.pl
```diff
@@ -313,7 +312,14 @@ sub matrix_create_room
...
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      # The /createRoom response can contain a non-standard room
+      # alias reply, which use has been deprecated.
+      if (defined $body->{room_alias}) {
+         assert_eq( $body->{room_alias}, $opts{rom_alias_name},
+            "room alias" );
+      }
+
+      Future->done( $body->{room_id}, $opts{room_alias_name} );
```
richvdh, line 322: This doesn't look quite right. `room_alias_name` is just the localpart of the alias (see https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-createroom), whereas the `room_alias` returned by synapse (and that expected by the rest of sytest is the *complete* alias, including `#` and the domain part.

richvdh (changes requested):
Thanks for the contribution! I think it's a sound idea, but needs a couple of tweaks.

richvdh:
closing, since it seems to have been abandoned.

PR #1115, opened 2021-08-11
Move from Buildkite to GitHub Actions

DMRobertson:
Ended up with a fairly direct translation from Buildkite to GHA. I tried to make more use of actions and so on, but generally found this to be more trouble that it's worth.

* https://github.com/DMRobertson/sytest/actions/runs/1124068606 is an example where I'd screwed up the blacklist for synapse with workers. But that shows how test failures will appear (annotations, with a link through to logs)
* https://github.com/DMRobertson/sytest/actions/workflows/pipeline.yml has the complete history
* https://github.com/DMRobertson/sytest/actions/runs/1124143216 is an example of a happy run---except I don't understand why the failing dendrite workflow was unhappy.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -16,47 +16,38 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Synapse - :python: 3.6 / SQLite / Monolith"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: bionic
 
-          - name: "Synapse - :python: 3.6 / :postgres: 10 / Monolith"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: bionic
             postgres: postgres
 
-          - name: "Synapse - :python: 3.6 / :postgres: 10 / Workers"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: bionic
             postgres: postgres
             synapse-workers: workers
 
-          - name: "Synapse - :python: 3.9 / :postgres: 13 / Monolith"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: testing
             postgres: postgres
 
-          - name: "Synapse - :python: 3.9 / :postgres: 13 / Workers"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: testing
             postgres: postgres
             synapse-workers: workers
 
-          - name: "Dendrite - :postgres:"
-            homeserver-implementation: dendrite
+          - homeserver-implementation: dendrite
             postgres: postgres
 
-          - name: "Dendrite - :postgres: / full HTTP APIs"
-            homeserver-implementation: dendrite
+          - homeserver-implementation: dendrite
             postgres: postgres
-            denrite-full-http: yes
+            denrite-full-http: full-http
 
-          - name: "Dendrite - :sqlite:"
-            homeserver-implementation: dendrite
+          - homeserver-implementation: dendrite
 
-          - name: "Dendrite - :sqlite: / full HTTP APIs"
-            homeserver-implementation: dendrite
-            denrite-full-http: yes
+          - homeserver-implementation: dendrite
+            denrite-full-http: full-http
 
     container:
       image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}:${{ matrix.sytest-tag }}
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,7 +1,7 @@
 name: Run sytest
 on:
   push:
-    branches: ["develop", "release-*"]
+    branches: ["develop", "release-*", "github-actions"]
   pull_request:
 
 concurrency:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -50,11 +50,12 @@ jobs:
             denrite-full-http: full-http
 
     container:
-      image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}:${{ matrix.sytest-tag }}
+      # The colon is icky. Better way? Could just include the image explicitly?
+      image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}${{ matrix.sytest-tag && ":" }}${{ matrix.sytest-tag }}
       volumes:
         # TODO PWD mounted as /sytest was originally mounted as readonly. Could pass -v directly?
         - ${{ github.workspace }}:/sytest
-        - ./logs:/logs
+        - ${{ github.workspace }}/logs:/logs
       env:
         POSTGRES: ${{ matrix.postgres && 1 }}
         WORKERS: ${{ matrix.synapse-workers && 1 }}
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -51,7 +51,7 @@ jobs:
 
     container:
       # The colon is icky. Better way? Could just include the image explicitly?
-      image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}${{ matrix.sytest-tag && ":" }}${{ matrix.sytest-tag }}
+      image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}${{ matrix.sytest-tag && ':' }}${{ matrix.sytest-tag }}
       volumes:
         # TODO PWD mounted as /sytest was originally mounted as readonly. Could pass -v directly?
         - ${{ github.workspace }}:/sytest
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,0 +1,32 @@
+name: Run sytest
+on:
+  push:
+    branches: ["develop", "release-*"]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  sytest:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - sytest-tag: bionic
+
+    container:
+      image: matrixdotorg/sytest-${{ matrix.sytest-tag }}
+      volumes:
+        - ${{ github.workspace }}:/src
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: TODO Prepare test blacklist
+      - name: Run bootstrap script
+        run: bash /bootstrap.sh synapse
+      - name: TODO upload artifacts
+      - name: TODO annotate logs
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -17,10 +17,10 @@ jobs:
       matrix:
         include:
           - homeserver-implementation: synapse
-          - sytest-tag: bionic
+            sytest-tag: bionic
 
     container:
-      image: matrixdotorg/sytest-${{ matrix.sytest-tag }}
+      image: matrixdotorg/sytest-${{ matrix.homeserver-implementation }}:${{ matrix.sytest-tag }}
       volumes:
         - ${{ github.workspace }}:/src
 
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -66,7 +66,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Run sytest
         working-directory: /sytest
-        run: bash /bootstrap.sh ${{ matrix.homeserver-implementation }}
+        run: bash -xe /bootstrap.sh ${{ matrix.homeserver-implementation }}
       - name: Summarise results.tap
         if: ${{ always() }}
         run:  /sytest/scripts/tap_to_gha.pl /logs/results.tap
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -16,24 +16,65 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - homeserver-implementation: synapse
+          - name: "Synapse - :python: 3.6 / SQLite / Monolith"
+            homeserver-implementation: synapse
             sytest-tag: bionic
+
+          - name: "Synapse - :python: 3.6 / :postgres: 10 / Monolith"
+            homeserver-implementation: synapse
+            sytest-tag: bionic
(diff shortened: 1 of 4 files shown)


DMRobertson:
Suggest we keep using BuildKite for a bit and see how this fares---then can turn off the webhook and remove the config from [matrix-org/pipelines](https://github.com/matrix-org/pipelines)

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
...
+
+      - name: Prepare blacklist file
+        if: ${{ matrix.workers }}
+        run: cat /src/sytest-blacklist /src/.ci/worker-blacklist > /src/synapse-blacklist-with-workers
+
+      - name: Run sytest
+        run: bash -xe /bootstrap.sh synapse
+
+      - name: Summarise results.tap
+        if: ${{ always() }}
```
anoadragon453, line 70: May also be worth adding a comment for what `if: ${{ always() }}` means - as it *appears* a paradox. But really it just means to run even if the previous test fails.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
...
+    container:
+      image: matrixdotorg/sytest-synapse:${{ matrix.sytest-tag }}
+      volumes:
+        # bootstrap.sh expects the sytest source available at /sytest.
+        # TODO Buildkite mounted /sytest as readonly. Can we do this on GHA? Do we need it?
+        - ${{ github.workspace }}/sytest:/sytest
+        # synapse_sytest.sh expects a synapse checkout at /src
+        - ${{ github.workspace }}/synapse:/src
+      env:
+        POSTGRES: ${{ matrix.postgres && 1 }}
```
anoadragon453, line 44: Is `&& 1` a trick to say that this should be enabled by default?

Why don't we do the same for dendrite's runs?

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
...
+        POSTGRES: ${{ matrix.db == 'postgres' }}
+        API: ${{ matrix.api == 'full-http' }}
+
+    steps:
+      - name: Checkout sytest
+        uses: actions/checkout@v2
+        with:
+          path: sytest
+
+      - name: Fetch corresponding synapse branch
```
anoadragon453, line 109:
```suggestion
      - name: Fetch corresponding dendrite branch
```

scripts/dendrite_sytest.sh
```diff
@@ -70,7 +70,7 @@ rsync -r --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /
 
 if [ $TEST_STATUS -ne 0 ]; then
     # Build the annotation
-    perl /sytest/scripts/format_tap.pl /logs/results.tap "$BUILDKITE_LABEL" >/logs/annotate.md
+    perl /sytest/scripts/format_tap.pl /logs/results.tap "$GITHUB_SHA" >/logs/annotate.md
```
anoadragon453, line 73: Where does this actually show up? If I understand correctly, this is going from:

```
### TAP output for :rocket: Synapse :: postgres :: workers
```

to 

```
### TAP output for 1b557303bd3e0d8d53a1e1764116e70c266db1bf
```

which seems less useful, but I also don't see any SHA's appearing in the failed or successful output of any of the linked check runs.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
...
+
+jobs:
+  synapse:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - sytest-tag: bionic
```
anoadragon453, line 19: I don't suppose this is easy, but is it possible to additionally include the Python version in each job name?

image

anoadragon453 (changes requested):
I can't say I know why the Dendrite run is appearing to fail. I don't see where the tap formatting script would return something other than 0 - is that failure flaky or deterministic?

Overall this is looking great. The linked output is very clean, and the resulting pipeline file doubly so!

Most of my comments are just clarifications.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
+name: Run sytest
+on:
+  push:
+    branches: ["develop", "release-*"]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
DMRobertson: In truth I just nicked this from synapse's tests.yml, but yeah---it behaves as you describe. Happy to add a comment though.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,130 @@
...
+
+          - sytest-tag: testing
+            postgres: postgres
+            workers: workers
+
+    container:
+      image: matrixdotorg/sytest-synapse:${{ matrix.sytest-tag }}
+      volumes:
+        # bootstrap.sh expects the sytest source available at /sytest.
+        # TODO Buildkite mounted /sytest as readonly. Can we do this on GHA? Do we need it?
```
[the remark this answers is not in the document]
DMRobertson: That will `Mount the container's root filesystem as read only`; I suspect that's not going to propagate to the mounts/binds. If we cared, I think we'd have to pass a `-v` invocation to options.

Originally done by @richvdh in https://github.com/matrix-org/pipelines/pull/16. I guess we might have used the pipelines locally too?

.github/workflows/pipeline.yml
DMRobertson: There's also [`continue-on-error`}(https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) which will allow a step to fail and continue onwards. But I think doing so will make it that step always show as a successful green tick, even if it failed.

.github/workflows/pipeline.yml
DMRobertson: Ahh, I inherited this from synapse's tests.yml too, [see here](https://github.com/matrix-org/synapse/blob/c12b5577f22ee587b60ad7b65e88322ce1d86b7b/.github/workflows/tests.yml#L195-L200).

I think we want the parameters in the `matrix` to have wordy values so they appear nicely in the workflow's output. But some of the BuildKite config used to use config like

```yaml
    env:
      POSTGRES: "1"
      WORKERS: "1"
      BLACKLIST: "synapse-blacklist-with-workers"
```
So I think `&& 1` was a way to ensure we provided the same environment as we were using on buildkite. With that said, synapse_sytest.sh only ever checks `[ -n "$POSTGRES" ]`, i.e. that $POSTGRES is a nonempty string. So maybe `&& 1` wasn't necessary.

Aside: not fully sure how GHA's expressions work, but [there's something written here](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#operators). I think `a && b` evaluates to `b` if both `a` and `b` are truthy, and `false` otherwise.

> Why don't we do the same for dendrite's runs?

Nothing stopping us.  But I realised that dendrite's test cases break down in to two orthogonal factors, whereas
synapse's don't decompose so nicely (`workers` requires `postgres`, I think). This meant that I could make the `matrix` config simpler: see dendrite's lines 88--89 versus synapse's 18--33.

Hmm. With that said, it looks to me that `dendrite_sytest.sh` only ever uses postgres. Maybe the BuildKite config I've been adapting was mistaken?

scripts/synapse_sytest.sh
```diff
@@ -203,7 +203,7 @@ rsync --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /log
 
 if [ $TEST_STATUS -ne 0 ]; then
     # Build the annotation
-    perl /sytest/scripts/format_tap.pl /logs/results.tap "$BUILDKITE_LABEL" >/logs/annotate.md
+    perl /sytest/scripts/format_tap.pl /logs/results.tap "$GITHUB_SHA" >/logs/annotate.md
```
DMRobertson, line 206: Also this.

scripts/dendrite_sytest.sh
DMRobertson: This puzzled me for a while. I think this is buildkite specific. We used to use https://github.com/matrix-org/annotate-buildkite-plugin to upload annotations.md. I think its functionality is replaced by the `Summarise results.tap` step.

.github/workflows/pipeline.yml
DMRobertson: Wasn't too bad: see https://github.com/DMRobertson/sytest/actions/runs/1127951256

Screenshot from 2021-08-13 15-16-40

.github/workflows/pipeline.yml
DMRobertson: > Hmm. With that said, it looks to me that dendrite_sytest.sh only ever uses postgres. Maybe the BuildKite config I've been adapting was mistaken?

The script always sets up postgres. Some other bits of perl then determine if we're going to actually use postgres. Need to ensure $POSTGRES is non-empty iff we want to use SQLite.

scripts/dendrite_sytest.sh
DMRobertson: Just on BuildKite. I've made it so that we don't run this unless `BUILDKITE_LABEL` is set.

DMRobertson:
@anoadragon453  Updated as per your comments, (many thanks).

These runs might be of interest:

* https://github.com/DMRobertson/sytest/actions/runs/1127951256
* https://github.com/DMRobertson/sytest/actions/runs/1128005188 - including a flakey failure?

.github/workflows/pipeline.yml
anoadragon453: Hmm, not sure if I understand. If `POSTGRES` is empty, then we set it to true/1 instead...?

Is the same required for `BLACKLIST` and `API`?

DMRobertson: Hmmm. I think I wrote out a larger comment but that seems to have disappeared.

I think it might be helpful to work through the details from scratch.

# What the vars actually do
## `POSTGRES`

`POSTGRES` is handled differently for synapse and dendrite.
* In `synapse_sytest.sh` we only read `POSTGRES` in tests `[ -n "$POSTGRES" ]`. We need `POSTGRES` to be empty or not set for SQLite, and nonempty to use postgres.
* In `Dendrite.pm` we test if `POSTGRES` is not defined or equal to `'0'`, and use SQLlite otherwise.

Taking both behaviours into account: 
* to use postgres, set `POSTGRES` to any non-empty string other than `'0'`
* to use sqllite, don't define `POSTGRES`

## `API`
Only for dendrite. If `$ENV{'API'} == '1'` in `Dendrite.pm`, we'll pass `-api` to dendrite. Not sure how that will handle the var being undefined.

## `BLACKLIST`
Only for synapse. If `"$BLACKLIST"` is empty (`test -z`) then we set `BLACKLIST=sytest-blacklist`. It ends up getting passed to `run-tests.pl` via `-B`.

# GitHub Actions' expression language

The language is typed, but presumably `${{ expr }}` converts `expr` to a string before substituting it into the yaml.

DMRobertson: I found the best way to experiment was to run a dummy action here: https://github.com/DMRobertson/gha-test/

As far as I can tell, `a && b` returns `a` if `a` is falsey, otherwise `b`. No type conversion done on the value that gets returned.

The two cases then:

*  when `postgres` isn't included in the strategy matrix
    - `matrix.postgres` is `null`
    - `null && 1` evaluates to null.
    - `${{ null }}` subsitutes the text `null` into the yaml file.
    - we end up with a yaml KV pair `POSTGRES: null`
    - GHA interprets this as "don't set `POSTGRES`"

* when `postgres: postgres` is in the strategy matrix
  - `matrix.postgres` is `'postgres'`
  - `'postgres && 1` is `1`
  - `${{ 1 }}` substitutes the text `null` into the yaml file
  - yaml KV `POSTGRES: 1`
  - GHA sets the `POSTGRES` env var to the string `1`

I'm about to add a commit which confirms this by dumping the environment before we call the bootstrap script.

DMRobertson:
Most recent run: https://github.com/DMRobertson/sytest/runs/3360789317?check_suite_focus=true

scripts/dendrite_sytest.sh
```diff
@@ -68,6 +68,7 @@ echo >&2 "--- Copying assets"
 # Copy out the logs
 rsync -r --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /logs --include "*/" --include="*.log.*" --include="*.log" --exclude="*"
 
+# Generate annotate.md This is Buildkite-specific.
```
anoadragon453, line 71:
```suggestion
# Generate annotate.md. This is Buildkite-specific.
```

scripts/synapse_sytest.sh
```diff
@@ -201,6 +201,7 @@ rsync --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /log
 #export TOP=/src
 #/venv/bin/coverage combine
 
+# Generate annotate.md This is Buildkite-specific.
```
anoadragon453, line 204:
```suggestion
# Generate annotate.md. This is Buildkite-specific.
```

.github/workflows/pipeline.yml
```diff
@@ -20,23 +20,23 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - label: Python 3.6 / SQLite / Monolith
+          - label: Py3.6 / SQLite / Monolith
```
anoadragon453: bionic has SQLite 3.22.0 - it can be useful to list it, we've had issues with older SQLite's in the past.

anoadragon453 (commented):
Thanks for answering my query in so much detail! I'm afraid at the moment that your links seem to lead to a private repository that I can't access. I'd like to take a look at the CI runs before approving.

Thanks for tackling the other items as well. Very small nits below in the meantime.

DMRobertson:
> I'm afraid at the moment that your links seem to lead to a private repository that I can't access. I'd like to take a look at the CI runs before approving.

Oh bugger. I initially had a public fork where I was testing this stuff out. It looks like I've gone on to delete that fork, presumably when I realised the done thing was to make a branch within the main matrix-org repo. That is unfortunate and presumably means that all the actions runs are gone.

I think it might be best to push the branch I was working on directly to the main sytest repo and open a new pull request.

anoadragon453:
Oh, I've just realised I reviewed the other PR while it was marked as a draft. I'm not sure if that PR was intended just to show off the CI or whether we were moving discussion there.

DMRobertson:
See the other PR, #1127

PR #1117, opened 2021-08-11
Don't send garbage to homeserver in more cross-signing tests

neilalexander:
Continuation of #1107 so that we send things that actually look like keys/signatures instead of just sending nonsense.

### DIFF
--- a/tests/41end-to-end-keys/08-cross-signing.pl
+++ b/tests/41end-to-end-keys/08-cross-signing.pl
@@ -160,12 +160,12 @@ test "Changing master key notifies local users",
          matrix_put_e2e_keys( $user1, device_keys => {
              "algorithms" => ["m.olm.curve25519-aes-sha256", "m.megolm.v1.aes-sha"],
              "keys" => {
-                 "curve25519:".$device_id => "curve25519+key",
-                 "ed25519:".$device_id => "ed25519+key",
+                 "curve25519:".$device_id => "IQ/Hu4GGOaxIpMavovFYGouVJeIP2miSfysv+Db3NXg",
+                 "ed25519:".$device_id => "MKkClRdltZlOHyCzxiDrm7MsDAsohXmAyeu2cYO6how",
              },
              "signatures" => {
                  $user_id => {
-                     "ed25519:".$device_id => "self+signature",
+                     "ed25519:".$device_id => "KUQR2vSQvcv40MiKoJSW7kfi9vBDhmUtzZ9/pBzteYZVNw8KB0INPle3w71huQY/6tkk3Kfi/mzo7cNkiiHkuQ",
                  },
              },
          } );
@@ -188,8 +188,8 @@ test "Changing master key notifies local users",
              "device_id" => $device_id,
              "algorithms" => ["m.olm.curve25519-aes-sha256", "m.megolm.v1.aes-sha"],
              "keys" => {
-                 "curve25519:".$device_id => "curve25519+key",
-                 "ed25519:".$device_id => "ed25519+key",
+                 "curve25519:".$device_id => "IQ/Hu4GGOaxIpMavovFYGouVJeIP2miSfysv+Db3NXg",
+                 "ed25519:".$device_id => "MKkClRdltZlOHyCzxiDrm7MsDAsohXmAyeu2cYO6how",
              }
          };
          sign_json(
@@ -222,7 +222,7 @@ test "Changing master key notifies local users",
                            "ed25519:EmkqvokUn8p+vQAGZitOk4PWjp7Ukp3txV2TbMPEiBQ" );
          assert_deeply_eq( $content->{device_keys}->{$user_id}->{$device_id}
                            ->{signatures}->{$user_id}, {
-            "ed25519:".$device_id => "self+signature",
+            "ed25519:".$device_id => "KUQR2vSQvcv40MiKoJSW7kfi9vBDhmUtzZ9/pBzteYZVNw8KB0INPle3w71huQY/6tkk3Kfi/mzo7cNkiiHkuQ",
             "ed25519:EmkqvokUn8p+vQAGZitOk4PWjp7Ukp3txV2TbMPEiBQ" => $cross_signature,
          } );
 
@@ -240,7 +240,7 @@ test "Changing master key notifies local users",
                            "ed25519:EmkqvokUn8p+vQAGZitOk4PWjp7Ukp3txV2TbMPEiBQ" );
          assert_deeply_eq( $content->{device_keys}->{$user_id}->{$device_id}
                            ->{signatures}->{$user_id}, {
-            "ed25519:".$device_id => "self+signature",
+            "ed25519:".$device_id => "KUQR2vSQvcv40MiKoJSW7kfi9vBDhmUtzZ9/pBzteYZVNw8KB0INPle3w71huQY/6tkk3Kfi/mzo7cNkiiHkuQ",
             "ed25519:EmkqvokUn8p+vQAGZitOk4PWjp7Ukp3txV2TbMPEiBQ" => $cross_signature,
          } );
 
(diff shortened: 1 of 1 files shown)


kegsay (approved):

richvdh (changes requested):
x

richvdh (changes requested):
I feel like we've discussed this in chat, failed to reach an agreement, and suddenly now I see PRs to update the integration test suite. Please can we actually agree that's the right thing to do rather than just making PRs which accord with one particular worldview?

Failure to treat signing keys as opaque binary is a server bug imho. Change my mind.

richvdh:
> signing keys

and signatures

richvdh:
> All I am trying to do here is to make the tests send some values that look remotely like what a server is supposed to see in the real world

ok, but sytest also needs to test that a server behaves correctly in the face of things that have not yet been specified, but might be in the future. In this case, the server needs to be able to handle potential future key algorithms which do not use base64.

> The spec doesn't say that keys are opaque — irrespective of what the key algorithm is or what length the key is, the spec says they should be encoded in base64:

It is possible for the spec to be wrong, or at least misleading, and I would argue heavily that this is such a case. I think what  this thing is *trying* to say is that the `<id>` part of the `<algorithm>:<id>` property name should be the same as the property value. In any case, this only applies to cross-signing keys, which I don't think are being changed in this PR.

> As for the signatures, for which everything seems to reference the "Signing JSON" section, which also states that they're base64 too:

Possibly, but I think it comes down to who needs to verify that signature. Again, I would say that if the server has no need to verify the signature, it has no need to enforce that it be Base64.

 > Surely the tests should just do what the spec says.

Yes, but:
 * (a) there's room for interpretation here. If the spec says "this should be base64-encoded", that can just be informational: "... because clients that receive it will expect that, given the current state of the ecosystem". It doesn't necessarily mean that servers should enforce that. And again, I argue in this case they should not, to allow clients to use other algorithms in future.
 * (b) If the spec says "jump off a cliff"... please don't write a test that enforces cliff-jumping.

---

At the end of the day, I don't think that attempting to refer to the letter of the spec is going to produce satisfactory answers here, and we should instead consider what makes most sense - possibly in consultation with @uhoreg, who designed most of these APIs.

My opinion is that enforcing Base64 in these fields is an unnecessary coupling between clients and servers, which reduces the future flexibility in clients. I still don't really get why you are decoding these fields anyway?

uhoreg:
> ... possibly in consultation with @uhoreg, who designed most of these APIs.

It looks like these changes are mostly with respect to the device keys rather than the cross-signing keys, so I'm not sure if my opinion is relevant here, but if my opinion is wanted, I agree with vdh in that servers should be able to handle unknown things fine, but if you're sending it data for a known algorithm (`curve25519`/`ed25519`), it doesn't seem unreasonable for homeservers to yell if you give them invalid data.

neilalexander:
> ok, but sytest also needs to test that a server behaves correctly in the face of things that have not yet been specified, but might be in the future.

This mentality is a source of constant frustration. When writing Dendrite, we try to follow the spec as far as possible, only then to find tests don't pass. We waste huge amounts of time digging into why only to find that the tests and the spec disagree. This will only get worse with the addition of Complement tests as well because now we have three "sources of truth" instead of two and still no one is willing to point at one of these and say "that one is authoritative". 

It's fine for the spec to be wrong but it does not feel fine at all for the spec to deliberately say one thing and for our integration tests to do another. We need to be prepared to keep both in lockstep or otherwise this is just an exercise in futility for anyone who tries to build a homeserver. Building a homeserver is already difficult enough without these extra landmines.

> In this case, the server needs to be able to handle potential future key algorithms which do not use base64.

I'm trying very hard here to avoid conflating base64 with the key algorithms because ultimately the requirement for base64 here is nothing to do with the key algorithms — neither ed25519 or curve25519 care about base64. They care about binary data and we only have base64 here because that's how we have chosen to express binary data in JSON. Pragmatically speaking, is that likely to change?

> Possibly, but I think it comes down to who needs to verify that signature. Again, I would say that if the server has no need to verify the signature, it has no need to enforce that it be Base64.

Is it unreasonable for the server to want to perform *any* sanity checks on the data that is uploaded? Without those kinds of validations, what stops me from just uploading a 4GB blob of nonsense as a "signature", or filling the database with signatures for key algorithms that are fictional and never going to exist?

> (a) there's room for interpretation here. If the spec says "this should be base64-encoded", that can just be informational: "... because clients that receive it will expect that, given the current state of the ecosystem". It doesn't necessarily mean that servers should enforce that. And again, I argue in this case they should not, to allow clients to use other algorithms in future.

It's not clear anywhere what level of enforcement servers should or shouldn't do because it's not documented anywhere. But, as above, I am still not convinced that "no enforcement" is the right answer.

kegsay:
I agree with @neilalexander - sytest should absolutely be testing what is defined in the spec. I also agree that we want some tests to assert that server implementations can deal with new and unknown algorithms. That, however, is not what these tests were written to do. They were done as a basic sanity check of the cross-signing APIs.

By all means, we can have additional tests for testing new and unknown algorithms and we absolutely should, but then we shouldn't be using the `ed25519:` and `curve25519:` prefixes. If Dendrite barfs on unknown prefixes then yes, we can fix that. But as it stands, sytest is sending bad data which I feel server implementations should absolutely try to catch at upload time.

richvdh:
> By all means, we can have additional tests for testing new and unknown algorithms and we absolutely should, but then we shouldn't be using the ed25519: and curve25519: prefixes. If Dendrite barfs on unknown prefixes then yes, we can fix that. But as it stands, sytest is sending bad data which I feel server implementations should absolutely try to catch at upload time.

If you'd like to add another test that uses a different algorithm, and *not* base64, I'll be happy to accept this PR.

If that isn't an acceptable compromise, I think we should schedule a call to discuss this further. Maybe we should do that anyway if Neil is frustrated with the approach in general.

neilalexander:
> If you'd like to add another test that uses a different algorithm, and not base64, I'll be happy to accept this PR.

I added another test to the PR, named, `Can upload keys with unknown algorithms and formats`. It fails as expected on Dendrite, but I was surprised to find it also fails on Synapse:

```
# Started: 2021-08-12 12:30:53.788
# Ended: 2021-08-12 12:30:54.037
# HTTP Request failed ( 400 Bad Request https://localhost:8840/_matrix/client/unstable/keys/device_signing/upload?access_token=syt_YW5vbi0yMDIxMDgxMl8xMjE4NDQtNjE3_GSmLyYyuctfNjMTKenOZ_1AKT3g ) from POST https://localhost:8840/_matrix/client/unstable/keys/device_signing/upload?...
# {"errcode":"M_INVALID_PARAM","error":"Invalid master key"}
# 0.140601: Registered new user @anon-20210812_121844-617:localhost:8840
```

It turns out Synapse also assumes base64 values in some places, like here: https://github.com/matrix-org/synapse/blob/develop/synapse/handlers/e2e_keys.py#L901-L902

The tests that hit this codepath are seemingly sending valid base64-encoded signatures from a device key to a master key so that it can check the master key is correctly signed. So even Synapse appears to care about the algorithm and format in places — first it assumes it's base64 encoded and secondly it assumes it's an ed25519 signature. If you were to just follow the letter of the spec, those would be reasonable assumptions to make.

The new test tries to upload a key with a `!` character specifically because it shouldn't be present in a base64-encoded string. In the old test, Synapse doesn't find any odd-looking characters and tries to decode as base64 anyway, even though the resulting value is nonsense.

It's basically doing the same thing Dendrite was, but because Synapse doesn't appear to round-trip the base64 (which sometimes results in a different encoded value which would trip `assert_deeply_eq`) or produce an HTTP error, the tests just never notice there's a problem. Ultimately the same failure mode but manifests in a different hidden way.

> If that isn't an acceptable compromise, I think we should schedule a call to discuss this further. Maybe we should do that anyway if Neil is frustrated with the approach in general.

This is probably a good idea anyway because it is clear that there's no consensus on how cross-signing should be implemented properly, or about what level of validation makes sense, and it is frustrating trying to solve this problem properly when the spec, the tests and Synapse already disagree with each other.

richvdh:
> It turns out Synapse also assumes base64 values in some places

hawkward

richvdh:
I wrote in chat: The new test slightly moves the goalposts by changing the algorithm of cross-signing keys, whereas the original test is about changing the algorithm of device keys. So this leaves us in the situation where synapse accepts unknown algorithms for device keys, but not cross-signing keys, which does feel a bit inconsistent.

@uhoreg said: there was no particular *need* for Synapse to verify cross-signing signatures, other than to help stop clients messing up.

There was agreement between @uhoreg and myself that a server should accept cross-signing keys and signatures using unknown algorithms. This implies that it *cannot* verify those signatures. Whether that means that the server should not verify *any* signatures, or whether it should verify signatures for known algorithms and ignore them for unknown ones, should be up to the implementation.

---

A later point arose: what if we consider the base64 encoding to be a property of the transport (REST/JSON) rather than of the algorithm (ed25519, etc). For example, if we used CBOR, would clients and servers be expected to use the raw data rather than base64 encoding? It seems likely they would - and it would therefore be necessary for servers to handle base64 encoding when bridging between REST and CBOR.  In that case, @neilalexander's original PR  (https://github.com/matrix-org/sytest/pull/1117/commits/8bb6e68a33db76c01dd46e1aca63cbbc26fd87f5) is entirely reasonable.

PR #1119, opened 2021-08-13
Don't rely on device list being updated on invite/join for x-signing federation tests

neilalexander:
The tests were using `sync_until_user_in_device_list` before,  so even if a device list update is generated for the key/signature change, it will still fail if device list updates aren't generated for the invite/join, which seems like a very unfortunate side effect.

### DIFF
--- a/tests/41end-to-end-keys/08-cross-signing.pl
+++ b/tests/41end-to-end-keys/08-cross-signing.pl
@@ -527,11 +527,37 @@ test "uploading self-signing key notifies over federation",
       })->then( sub {
          matrix_invite_user_to_room( $user1, $user2, $room_id )
       })->then( sub {
-         sync_until_user_in_device_list( $user1, $user2 );
+         await_sync_timeline_contains( $user1, $room_id, check => sub {
+            my ( $event ) = @_;
+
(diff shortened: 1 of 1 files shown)


richvdh:
For links: some of these lines were added in #749.

richvdh:
The problem here is that it's presumably legitimate for a server to send the join event in one /sync and the device update in the next (or vice-versa)

If your only reason for wanting to change these things is "there are already separate tests for device list updates", I'd probably rather leave it as-is rather than risk making the tests flaky again.

neilalexander:
Yeah — it was largely because I still wanted to test the x-signing behaviour but I don't think Dendrite always generates device change notifications when users are invited/join yet, but if it introduces a race then we may have to rethink.

richvdh:
I'm not absolutely convinced that your change will introduce a race - note that #749 added a `sync_until_user_in_device_list` where there was nothing at all before - having a `sync_until_join_event` (isn't there a utility method we can use for that?) may well serve the purpose adequately, so if this is a blocker to you we can always try merging it and seeing what happens. But if it's just a matter of holding off until Dendrite implements a bit more stuff, I'd wait for that.

PR #1121, opened 2021-08-18
Extend cross-room backfill request to work with different impls

richvdh:
Servers should not be *obliged* to send a `state_ids` request if it already has
the `auth_events` of a given event.


lib/SyTest/Federation/Server.pm
```diff
@@ -338,7 +338,6 @@ sub mk_await_request_pair
 
       return $self->{$okey}{$ikey} = Future->new
          ->on_cancel( sub {
-            warn "Cancelling unused $shortname await for @paramvalues";
```
richvdh, line 341: this was required because we now cancel the `await_request_state_ids` request if it is unused. That's normal behaviour, so we shouldn't warn on it.

richvdh:
this seems to be consistently failing CI with a test that isn't normally flaky. will take a closer look.

PR #1122, opened 2021-08-19
We do actually want CI to pass a branch

DMRobertson:


DMRobertson:
This one's in my personal fork, which means the branch lookups are going to fail. Will reopen in the not-fork.

PR #1123, opened 2021-08-19
We do actually want CI to pass a branch

DMRobertson:
this pr not from my fork but from the repo itself

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,7 +1,8 @@
 # SyTest Docker Images
 
 These Dockerfiles create containers for running SyTest in various
-configurations. SyTest is not [included in these images by default](#selecting-a-sytest-checkout),
+configurations. SyTest is not included in these images (see 
+[Selecting a checkout of SyTest](#selecting-a-sytest-checkout)),
 but its dependencies are.
 
 Included currently is:
--- a/docker/README.md
+++ b/docker/README.md
@@ -84,6 +84,9 @@ docker run --rm -it \
     matrixdotorg/sytest-synapse:buster
 ```
 
+If the branch referred to by `SYTEST_BRANCH` does not exist, `develop` is used
+instead.
+
 If you would like to run tests with an existing checkout of Sytest, add a
 volume to the docker command mounting the checkout to the `/sytest` folder in
 the container:
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,8 +1,8 @@
 # SyTest Docker Images
 
 These Dockerfiles create containers for running SyTest in various
-configurations. SyTest is not included in these images, but its dependencies
-are.
+configurations. SyTest is not [included in these images by default](#selecting-a-sytest-checkout),
+but its dependencies are.
 
 Included currently is:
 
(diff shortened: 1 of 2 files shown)


richvdh (commented):
looks sensible, but we should document it - https://github.com/matrix-org/sytest/blob/7178c8bd53557c739e99771bf67c9116ec0d731f/docker/README.md lists all the other env vars that the docker image understands, so we should include this one

docker/README.md
```diff
@@ -17,9 +17,12 @@ Included currently is:
 
 ### Synapse
 
-The `sytest-synapse` images expect a checkout of the synapse git repository to
-be mounted at `/src`; additionally, server logs will be written to `/logs`, so
-it is useful to mount a volume there too.
+The `sytest-synapse` and `sytest-dendrite` images expect a checkout of the 
+homeserver git repository to be mounted at `/src`; otherwise, the image will
+try to fetch that repository from GitHub.
```
richvdh, line 22: Is this correct? I thought that it would just fail if `/src` didn't exist.

docker/README.md
```diff
@@ -29,6 +32,8 @@ docker run --rm -it -v /path/to/synapse\:/src:ro -v /path/to/where/you/want/logs
 
 The following environment variables can be set with `-e` to control the test run:
 
+ * `SYTEST_BRANCH`: controls which branch of synapse or dendrite we fetch when  `/src`
+   is missing. If omitted or if that branch doesn't exist, we'll fetch the  appropriate development branch.
```
richvdh, line 36: this isn't quite right. As above: if `/src` is missing, I don't think it works at all...

richvdh (changes requested):

docker/README.md
DMRobertson: Argh! Thanks. I think I keep confusing `/src` and `/sytest`. It's the CI job that has to fetch the source

richvdh: No worries! It is pretty confusing.

docker/README.md
```diff
@@ -1,8 +1,8 @@
 # SyTest Docker Images
 
 These Dockerfiles create containers for running SyTest in various
-configurations. SyTest is not included in these images, but its dependencies
-are.
+configurations. SyTest is not [included in these images by default](#selecting-a-sytest-checkout),
+but its dependencies are.
```
richvdh, line 5: it's not actually possible to include Sytest itself in the Docker images. How about:


```suggestion
configurations. SyTest is not included in these images (see 
[Selecting a checkout of SyTest](#selecting-a-sytest-checkout)),
but its dependencies are.
```

docker/README.md
```diff
@@ -68,11 +68,20 @@ it is useful to mount a volume there too.
...
+the same name as that of the target homeserver, or `develop`.
+This can be overridden with the environment variable `SYTEST_BRANCH`,
+for example:
+
+```
+docker run --rm -it \
+    -e SYTEST_BRANCH="my-sytest-branch"
+    -v /path/to/synapse\:/src:ro -v /path/to/where/you/want/logs\:/logs
+    matrixdotorg/sytest-synapse:buster
+```
```
richvdh, line 85:
```suggestion

If the branch referred to by `SYTEST_BRANCH` does not exist, `develop` is used
instead.

```

richvdh (approved):
lgtm otherwise, thanks!

DMRobertson:
I'll run the docker `build.sh` and `publish.sh` scripts so that synapse's CI can pick this up.

PR #1127, opened 2021-08-25
Move from Buildkite to GitHub Actions, attempt 2

DMRobertson:
Migrated from https://github.com/matrix-org/sytest/pull/1115  after I screwed things up by deleting a personal fork.

On the plus side, sticking it in this repo should mean that we can see the Actions directly here.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -16,47 +16,38 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Synapse - :python: 3.6 / SQLite / Monolith"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: bionic
 
-          - name: "Synapse - :python: 3.6 / :postgres: 10 / Monolith"
-            homeserver-implementation: synapse
+          - homeserver-implementation: synapse
             sytest-tag: bionic
             postgres: postgres
 
-          - name: "Synapse - :python: 3.6 / :postgres: 10 / Workers"
-            homeserver-implementation: synapse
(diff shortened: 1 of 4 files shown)


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,156 @@
...
+
+          - label: SQLite, full HTTP APIs
+            api: full-http
+
+          - label: Postgres
+            postgres: postgres
+
+          - label: Postgres, full HTTP APIs
+            postgres: postgres
+            api: full-http
```
anoadragon453, line 112: I suppose we don't have a clear idea of postgres versions here as there's no distribution label for `sytest-dendrite`?

anoadragon453 (approved):
(Interestingly the actions runs seem to have appeared on the other PR as well).

Looking through the test results I think they look good! Some test failures, but they look like flakes - possibly an outdated Sytest branch?

Carrying over from the old PR, I think there's only minor tweaks. Otherwise this looks fine to merge to me.

.github/workflows/pipeline.yml
DMRobertson: `sytest-dendrite:latest` looks to be based on Debian buster at the time of writing according to `docker/build.sh:16`. I was a bit hesitant to write in a PG version for a tag that could change under our feet. But we already do this for `sytest-synapse:testing`, based on `sytest:testing`, based on `debian:testing`.

anoadragon453: Good point, yeah. Actually, I suppose it might be a good idea to simply remove the version of `debian:testing` as well. I can't see us really remembering to ever update that version number...

PR #1128, opened 2021-09-03
Move /createRoom to the event_creator worker

anoadragon453:
This tests that `/createRoom` actually works on workers, which turns out is only true after https://github.com/matrix-org/synapse/pull/10757 lands.

Thus this PR shares a branch name with that PR.

The failing tests are a bug with the GHA runs, but buildkite should be passing!


DMRobertson:
> The failing tests are a bug with the GHA runs, but buildkite should be passing!

I think I understand the problem. We're fetching

```
https://github.com/matrix-org/synapse/archive/refs/pull/1128/merge.tar.gz
```

which is an ancient commit from https://github.com/matrix-org/synapse/tree/matthew/fix-roomdir-pagination that doesn't contain the blacklist file.

We want to fetch
```
https://github.com/matrix-org/synapse/archive/anoa/create_room_workers_aliases/archive.tar.gz
```

Will have to see what the right invocation is here. But it's not `${GITHUB_REF#refs/heads/}`!

DMRobertson:
Looks like @clokep  has [done the hard work for us already](https://github.com/matrix-org/synapse/commit/a14884fbb050f5069b83e344e4ef12a54fe42111).

erikjohnston (approved):

PR #1129, opened 2021-09-06
Correctly fetch same-named homeserver branches

DMRobertson:
In #1115  I didn't correctly handle the situation where there's a corresponding synapse/dendrite branch to fetch. #1128 fell victim to this.

Fix: reuse the shell script from matrix-org/synapse#10160.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -147,12 +147,30 @@ jobs:
         with:
           path: sytest
 
+      # TODO the shell script below is nicked from complement. We use this pattern
+      # in a few places. Can we make this an Action so it's easier to reuse?
       - name: Fetch corresponding dendrite branch
         shell: bash
         run: |
-          BRANCH=${GITHUB_REF#refs/heads/}
-          (wget -O - https://github.com/matrix-org/dendrite/archive/$BRANCH.tar.gz || wget -O - https://github.com/matrix-org/dendrite/archive/master.tar.gz) \
-            | tar -xz --strip-components=1 -C /src/
+          # Attempt to use the version of dendrite which best matches the current
+          # build. Depending on whether this is a PR or release, etc. we need to
+          # use different fallbacks.
+          #
+          # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
+          #    for pull requests, otherwise GITHUB_REF).
+          # 2. Attempt to use the base branch, e.g. when merging into release-vX.Y
+          #    (GITHUB_BASE_REF for pull requests).
+          # 3. Use the default dendrite branch ("master").
+          for BRANCH_NAME in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "master"; do
+            # Skip empty branch names and merge commits.
+            if [[ -z "$BRANCH_NAME" || $BRANCH_NAME =~ ^refs/pull/.* ]]; then
+              continue
+            fi
+            (wget -O - "https://github.com/matrix-org/dendrite/archive/$BRANCH_NAME.tar.gz" \
+              | tar -xz --strip-components=1 -C /src/) \
+              && echo "Successully downloaded and extracted $BRANCH_NAME.tar.gz" \
+              && break
+          done
 
       - name: Run sytest
         run: |
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -81,7 +81,7 @@ jobs:
             fi
             (wget -O - "https://github.com/matrix-org/synapse/archive/$BRANCH_NAME.tar.gz" \
               | tar -xz --strip-components=1 -C /src/) \
-              && echo "Successully downloaded and extracted $BRANCH_NAME.tar.gz" \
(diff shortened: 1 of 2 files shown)


DMRobertson:
Slightly anxious that dendrite fails on GHA but doesn't on BK. Flakey test?

.github/workflows/pipeline.yml
```diff
@@ -60,12 +60,30 @@ jobs:
...
+          # Attempt to use the version of synapse which best matches the current
+          # build. Depending on whether this is a PR or release, etc. we need to
+          # use different fallbacks.
+          #
+          # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
+          #    for pull requests, otherwise GITHUB_REF).
+          # 2. Attempt to use the base branch, e.g. when merging into release-vX.Y
+          #    (GITHUB_BASE_REF for pull requests).
+          # 3. Use the default synapse branch ("develop").
+          for BRANCH_NAME in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "develop"; do
```
anoadragon453, line 77: Since we had an issue with `"${GITHUB_REF#refs/heads/}"` due to PR numbers, is it even worth including it on this list?

.github/workflows/pipeline.yml
```diff
@@ -60,12 +60,30 @@ jobs:
         with:
           path: sytest
 
+      # TODO the shell script below is nicked from complement. We use this pattern
+      # in a few places. Can we make this an Action so it's easier to reuse?
```
anoadragon453, line 64: +1

.github/workflows/pipeline.yml
```diff
@@ -129,12 +147,30 @@ jobs:
...
+          #    (GITHUB_BASE_REF for pull requests).
+          # 3. Use the default dendrite branch ("master").
+          for BRANCH_NAME in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "master"; do
+            # Skip empty branch names and merge commits.
+            if [[ -z "$BRANCH_NAME" || $BRANCH_NAME =~ ^refs/pull/.* ]]; then
+              continue
+            fi
+            (wget -O - "https://github.com/matrix-org/dendrite/archive/$BRANCH_NAME.tar.gz" \
+              | tar -xz --strip-components=1 -C /src/) \
+              && echo "Successully downloaded and extracted $BRANCH_NAME.tar.gz" \
```
anoadragon453, line 171:
```suggestion
              && echo "Successfully downloaded and extracted $BRANCH_NAME.tar.gz" \
```

anoadragon453 (commented):
> Slightly anxious that dendrite fails on GHA but doesn't on BK. Flakey test?

I haven't seen that test flake yet, but I suppose see what happens after another run.

.github/workflows/pipeline.yml
DMRobertson: GITHUB_HEAD_REF and GITHUB_BASE_REF only exist when the action is triggered by a pull request. https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables. This action runs on specific branches too, outside of the context of pull requests:

```
name: Run sytest
on:
  push:
    branches: ["develop", "release-*"]
  pull_request
```

(but note that's only there because I copied it from synapse's CI file)

Assuming it's meaningful to run sytest on develop and release branches, I think we do still want to read GITHUB_REF.

DMRobertson:
> > Slightly anxious that dendrite fails on GHA but doesn't on BK. Flakey test?
> 
> I haven't seen that test flake yet, but I suppose see what happens after another run.

Most recent run seems to have fallen over with workers. Maybe the blacklist is still wrong?

DMRobertson:
Those failures seem to be replicated on buildkite. Maybe I need to update the corresponding synapse branch to pull in a fix there? /shrug

DMRobertson:
That looks happier. 

> > Slightly anxious that dendrite fails on GHA but doesn't on BK. Flakey test?
> 
> I haven't seen that test flake yet, but I suppose see what happens after another run.

There we were talking about a dendrite failure (`#458: Local device key changes appear in /keys/changes`) in  https://github.com/matrix-org/sytest/runs/3524830553.

Later we had https://github.com/matrix-org/sytest/runs/3527019236 where all dendrite tests suceeded.

The most recent run is https://github.com/matrix-org/sytest/runs/3527233610 where that dendrite failure reoccurred. 

In both failure cases it was Dendrite with Postgres but without Full HTTP APIs that saw this. Some kind of race?

.github/workflows/pipeline.yml
anoadragon453: Aha, I see. So in the case of a pull request, it's unlikely that we'll end up using `"${GITHUB_REF#refs/heads/}"`. But it's useful on `develop` / `release-*`.

Yes, I think it's worthwhile to run Sytest on those branches.

anoadragon453 (approved):
> In both failure cases it was Dendrite with Postgres but without Full HTTP APIs that saw this. Some kind of race?

Hmm, it does look that way, though I can't quickly tell you why that might be.

Perhaps worth raising on issue on the Dendrite repo. Otherwise I think this PR as it stands is good to go.

PR #1132, opened 2021-09-08
Pass format, url as data and http as a kind to /pushers/set

PiotrKozimor:
According to documentation (https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-pushers-set) `kind` field can be `http`, `email` or  `null` (although it's not defined as an enum). Also homeserver should not respect `testcustom` because `PusherData` is defined with another fields.

Signed-off-by: Piotr Kozimor <p1996k@gmail.com>

### DIFF
--- a/tests/61push/06_get_pusher.pl
+++ b/tests/61push/06_get_pusher.pl
@@ -10,7 +10,8 @@ test "Can fetch a user's pushers",
       my $device_display_name = "A testing machine";
       my $pushkey = "This is my pushkey";
       my $lang = "en";
-      my $customdata = "This is some custom data";
+      my $url = "https://dummy.url/_matrix/push/v1/notify";
+      my $format = "id_event_only";
 
       # create a pusher
       do_request_json_for( $alice,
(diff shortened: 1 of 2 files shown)


tests/61push/06_get_pusher.pl
```diff
@@ -50,7 +52,8 @@
          assert_eq( $pusher->{device_display_name}, $device_display_name, "device_display_name");
          assert_eq( $pusher->{pushkey}, $pushkey, "pushkey");
          assert_eq( $pusher->{lang}, $lang, "lang");
-         assert_eq( $pusher->{data}{testcustom}, $customdata, "custom data");
```
clokep, line 53: I think we still want to ensure that custom fields are passed through.

Looking at the [push gateway spec](https://matrix.org/docs/spec/push_gateway/r0.1.1#post-matrix-push-v1-notify), `data` that is sent to the push gateway should be:

> A dictionary of additional pusher-specific data. For 'http' pushers, this is the data dictionary passed in at pusher creation minus the `url` key.

I think the changes should be:

1. Keep the custom field.
2. Ensure the custom field, `format` are both passed through.
3. Ensure the `url` is not passed through.

clokep (changes requested):
Looks like it is on the right track, but I don't think it is handling custom data correctly.

tests/61push/06_get_pusher.pl
clokep: I think this looks OK, but I'm unsure if `url` should be there or not, going to see what others think.

richvdh: this is checking the response from `GET /r0/pushers`, rather than the data sent to the push gateway. This  looks correct to me.

richvdh (approved):

richvdh:
@PiotrKozimor  please can you sign off your contribution, per https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off ?

PiotrKozimor:
@richvdh I have signed off this PR.

richvdh:
thank you!

PR #1133, opened 2021-09-09
Fix missing `fi` in bootstrap.sh

DMRobertson:
I guess we never noticed this because no-one build and deployed the
sytest containers?

Context: I wasn't able to reproduce sytest failures from elsewhere and wanted to ensure I was running the latest version. Building the docker images locally and then trying to run tests with them failed with an error in bootstrap.sh


anoadragon453 (approved):
Huh, yes I'm not really sure how we missed that. Though this script has been kicking around for a good long while. It's possible that it hasn't been run outside of the context of CI since that bug was introduced.

Either way, thanks for fixing!

DMRobertson:
The sytest dendrite failure looks like a flake, which I've noted in https://github.com/matrix-org/dendrite/issues/2007

Issue #1136, opened 2021-09-10
`Device list doesn't change if remote server is down` is flaky


clokep:
https://github.com/matrix-org/synapse/runs/3570688017

I think this was just re-enabled?

clokep:
And https://github.com/matrix-org/synapse/runs/3587797065

clokep:
And https://github.com/matrix-org/synapse/runs/3588399409

DMRobertson:
Well, I mean to make a PR for this but screwed up my git branch management locally.

Let's see how this goes?

H-Shay:
Just noting that I had a failure on this test as well today (with several re-runs)

DMRobertson:
I resorted to sticking debug statements into synapse. My understanding: on the first keys query,

-  Synapse under test receives keys from mock server in sytest
- Synapse ends up calling `self.device_handler.device_list_updater.user_device_resync` on the master process (it seems device lists are always handled by the master)
- This later calls `update_remote_device_list_cache`
- At the end of that function we upsert into  `device_lists_remote_extremeties`
  -  `user_id='@__ANON__-0:localhost:45313' stream_id=3`

All normal operation as far as I can see. On the second keys query, 

- in `get_user_devices_from_cache` we call `get_device_list_last_stream_id_for_remotes` which reads from `device_lists_remote_extremeties`
- The response from this leads us to build `user_map={'@__ANON__-0:localhost:41307': None}`
- we act as if we've not stored keys from the previous federation request, so go and fetch them again
- this fails the test.

Guessing there's a race between the cache invalidation and the second device keys call?? Commenting out the `@cachedList` call makes the test pass (with `user_map={'@__ANON__-0:localhost:34477': '3'}`).

Notes:
  - `get_device_list_last_stream_id_for_remotes` function is decorated with `@cachedList`
  - it looks like we try to invalidate the cache right before we do the upsert mentioned above
  - cache done on the main worker, as is the query that builds "user_map"
  - the cache invalidation will be called "on the main twisted thread after the transaction has finished".

erikjohnston:
Oh bleurgh, that's annoying.

The other way we deal with this sort of thing is by using `retry_until_success { ... }` to wrap the test code, which will repeatedly run it until it passes (with some appropriate exponential backoff). The downside with this approach is that failures will result in the test timing out, rather than in a more helpful manner, so it's important to add helpful logging. An example usage:

[...]

PR #1140, opened 2021-09-15
Use right dockerfile for sytest-{synapse,dendrite}

DMRobertson:


erikjohnston (approved):

DMRobertson:
CI was expected to fail because this uses the (currently broken) containers that we're trying to build.

Ran this manually and after an additional fixup in https://github.com/matrix-org/sytest/commit/b4352ff050fe7865d9bba605eb428a08bb1f3f06 we build and pushed containers successfully.

@erikjohnston's CI reran at https://github.com/matrix-org/synapse/runs/3612174967?check_suite_focus=true and got further. As in, it actually ran a sytest. Once that run completes I'll merge this change in.

DMRobertson:
That build completed with only two sytest failures, both known flakes (#1136).

PR #1141, opened 2021-09-16
Configure push server database for dendrite

PiotrKozimor:
New component in dendrite is developed: https://github.com/matrix-org/dendrite/pull/1842. Sytest fails there because second instance of homeserver tries to access database file that is already opened by first instance.

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -236,6 +236,14 @@ sub _get_config
          },
       },
 
+      push_server => {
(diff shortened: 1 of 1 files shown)


kegsay (approved):

kegsay:
We plan to remove manual configs in Sytest in the near future, so new features can directly add their new config options to the [generate-config](https://github.com/matrix-org/dendrite/blob/master/cmd/generate-config/main.go) script. For now though, this PR is fine, it's just a heads up that this will change in the future.

PR #1143, opened 2021-09-16
Configure threepid database and emails for Dendrite

PiotrKozimor:
Configuration required for https://github.com/matrix-org/dendrite/pull/1970.

When session ID is composed of digits only, perl parses it to number. Then is passes number to `/r0/account/3pid` leading to errors. That's why I suggest converting back to string.

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -252,6 +252,7 @@ sub _get_config
             smtp => {
                host => $self->{smtp_server_config}->{host} . ':' . $self->{smtp_server_config}->{port},
             },
+         },
       },
 
       push_server => {
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -240,6 +240,19 @@ sub _get_config
                ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
                "file:$self->{hs_dir}/presence.db" : $db_uri,
          },
+         threepid_database => {
+            connection_string => 
+               ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
+               "file:$self->{hs_dir}/threepids.db" : $db_uri,
+         },
+         email => {
+            enabled => JSON::false,
+            templates_path => "/src/res/default",
+            smtp => {
+               host => "localhost:25",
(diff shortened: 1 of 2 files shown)


lib/SyTest/Homeserver/Dendrite.pm
```diff
@@ -240,6 +240,19 @@ sub _get_config
...
          },
+         threepid_database => {
+            connection_string => 
+               ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
+               "file:$self->{hs_dir}/threepids.db" : $db_uri,
+         },
+         email => {
+            enabled => JSON::true,
+            from => 'synapse@localhost',
+            templates_path => '/src/res/default',
```
richvdh, line 251: this seems to be a bit of a large assumption? Who is to say that the templates are here?

tests/11register.pl
```diff
@@ -227,7 +227,7 @@ sub add_email_for_user {
             three_pid_creds => {
                id_server       => $id_server->name,
                id_access_token => $id_access_token,
-               sid             => $sid,
+               sid             => "$sid",
```
richvdh, line 230: this isn't the only place that `sid` is passed from `validate_email` to `do_request_json` - if we're going to change it, we should probably change all of them. However, I'm surprised it's necessary - isn't it already returned as a string by `validate_email` (provided `/r0/account/3pid/email/requestToken` returns it as a string)?

Anyway, this feels like an orthogonal change to the configuration. Please could you make it a separate PR?

richvdh (changes requested):

lib/SyTest/Homeserver/Dendrite.pm
PiotrKozimor: Maybe we need review from someone from Dendrite as well. I put default templates [here](https://github.com/matrix-org/dendrite/pull/1970/files#diff-37e05f9cb30ba9467db9468307a09d046072f2b517f8e7738753b2515873018d) in the mentioned PR.

richvdh: my point is that you cannot assume Dendrite is at `/src`.

tests/11register.pl
PiotrKozimor: I will check again `/r0/account/3pid/email/requestToken`. It could be my implementation that is bad.

kegsay:
This is taking longer than I'd expect, so merging this for now.

PR #1144, opened 2021-09-20
Log request ids in do_json_request_for on failure

DMRobertson:
See corresponding synapse PR #TODO

### DIFF
--- a/tests/10apidoc/00prepare.pl
+++ b/tests/10apidoc/00prepare.pl
@@ -31,7 +31,11 @@ sub do_request_json_for
       params       => \%params,
       request_user => $user_id,
       %args,
-   );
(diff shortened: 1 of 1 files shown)


DMRobertson:
I chose `X-REQUEST-ID` based on https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_response_fields. It looks like that might be intended for client-generated ids rather than server-generated.

I didn't want to use a name that was synapse-specific in sytest though, e.g. what if dendrite or conduit want to expose a mechanism like this?

Could change synapse to only generate an id itself if the client hasn't provided a request id?

Another paranoid thought: is leaking the instance names and the id numbers a security risk? It's exposing internal details...

PR #1145, opened 2021-09-20
Attempt to deflake `The only membership state included in an initial …

### LINKED ISSUE #1050: `The only membership state included in an initial sync is for all the senders in the timeline` is flaky
https://github.com/matrix-org/synapse/pull/10040/checks?check_run_id=2668894009

DMRobertson:
…sync is for all the senders in the timeline`

Hopefully fixes #1050

### DIFF
--- a/tests/10apidoc/09synced.pl
+++ b/tests/10apidoc/09synced.pl
@@ -393,7 +393,7 @@ sub assert_state_room_members_match {
 
    my @members = grep { $_->{type} eq 'm.room.member' } @{ $events };
    @members == scalar @{ $member_ids }
-      or die "Expected only ".(scalar @{ $member_ids })." membership events";
+      or die "Expected only ".(scalar @{ $member_ids })." membership events (got ".(scalar @members).")";
 
    my $found_senders = {};
    my $found_state_keys = {};
--- a/tests/31sync/15lazy-members.pl
+++ b/tests/31sync/15lazy-members.pl
@@ -99,9 +99,21 @@ test "The only membership state included in an initial sync is for all the sende
             )
          }, foreach => [ 1 .. 10 ])
       })->then( sub {
+         # Ensure synapse has propagated this message to Alice's synchotron
+         await_sync_timeline_contains( $alice, $room_id, check => sub {
+            my ($event) = @_;
+            log_if_fail "event", $event;
+            my $result = $event->{type} eq "m.room.message" &&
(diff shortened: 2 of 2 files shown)


tests/31sync/15lazy-members.pl
```diff
@@ -99,9 +99,21 @@
             )
          }, foreach => [ 1 .. 10 ])
       })->then( sub {
+         # Ensure synapse has propagated this message to Alice's synchotron
```
anoadragon453, line 102:
```suggestion
         # Ensure synapse has propagated this message to Alice's sync stream
```

"synchrotron" is a bit of an unofficial name that we mostly just use to describe sync workers on Synapse.

tests/31sync/15lazy-members.pl
```diff
@@ -99,9 +99,21 @@
...
+         await_sync_timeline_contains( $alice, $room_id, check => sub {
+            my ($event) = @_;
+            log_if_fail "event", $event;
+            my $result = $event->{type} eq "m.room.message" &&
+               $event->{sender} eq $charlie->user_id &&
+               $event->{content}->{body} eq "Message 10";
+            $result
```
anoadragon453, line 109: Is this the equivalent of `return $result;`, similar to rust?

anoadragon453 (changes requested):
Looks sensible, just a couple clarifying comments.

Failing tests in Dendrite are unrelated.

tests/31sync/15lazy-members.pl
anoadragon453: Aha, OK. Thanks for the explanation. Also looks like there is no differentiator through the semicolon as I originally thought.

anoadragon453 (approved):

tests/10apidoc/09synced.pl
```diff
@@ -393,7 +393,7 @@ sub assert_state_room_members_match {
 
    my @members = grep { $_->{type} eq 'm.room.member' } @{ $events };
    @members == scalar @{ $member_ids }
-      or die "Expected only ".(scalar @{ $member_ids })." membership events";
+      or die "Expected only ".(scalar @{ $member_ids })." membership events (got ".(scalar @members).")";
```
richvdh, line 396: for future reference, it's generally easier to replace this sort of construction with a call to `assert_eq`. (`assert_*` didn't exist very early in sytest's history, so this sort of `foo == bar or die` construction typically predates its existence)

PR #1149, opened 2021-10-06
Update the soft-fail tests to work with recent room versions

richvdh:
We have some tests in `tests/50federation/52soft-fail.pl` which test soft-failing, but which have never been updated to work with room versions other than v1.

For the most part, this is just a matter of updating them to use utility functions instead of making assumptions about the shapes of the events, but it turns out that there is another wrinkle, which is that they were making some assumptions about the state-resolution algorithm which are not true in subsequent room versions. More info in the commit comments.

Reviewable commit-by-commit.

### DIFF
--- a/lib/SyTest/Assertions.pm
+++ b/lib/SyTest/Assertions.pm
@@ -12,6 +12,7 @@ our @EXPORT_OK = qw(
    assert_ok
    assert_eq
    assert_deeply_eq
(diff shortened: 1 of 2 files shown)


clokep (approved):
Pretty sure all this makes sense!

Is this the sort of test that would benefit from running over multiple room versions?

PR #1150, opened 2021-10-06
Test basic functionality on room versions 7, 8, and 9.

### LINKED ISSUE #1124: room_versions tests lack tests for room v7 and v8
`tests/32room-versions.pl` includes a bunch of tests which are supposed to be done for *all* room versions, but it doesn't include v7 or v8

clokep:
Fixes #1124 by re-using the declared supported room versions.

I'm not 100% sure this is the right thing to do, but it seems frustrating to have two lists of room versions which are almost identical.


clokep:
(I think the Dendrite failures are expected right now?)

PR #1161, opened 2021-10-29
Three fixes for aiding Dendrite push notifications development

tommie:
While working on https://github.com/matrix-org/dendrite/pull/1842, I discovered some issues that would help development:

* Clear pending HTTP awaiters between tests.
* Make "Notifications can be viewed with GET /notifications" sync before sending receipts.
* Fix `scripts/dendrite_sytest.sh` to show are-we-synapse-yet in the console, and chmod logs.

More context is in the commit messages.

Since they are all small, I'm hoping a single PR is acceptable.

### DIFF
--- a/scripts/dendrite_sytest.sh
+++ b/scripts/dendrite_sytest.sh
@@ -61,12 +61,13 @@ fi
 
 # Check for new tests to be added to the test whitelist
 /src/show-expected-fail-tests.sh /logs/results.tap /src/sytest-whitelist \
-    /src/sytest-blacklist > /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
+    /src/sytest-blacklist | tee /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
 
 echo >&2 "--- Copying assets"
 
 # Copy out the logs
 rsync -r --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /logs --include "*/" --include="*.log.*" --include="*.log" --exclude="*"
+find /logs | xargs -r chmod go+rX
 
 # Generate annotate.md. This is Buildkite-specific.
 if [ -n "$BUILDKITE_LABEL" ] && [ $TEST_STATUS -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


scripts/dendrite_sytest.sh
```diff
@@ -61,12 +61,14 @@ fi
...
 /src/show-expected-fail-tests.sh /logs/results.tap /src/sytest-whitelist \
-    /src/sytest-blacklist > /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
+    /src/sytest-blacklist | tee /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
 
 echo >&2 "--- Copying assets"
 
 # Copy out the logs
 rsync -r --ignore-missing-args --min-size=1B -av /work/server-0 /work/server-1 /logs --include "*/" --include="*.log.*" --include="*.log" --exclude="*"
+find /logs | xargs -r chmod go+r
+find /logs -type d | xargs -r chmod go+x
```
kegsay, line 71: Why do they need to be executable? Surely read-only is enough?

scripts/dendrite_sytest.sh
```diff
@@ -61,12 +61,14 @@ fi
 
 # Check for new tests to be added to the test whitelist
 /src/show-expected-fail-tests.sh /logs/results.tap /src/sytest-whitelist \
-    /src/sytest-blacklist > /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
+    /src/sytest-blacklist | tee /work/show_expected_fail_tests_output.txt || TEST_STATUS=$?
```
kegsay, line 64: This would be very useful to have! Currently some sytest branches are `exit 1`ing because of this script and yet I have no idea why :S

scripts/dendrite_sytest.sh
reivilibre: For what it's worth, you can also use `chmod go+rX` where capital `X` only applies to directories, so you wouldn't need to do this in two separate steps.

tommie: @kegsay the directories need +rx, as per POSIX listing access.

@reivilibre ah, nice! Done.

kegsay (approved):

PR #1165, opened 2021-11-01
Capture the haproxy logs in worker mode

richvdh:
It can be quite useful to see the haproxy logs, so let's capture them. Annoyingly, this means running our own syslog daemon...

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -6,6 +6,7 @@ use 5.010;
 use base qw( SyTest::Homeserver );
 
 use Carp;
+use Socket qw( pack_sockaddr_un );
 
 use Future::Utils qw( try_repeat );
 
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -1189,15 +1190,37 @@ sub generate_haproxy_config
...
+   # create a syslog listener on a unix pipe, which will write to the logfile.
+   my $socket = IO::Async::Socket->new(
+      on_recv => sub {
+         my ( undef, $dgram, $addr ) = @_;
+         # syslog messages have some basic format, but
+         # they are readable enough for our purposes.
+         syswrite( $log_fh, $dgram );
+      },
+   );
+   my $log_sock = $self->{hs_dir} . "/log.sock";
```
anoadragon453, line 1207:
```suggestion
   my $log_sock = $self->{hs_dir} . "/haproxy_log.sock";
```

As this seems specific to HAProxy, we may want to name it as such?

anoadragon453 (approved):
Looks cautiously good to me otherwise.

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -6,6 +6,7 @@ use 5.010;
 use base qw( SyTest::Homeserver );
 
 use Carp;
+use Socket qw( pack_sockaddr_un );
```
DMRobertson, line 9: Is this from a CPAN module? I can't see to find it when using the docker container.

Issue #1164, opened 2021-11-01
Suspected flakiness in `New federated private chats get full presence information (SYN-115)`


DMRobertson:
https://github.com/matrix-org/synapse/runs/4067253219?check_suite_focus=true had this test fail in 1/6 sytest runs.

I wondered if the changes in #1160 had unintentionally introduced some kind of flakey problem. But the failure in that link does not arise from running sytest on synapse with workers (the usual suspect for for flakiness). That was a surprise.

Raising this so we can track any future sightings.

DMRobertson:
https://github.com/matrix-org/sytest/runs/4091934296?check_suite_focus=true

anoadragon453:
https://github.com/matrix-org/synapse/runs/4224174216?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/synapse/runs/4239643028?check_suite_focus=true

richvdh:
per matrix-org/synapse#1663, this seems to have been known for a long time, but I'm not sure why it suddenly flared up again?

anoadragon453:
https://github.com/matrix-org/synapse/runs/4870079986?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/synapse/pull/12407

DMRobertson:
https://github.com/matrix-org/synapse/runs/5986096002?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/synapse/runs/6078594145?check_suite_focus=true

babolivier:
https://github.com/matrix-org/synapse/runs/7444165977?check_suite_focus=true

babolivier:
https://github.com/matrix-org/synapse/runs/8112053299?check_suite_focus=true

Issue #1044, opened 2021-05-14
`Can paginate public room list` is flaky


richvdh:
https://github.com/matrix-org/synapse/pull/9981/checks?check_run_id=2582973111

richvdh:
and https://buildkite.com/matrix-dot-org/synapse/builds/17339#650a0408-a6ac-4d1b-8823-b26e43d0a379

clokep:
And https://github.com/matrix-org/synapse/pull/10574/checks?check_run_id=3293529951

DMRobertson:
This showed up twice today.

https://github.com/matrix-org/sytest/runs/4091934326?check_suite_focus=true
https://github.com/matrix-org/sytest/runs/4091934326?check_suite_focus=true

both in #1116. I couldn't see any changes there or in the linked Synapse branch which would explain this.

ShadowJonathan:
Another run in https://github.com/matrix-org/synapse/runs/4447433909?check_suite_focus=true

clokep:
Seen again: https://github.com/matrix-org/synapse/runs/4805751717

clokep:
https://github.com/matrix-org/synapse/runs/5160171515?check_suite_focus=true

clokep:
https://github.com/matrix-org/synapse/runs/5199772896?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/synapse/runs/6076311915?check_suite_focus=true

clokep:
And https://github.com/matrix-org/synapse/runs/6287201691?check_suite_focus=true

DMRobertson:
[Can paginate public room list](https://github.com/matrix-org/synapse/runs/6346842207?check_suite_focus=true)

DMRobertson:
https://github.com/matrix-org/synapse/runs/6899135826?check_suite_focus=true

richvdh:
https://github.com/matrix-org/synapse/runs/6932637886?check_suite_focus=true

richvdh:
Pretty sure this is just timing out. It creates 10 rooms, which takes 10 seconds, which is the default test timeout.

See also https://github.com/matrix-org/synapse/issues/6068

PR #1166, opened 2021-11-03
Check that master key is returned when querying device for first time

erikjohnston:
See https://github.com/matrix-org/synapse/pull/11234


anoadragon453:
There seem to be a mix of tabs vs. spaces going on here.

erikjohnston:
> There seem to be a mix of tabs vs. spaces going on here.

WHY WOULD YOU DO THIS TO ME VIM WHY :sob:

lib/SyTest/Homeserver.pm
```diff
@@ -241,7 +241,7 @@ sub configure_logger
 
       loggers => {
          synapse => {
-            level => "INFO"
```
DMRobertson, line 244: Was this just for temporary debugging?

DMRobertson:
https://github.com/matrix-org/sytest/runs/4091934296?check_suite_focus=true Looks like #1164.
https://github.com/matrix-org/sytest/runs/4091934326?check_suite_focus=true and https://github.com/matrix-org/sytest/runs/4091934384?check_suite_focus=true look like #1044. Does your branch contain the change from #1154 ?

No idea what happened in https://github.com/matrix-org/sytest/runs/4091934438?check_suite_focus=true

lib/SyTest/Homeserver.pm
erikjohnston: Oh blast, I carefully didn't add this, and then clearly did a `git commit -a`.

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -652,6 +652,21 @@
...
+         my ( $content ) = @_;
+
+         log_if_fail "key query content1", $content;
+
+         # Check that we do in fact see the master key when querying the
+         # devices.
+         assert_json_keys( $content->{master_keys}, $user2_id );
+         assert_json_keys( $content->{master_keys}->{$user2_id}, "keys");
+         assert_json_keys( $content->{master_keys}->{$user2_id}{keys},
+            "ed25519:nqOvzeuGWT/sRx3h7+MHoInYj3Uk2LD/unI9kDYcHwk");
```
DMRobertson, line 667: Should we also expect to see this in the second call to matrix_get_e2e_keys on +685?

DMRobertson (commented):
Seems reasonable; one question

tests/41end-to-end-keys/08-cross-signing.pl
erikjohnston: Yes we should. I suppose its worth checking that its still there

DMRobertson (approved):
Thanks!

PR #1168, opened 2021-11-16
Fix reference to `unpack_sockaddr_un`

richvdh:
Hopefully, fixes:

```
WARN: Error starting server-0: Undefined subroutine &main::pack_sockaddr_un called at lib/SyTest/Homeserver/Synapse.pm line 1208.
```

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -1205,7 +1205,7 @@ sub generate_haproxy_config
       },
    );
    my $log_sock = $self->{hs_dir} . "/haproxy_log.sock";
-   my $sockaddr = ::pack_sockaddr_un( $log_sock );
+   my $sockaddr = Socket::pack_sockaddr_un( $log_sock );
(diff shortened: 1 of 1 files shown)


richvdh:
(seems to be a problem introduced by #1165. Unclear why it works for some and not others.)

DMRobertson:
Works for me. With this change I can now run tests with the sytest docker container. For completeness, I was using

```
docker run --rm -v /home/dmr/workspace/synapse:/src:ro -v /home/dmr/workspace/sytest/:/sytest -v /tmp/logs:/logs -e POSTGRES=1 -e WORKERS=1 -e MULTI_POSTGRES=1 matrixdotorg/sytest-synapse:testing --torture-replication 50 -s tests/41end-to-end-keys/08-cross-signing.pl
```

DMRobertson (approved):
No idea why I'm seeing this but you weren't. Maybe your perl version is different to the perl version in the container I'm using? :shrug:

richvdh:
don't know don't care lalalala :hear_no_evil:

PR #1169, opened 2021-11-17
Attempt to deflake `uploading signed devices gets propagated over federation`

### LINKED ISSUE #1167: `uploading signed devices gets propagated over federation` is flaky
https://github.com/matrix-org/synapse/runs/4213207252?check_suite_focus=true

DMRobertson:
Fixes #1167 hopefully.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -462,8 +462,8 @@ sub retry_until_success(&%)
 # value. If the block fails entirely then it aborts, does not retry.
 sub repeat_until_true(&)
 {
-  warnings::warnif("deprecated",
-     "repeat_until_true is deprecated, use retry_until_success instead");
+   warnings::warnif( "deprecated",
+      "repeat_until_true is deprecated, use retry_until_success instead" );
    my ( $code ) = @_;
 
    my $delay = 0.1;
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -454,10 +454,16 @@ sub retry_until_success(&%)
    } until => sub { $iter > $max_iter || !$_[0]->failure };
 }
 
+# /!\ You probably DON'T want to use repeat_until true /!\
+# It means that genuine test failures get turned into test timeouts.
+# Prefer retry_until_success, which limits the number of retries.
+#
 # Another wrapper which repeats (with delay) until the block returns a true
 # value. If the block fails entirely then it aborts, does not retry.
 sub repeat_until_true(&)
 {
+  warnings::warnif("deprecated",
+     "repeat_until_true is deprecated, use retry_until_success instead");
    my ( $code ) = @_;
 
    my $delay = 0.1;
--- a/tests/41end-to-end-keys/08-cross-signing.pl
+++ b/tests/41end-to-end-keys/08-cross-signing.pl
@@ -686,9 +686,9 @@ test "uploading signed devices gets propagated over federation",
             # Wait until user1 sees signatures uploaded by user2.
             # We repeat_until_true because this is sensitive to replication delays.
             matrix_get_e2e_keys( $user1, $user2_id )
-            ->then(sub {
+            ->then( sub {
                my ( $content ) = @_;
-               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"}) {
+               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"} ) {
                   return $content;
                }
             });
--- a/tests/41end-to-end-keys/08-cross-signing.pl
+++ b/tests/41end-to-end-keys/08-cross-signing.pl
@@ -680,8 +680,6 @@ test "uploading signed devices gets propagated over federation",
              }
          } );
       })->then( sub {
-         sync_until_user_in_device_list( $user1, $user2 );
-      })->then( sub {
          retry_until_success {
             # Wait until user1 sees signatures uploaded by user2.
             # We retry_until_success because this is sensitive to replication delays.
--- a/tests/41end-to-end-keys/08-cross-signing.pl
+++ b/tests/41end-to-end-keys/08-cross-signing.pl
@@ -682,21 +682,21 @@ test "uploading signed devices gets propagated over federation",
       })->then( sub {
          sync_until_user_in_device_list( $user1, $user2 );
       })->then( sub {
-         repeat_until_true {
+         retry_until_success {
             # Wait until user1 sees signatures uploaded by user2.
-            # We repeat_until_true because this is sensitive to replication delays.
-            matrix_get_e2e_keys( $user1, $user2_id )
-            ->then( sub {
+            # We retry_until_success because this is sensitive to replication delays.
+            matrix_get_e2e_keys( $user1, $user2_id )->then( sub {
                my ( $content ) = @_;
-               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"} ) {
-                  return $content;
-               }
+               log_if_fail "key query content2", $content;
+               $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"}
+                  or die "No 'signatures' key present";
+               Future->done( $content );
             });
          };
       })->then( sub {
          my ( $content ) = @_;
 
-         log_if_fail "key query content2", $content;
+         log_if_fail "key query content3", $content;
 
          # Check that fetching the devices again returns the new signature
          assert_json_keys( $content->{device_keys}->{$user2_id}->{$user2_device}, "signatures" );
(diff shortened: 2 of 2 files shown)


tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,7 +682,17 @@
...
       })->then( sub {
-         matrix_get_e2e_keys( $user1, $user2_id );
+         repeat_until_true {
+            # Wait until user1 sees signatures uploaded by user2.
+            # We repeat_until_true because this is sensitive to replication delays.
+            matrix_get_e2e_keys( $user1, $user2_id )
+            ->then(sub {
+               my ( $content ) = @_;
+               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"}) {
+                  return $content;
```
anoadragon453, line 692: From what I'm seeing of other instances of `repeat_until_true`, one typically calls `Future->done( <true || false> )` to retry or complete the loop. Does `return` do the same thing here?

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,7 +682,17 @@
       })->then( sub {
          sync_until_user_in_device_list( $user1, $user2 );
       })->then( sub {
-         matrix_get_e2e_keys( $user1, $user2_id );
+         repeat_until_true {
+            # Wait until user1 sees signatures uploaded by user2.
+            # We repeat_until_true because this is sensitive to replication delays.
+            matrix_get_e2e_keys( $user1, $user2_id )
+            ->then(sub {
```
anoadragon453, line 689:
```suggestion
            ->then( sub {
```

looks to be our usual style.

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,7 +682,17 @@
...
+               my ( $content ) = @_;
+               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"}) {
```
anoadragon453, line 691:
```suggestion
               if ( exists $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"} ) {
```

likewise

anoadragon453 (dismissed):
With my non-perl hat on.

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,7 +682,17 @@
       })->then( sub {
          sync_until_user_in_device_list( $user1, $user2 );
       })->then( sub {
-         matrix_get_e2e_keys( $user1, $user2_id );
+         repeat_until_true {
```
richvdh, line 685: generally, I see `repeat_until_true` as an antipattern, because it means that genuine test failures get turned into test timeouts.

Prefer `retry_until_success`.

tests/41end-to-end-keys/08-cross-signing.pl
DMRobertson: > From what I'm seeing of other instances of `repeat_until_true`, one typically calls `Future->done( <true || false> )` to retry or complete the loop. Does `return` do the same thing here?

I think those examples will probably have `Future->done( <true || false> )` as their last expression---which is a Rust-like implicit return. Quoting https://perldoc.perl.org/perlsub :

> If no`return` is found and if the last statement is an expression, its value is returned.

I think what I actually wanted to use here was `retry_until_success`, which has an upper bound on the number of retry attempts.

DMRobertson: In a little more detail:

- repeat_until_true repeatedly calls a function `f` which returns a future.
- If `f` returns a future that is `done`, We inspect its encapsulated value. If that value is true, we stop the loop.
- Otherwise we continue.
So the call to `Future->done( <true || false> )` that's special; it's the fact that we're returning it. (Think of it as a Rustic `Result<bool, Error>` holding `true` or `false`---it's just a lump of data)

anoadragon453: I see, and yes I forgot about that Perl quirk - thanks!

By writing `return $content;`, are we returning a `Future` though? To me it just looks like we're returning a dictionary?

richvdh: I'm with anoa on this - I thought that functions passed to `then` had to return a `Future` rather than an immediate value (whether that be by calling another `Future`-returning function, or by creating a completed `Future` with `Future->done()`).

It's possible that requirement has been relaxed, however.

richvdh: (and I don't think this is specific to `repeat_until_true` or `retry_until_success`).

DMRobertson: I think you're both right. Quoting https://metacpan.org/pod/Future#then:

Returns a new sequencing Future that runs the code if the first succeeds. Once $f1 succeeds the code reference will be invoked and is passed the list of results. **It should return a future, $f2.** Once $f2 completes the sequence future will then be marked as complete with whatever result $f2 gave. If $f1 fails then the sequence future will immediately fail with the same failure and the code will not be invoked.

So I have no idea how this worked.

In any case, c820c95ef4f968a20a2390e544b28842e91d188c makes sure that the argument to `then` returns a `Future` or `die`s.

richvdh: a bit further up:

> Since version 0.45: if a non-future result is returned it will be wrapped in a new immediate Future instance.

DMRobertson: /me gains a new appreciation for `python -m this | grep obvious`

run-tests.pl
```diff
@@ -454,10 +454,16 @@ (&%)
...
+# /!\ You probably DON'T want to use repeat_until true /!\
+# It means that genuine test failures get turned into test timeouts.
+# Prefer retry_until_success, which limits the number of retries.
+#
 # Another wrapper which repeats (with delay) until the block returns a true
 # value. If the block fails entirely then it aborts, does not retry.
 sub repeat_until_true(&)
 {
+  warnings::warnif("deprecated",
+     "repeat_until_true is deprecated, use retry_until_success instead");
```
richvdh, line 466:
```suggestion
   warnings::warnif( "deprecated",
      "repeat_until_true is deprecated, use retry_until_success instead" );
```

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,11 +682,21 @@
...
          sync_until_user_in_device_list( $user1, $user2 );
       })->then( sub {
-         matrix_get_e2e_keys( $user1, $user2_id );
+         retry_until_success {
+            # Wait until user1 sees signatures uploaded by user2.
+            # We retry_until_success because this is sensitive to replication delays.
+            matrix_get_e2e_keys( $user1, $user2_id )->then( sub {
+               my ( $content ) = @_;
+               log_if_fail "key query content2", $content;
+               $content->{device_keys}->{$user2_id}->{$user2_device}->{"signatures"}
```
richvdh, line 691: what you wrote here is entirely valid, but the `->` is implied between two subscripts (https://perldoc.perl.org/perlreftut#Arrow-Rule), and convention is to drop them.

```suggestion
               $content->{device_keys}{$user2_id}{$user2_device}{"signatures"}
```

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,11 +682,21 @@
...
+                  or die "No 'signatures' key present";
```
richvdh, line 692: is it sufficient to wait for *anything* at `$content->{device_keys}{$user2_id}{$user2_device}`, or do we need to wait for there to be a `signatures` element?

tests/41end-to-end-keys/08-cross-signing.pl
```diff
@@ -682,11 +682,21 @@
       })->then( sub {
          sync_until_user_in_device_list( $user1, $user2 );
```
richvdh, line 683: hrm. I believe the intention of this line is to wait for the device info to become available. Do you know why it's insufficient? (And if there's a good reason, we may as well remove it.)

tests/41end-to-end-keys/08-cross-signing.pl
DMRobertson: As I understand it, we need to wait for the signatures in particular. 

On `server1:`
1. user2 joins a room
2. user2 uploads signatures

On `server0`, user1 syncs until they see user2's device. This is racey: the sync may complete before the signatures have uploaded, propagated over federation and then over replication.

tests/41end-to-end-keys/08-cross-signing.pl
DMRobertson: I think
- this is explained by https://github.com/matrix-org/sytest/pull/1169#discussion_r754194531.
- this `then` call is therefore redundant.

tests/41end-to-end-keys/08-cross-signing.pl
richvdh: right. please could you add these words to a comment in the code.

richvdh (approved):
generally lgtm, if you can expand the comment about what we're waiting for and why.

PR #1170, opened 2021-11-22
Check both `SYTEST_BRANCH` and `BUILDKITE_BRANCH` in Docker bootstrap

neilalexander:
PR #1123 broke the `Trying to get same-named sytest branch` stage for Dendrite CI which is still in Buildkite.

### DIFF
--- a/docker/bootstrap.sh
+++ b/docker/bootstrap.sh
@@ -17,6 +17,8 @@ else
     # Synapse/Dendrite branch we're running
     if [ -n "$SYTEST_BRANCH" ]; then
         branch_name="$SYTEST_BRANCH"
+    elif [ -n "$BUILDKITE_BRANCH" ]; then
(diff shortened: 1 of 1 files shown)


anoadragon453 (commented):
This works, though is it not possible to set `SYTEST_BRANCH` to `$BUILDKITE_BRANCH` in [dendrite's pipeline.yml](https://github.com/matrix-org/pipelines/blob/master/dendrite/pipeline.yml) instead?

neilalexander:
> though is it not possible to set SYTEST_BRANCH to $BUILDKITE_BRANCH in dendrite's pipeline.yml instead?

Possibly, but I'm having trouble getting the pipeline to run under `bk local run` to prove it one way or the other. We know this way works.

kegsay (approved):

PR #1172, opened 2021-11-28
Add push_server.disable_tls_validation to Dendrite configuration

tommie:
This disables the TLS certificate check for Push gateways. Sytest uses HTTPS, but with a self-signed certificate. Synapse already has an option like this.

Dendrite-side must be committed first: https://github.com/matrix-org/dendrite/pull/2061

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -105,7 +105,7 @@ sub _get_config
          kafka => {
             use_naffka => $JSON::true,
             naffka_database => {
-               connection_string => 
+               connection_string =>
                    ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
                    "file:$self->{hs_dir}/naffka.db" : $db_uri,
             },
@@ -122,7 +122,7 @@ sub _get_config
 
       app_service_api => {
          database => {
-            connection_string => 
+            connection_string =>
                 ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
                 "file:$self->{hs_dir}/appservice_api.db" : $db_uri,
          },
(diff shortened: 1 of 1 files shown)


tommie:
Reviewer: @kegsay

lib/SyTest/Homeserver/Dendrite.pm
```diff
@@ -101,6 +101,7 @@ sub _get_config
          server_name => $self->server_name,
          private_key => $self->{paths}{matrix_key},
          presence_enabled => $JSON::true,
+         http_client_tls_insecure_skip_verify_for_testing => $JSON::true,
```
kegsay, line 104: This is in the wrong place. It's in global here but the dendrite pr adds this flag to the pushers config.

tommie: Yepp, revised.

tommie:
See https://github.com/matrix-org/dendrite/pull/1842#issuecomment-985752226 for the commit ordering relative Dendrite.

tommie:
@PiotrKozimor raised the question if this can be merged before the config option has been added to Dendrite. I've assumed that Dendrite will fail if we try, but maybe that's not the case?

tommie:
... and @neilalexander just answered it's okay on the Dendrite 1842.

clokep:
I removed the review on synapse-core since I don't think it is necessary here. Please shout if you'd like us to take a look though!

neilalexander (approved):

lib/SyTest/Homeserver/Dendrite.pm
```diff
@@ -195,37 +195,41 @@ sub _get_config
...
          database => {
-            connection_string => 
+            connection_string =>
                ( ! defined $ENV{'POSTGRES'} || $ENV{'POSTGRES'} == '0') ?
                "file:$self->{hs_dir}/mscs.db" : $db_uri,
          },
          mscs => ["msc2836", "msc2946", "msc2444", "msc2753"],
       },
 
+      push_server => {
```
PiotrKozimor, line 205: @tommie I just realized that we already have `push_server` section below

neilalexander: I've fixed that. Good catch.

tommie: Sorry about that. Thank you!

Issue #1173, opened 2021-11-29
When bootstrapping, check the user's namespace first with their own branch before falling back on matrix-org


ShadowJonathan:
https://github.com/matrix-org/sytest/blob/cfbf7430a66d6bcd06b151c7c477ecd2566e1e8b/docker/bootstrap.sh#L32

Edit that to first check `user_repo/sytest@user_branch`, before going to `matrix-org/sytest@user_branch` and then `matrix-org/sytest@develop`

richvdh:
I have a feeling we avoided this because "security" (something to do with untrusted code in CI). But possibly that was pre-GHA, and is no longer a problem?

*handwave*. I can't really remember.

ShadowJonathan:
Maybe it could still become a security concern if/when matrix.org adds their own self-hosted runners to the GHA pool, though maybe that could be re-routed for untrusted PRs?

PR #1174, opened 2021-12-06
Add `worker_to_update_user_directory config` option

ShadowJonathan:
Works together with https://github.com/matrix-org/synapse/pull/11450

I chose to do this over https://github.com/matrix-org/sytest/issues/1173, because this is just less hassle.

The issue stated in 1173 still stands, but this is just an easier fix for now.

This update is backwards-compatible with other synapse branches and versions, as those just ignore extra config keys.

`update_user_directory` is not removed for backwards compat, and likely will stay this way.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -263,10 +263,12 @@ sub start
         start_pushers         => ( not $self->{workers} ),
         notify_appservices    => ( not $self->{workers} ),
         send_federation       => ( not $self->{workers} ),
-        update_user_directory => ( not $self->{workers} ),
         enable_media_repo     => ( not $self->{workers} ),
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -266,6 +266,7 @@ sub start
         update_user_directory => ( not $self->{workers} ),
         enable_media_repo     => ( not $self->{workers} ),
         run_background_tasks_on  => ( $self->{workers} ? "background_worker1" : "master" ),
+        worker_to_update_user_directory  => ( $self->{workers} ? "user_dir" : "null" ),
```
squahtx, line 269: It may be useful to add a comment about keeping `update_user_directory` for backwards compatibility.
Otherwise an enthusiastic developer may try to remove it in the future.

squahtx (approved):
lgtm

PR #1177, opened 2021-12-08
Update user directory variable to right type and name

ShadowJonathan:
This fixes a small problem introduced by https://github.com/matrix-org/sytest/pull/1174, i thought that `"null"` would become a null yaml value, but it seems not.

This tries ~~`undef`~~`()` instead.

---

After some review over in https://github.com/matrix-org/synapse/pull/11450 im also using this PR to change over the config variable name to `update_user_directory_on`

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -265,7 +265,7 @@ sub start
         send_federation       => ( not $self->{workers} ),
         enable_media_repo     => ( not $self->{workers} ),
         run_background_tasks_on  => ( $self->{workers} ? "background_worker1" : "master" ),
-        worker_to_update_user_directory  => ( $self->{workers} ? "user_dir" : "null" ),
+        worker_to_update_user_directory  => ( $self->{workers} ? "user_dir" : undef ),
         # update_user_directory is kept for backwards compatibility,
         # worker_to_update_user_directory is prioritized before this option.
         update_user_directory => ( not $self->{workers} ),
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -266,7 +266,7 @@ sub start
         enable_media_repo     => ( not $self->{workers} ),
         run_background_tasks_on  => ( $self->{workers} ? "background_worker1" : "master" ),
         $self->{workers} ? (
-            worker_to_update_user_directory  => "user_dir",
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -265,7 +265,7 @@ sub start
         send_federation       => ( not $self->{workers} ),
         enable_media_repo     => ( not $self->{workers} ),
         run_background_tasks_on  => ( $self->{workers} ? "background_worker1" : "master" ),
-        worker_to_update_user_directory  => ( $self->{workers} ? "user_dir" : "null" ),
+        worker_to_update_user_directory  => ( $self->{workers} ? "user_dir" : undef ),
```
richvdh, line 268: wouldn't it be better not to set it at all, so that synapse uses its default?

```suggestion
        $self->{workers} ? (
            worker_to_update_user_directory  => "user_dir",
        ) : (),
```

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -265,7 +265,7 @@ sub start
...
         # update_user_directory is kept for backwards compatibility,
         # worker_to_update_user_directory is prioritized before this option.
         update_user_directory => ( not $self->{workers} ),
```
richvdh, line 271: presumably once https://github.com/matrix-org/synapse/pull/11450, this will be removed?

ShadowJonathan: Not directly, i'd like to wait until most/all branches have been updated with the develop branch, else you'll get unrelated sytest errors :/

I'll make a reminder to revisit this in a few months or so, then enumerate all open PRs, and see if they include the commit in their branch or not. If it's not a lot of PRs left, ill just boop them with "hey, please update to `develop`", and throw a PR in here which removes this option.

lib/SyTest/Homeserver/Synapse.pm
ShadowJonathan: Yes, this'd be better. Thanks, i didn't know this was possible in perl.

ShadowJonathan:
I'm blocking this to address https://github.com/matrix-org/synapse/pull/11450#discussion_r765658719, so i dont throw in a third PR after the fact, to make a single change, stand by.

richvdh (approved):
Normally I'd say that we should wait for the synapse PR to be a bit closer to being ready, but that ship already sailed with #1174, so this isn't making anything worse.

Issue #1114, opened 2021-08-09
install-deps.pl fails to build libsodium with gcc 11.2.1


DMRobertson:
**Describe the bug**
Unable to install sytest's perl dependencies following step 3 in README.md. 

**To Reproduce**

1. Clone sytest repo. Setup perl env vars as in the readme.
2. `./install-deps.pl`
3. This fails when installing Alien::Sodium. There's a complaint when building libsodium about alignment.

```
make[3]: Entering directory '/home/dmr/.local/share/.cpan/build/Alien-Sodium-1.0.8.0-2/_alien/libsodium-1.0.8/src/libsodium'

[...]

  CC       crypto_generichash/blake2/ref/libsodium_la-blake2b-compress-ref.lo
In file included from crypto_generichash/blake2/ref/blake2b-compress-ref.c:5:
crypto_generichash/blake2/ref/blake2.h:118:5: error: size of array element is not a multiple of its alignment
  118 |     blake2s_state S[8][1];
      |     ^~~~~~~~~~~~~
crypto_generichash/blake2/ref/blake2.h:119:5: error: size of array element is not a multiple of its alignment
  119 |     blake2s_state R[1];
      |     ^~~~~~~~~~~~~
crypto_generichash/blake2/ref/blake2.h:126:5: error: size of array element is not a multiple of its alignment
  126 |     blake2b_state S[4][1];
      |     ^~~~~~~~~~~~~
crypto_generichash/blake2/ref/blake2.h:127:5: error: size of array element is not a multiple of its alignment
  127 |     blake2b_state R[1];
      |     ^~~~~~~~~~~~~
make[3]: *** [Makefile:2284: crypto_generichash/blake2/ref/libsodium_la-blake2b-compress-ref.lo] Error 1
```

<details>
<summary>Gory details</summary>

```
dmr on titan in sytest on  develop via 🐪 v5.32.1 
2021-08-09 16:32:06 ✔  $ ./install-deps.pl 

**** install-deps.pl: Installing Alien::Sodium ****
CPAN: Storable loaded ok (v3.21)
Reading '/home/dmr/.local/share/.cpan/Metadata'
  Database was generated on Mon, 09 Aug 2021 14:41:03 GMT
CPAN: Digest::SHA loaded ok (v6.02)
CPAN: Compress::Zlib loaded ok (v2.102)
[...]

PR #1179, opened 2021-12-09
Mark `testing` image as experimental

neilalexander:
This marks the `testing` image as experimental and allows the builds of the stable images to continue even if the experimental ones fail.

It's nice to know when something in `testing` will break CI for a future Debian version, but it probably shouldn't stop us from successfully building existing images based on the stable distributions.

It also removes a tag from the Dendrite build that we don't care about anymore.

(Related to #1114)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,8 +14,10 @@ jobs:
   build-sytest-images:
     name: "Build sytest:${{ matrix.tag }}"
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


richvdh (changes requested):
it looks like we don't currently build on debian stable (only oldstable and testing)?

I think these images get used in the synapse pipeline. I'm not keen on us using outdated images without noticing there. Please could someone correlate to that pipeline?

richvdh:
I'm going to go ahead and close this, since it doesn't seem to be going anywhere.

PR #1181, opened 2021-12-21
Drop bionic from docker builds and add focal

H-Shay:
Related to #11552, drops bionic from the docker builds which uses EOL python 3.6 and add focal, which uses python 3.9.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,8 +17,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - base_image: ubuntu:bionic
-            tag: bionic
+          - base_image: ubuntu:focal
+            tag: focal
           - base_image: debian:buster
             tag: buster
           - base_image: debian:testing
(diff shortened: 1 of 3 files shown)


callahad:
> which uses python 3.9.

I think Focal is actually on 3.8 by default: https://packages.ubuntu.com/focal/python3

callahad:
Curious chicken/egg problem with CI: it wants to pull the images that you're naming in this PR, but since the PR hasn't yet been merged, the images haven't yet been built.

callahad:
I suspect we'll need to somewhat blindly merge, and then submit a dummy follow-up PR to make sure it works.

H-Shay:
well I could do a manual push and build but there is an issue with the `testing` image being built which is causing problems.

H-Shay:
This is now unblocked.

DMRobertson:
> Well I could do a manual push and build but there is an issue with the `testing` image being built which is causing problems: see #1114

I think ops are keen for us to avoid personally publishing artefacts where possible.

DMRobertson (commented):
I don't quite understand why CI is now passing. It manages to download the image `docker.io/matrixdotorg/sytest-synapse:focal`, but that shouldn't have been published until this gets merged(?)

DMRobertson (approved):
Oh, I see: you manually [invoked](https://github.com/matrix-org/sytest/actions/runs/1659721512) the docker workflow on this PR via github's UI.

I can see [here](https://hub.docker.com/r/matrixdotorg/sytest-synapse/tags) that Bionic was pushed at 5:30pm but Focal, Buster and Testing were pushed in a batch at 6:20. I think that, plus green CI is evidence enough that this is working!

PR #1184, opened 2022-01-25
Add support for testing on Python 3.10

H-Shay:
As the title states. One of the tasks outlined here:  #11780. Builds a new Docker image with debian:bookworm as the base, pushes it to DockerHub and adds the new image to CI for testing.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,6 +23,8 @@ jobs:
             tag: buster
           - base_image: debian:testing
             tag: testing
+          - base_image: debian:bookworm
+            tag: bookworm
     steps:
       - name: Set up Docker Buildx
         id: buildx
@@ -60,6 +62,9 @@ jobs:
           - sytest_image_tag: buster
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:buster"
+          - sytest_image_tag: bookworm
+            dockerfile: synapse
+            tags: "matrixdotorg/sytest-synapse:bookworm"
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -32,6 +32,15 @@ jobs:
             postgres: postgres
             workers: workers
 
+          - label: Py 3.10, PG 14, Monolith
+            sytest-tag: bookworm
+            postgres: postgres
+
+          - label: Py 3.10, PG 14, Workers
+            sytest-tag: bookworm
+            postgres: postgres
+            workers: workers
+
           - label: debian-testing, Monolith
             sytest-tag: testing
             postgres: postgres


DMRobertson:
For the sytest side of this, see https://github.com/matrix-org/sytest/pull/1184

DMRobertson (changes requested):
https://packages.debian.org/bookworm/python3 seems to suggest that bookworm includes Python 3.9 by default. We'd have to explicitly opt-in to using 3.10 I think, by apt-installing [this](https://packages.debian.org/sid/source/python3.10).

Does that sound right @richvdh ?

richvdh:
Yup, looks that way. Not only that, we need to use it explicitly:

```
rav@fred:~$ docker run --rm -it debian:bookworm
root@294ab397ee09:/# apt update -qq
1 package can be upgraded. Run 'apt list --upgradable' to see it.
root@294ab397ee09:/# apt install -y python3
...
root@294ab397ee09:/# python3 --version
Python 3.9.10

root@294ab397ee09:/# apt install -y python3.10
...
root@294ab397ee09:/# python3 --version
Python 3.9.10
root@294ab397ee09:/# python3.10 --version
Python 3.10.2
```

... so I think we need to change [`synapse.Dockerfile`](https://github.com/matrix-org/sytest/blob/167d2ef3bfb6e8d4068339e973aa58adfdc990d0/docker/synapse.Dockerfile#L13) to use `python3.10` explicitly, as well as installing the right package.

Not sure if this is best done via docker build args (https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) or with a completely separate dockerfile.

H-Shay:
I took a stab at using a separate dockerfile for building the 3.10 image and using it explicitly. I am none too familiar with docker but I think I succeeded, very open to feedback if I am wrong or there's a more elegant way to do this!

.github/workflows/docker.yml
```diff
@@ -90,3 +92,32 @@ jobs:
...
+
+      - name: Build and push 3.10
+        uses: docker/build-push-action@v2
+        with:
+          pull: true
+          push: true
+          labels: "gitsha1=${{ github.sha }}"
+          file: docker/synapse3_10.Dockerfile
+          build-args: "SYTEST_IMAGE_TAG=bookworm"
+          tags: "matrixdotorg/sytest-synapse:bookworm"
```
richvdh, line 123: I wonder if we should give this image a different tag. It's not stock bookworm, after all. Maybe better to call it `matrixdotorg/sytest-synapse:bookworm-python3.10` or something.

richvdh, line 123: I think you can do this more easily by adding an entry to the `matrix` for the `build-dependent-images` job ?

docker/synapse3_10.Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG SYTEST_IMAGE_TAG=bookworm
```
richvdh, line 1: we could do with a comment here. What does this file do, how is it different from `synapse.Dockerfile`, etc?

richvdh, line 1: ARG defines an argument that can be set at build time with a parameter to `docker build` (see https://docs.docker.com/engine/reference/builder/#arg). As things currently stand, `SYTEST_IMAGE_TAG` is unused in your dockerfile, so this is rendundant.

That said, I don't think we *need* to hardcode the base image on line 2: currently the only sensible value is `bookworm`, but that might change in the future.

docker/synapse3_10.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+FROM matrixdotorg/sytest:bookworm
+
+RUN apt-get -qq update && apt-get -qq install -y \
+    python3.10 python3.10-dev python3.10-venv eatmydata \
+    redis-server
```
richvdh, line 6: one thought that occurs to me: if we add an `ARG PYTHON_VERSION=python3` to `synapse.Dockerfile`, and replace each reference to `python3` with `${PYTHON_VERSION}` ... then we can replace this whole dockerfile with a build arg of `PYTHON_VERSION=python3.10`.

wdyt?

richvdh (changes requested):

.github/workflows/docker.yml
H-Shay: Yes, that makes sense!

docker/synapse3_10.Dockerfile
H-Shay: I think that's a more sensible approach than the one I took-I'll give using build args a shot!

H-Shay:
So I have managed to ditch the custom dockerfile for python 3.10 and achieve the same result using a build arg passed to the generic `synapse` dockerfile. However, I was unable to figure out how to add this particular build to the strategy matrix: in particular, I could not figure out how to pass an extra build arg for python 3.10 in that one case. I am sure there is something very simple I am missing, but I banged my head against the wall long enough that's time to ask. Otherwise I think this should be good to go.

.github/workflows/docker.yml
```diff
@@ -57,15 +59,25 @@ jobs:
...
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
+            build_arg: "SYTEST_IMAGE_TAG=testing"
+          - sytest_image_tag: bookworm
+            dockerfile: synapse
+            tags: "matrixdotorg/sytest-synapse:bookworm-python3.10"
+            build_arg: |
+              "SYTEST_IMAGE_TAG=bookworm"
+              "PYTHON_VERSION=python3.10"
```
richvdh, line 76: I think the `"` are redundant here, because the `|` means it's already a string from YAML's point of view.

(in fact, I'm a bit surprised they work. best get rid of them I think.)

.github/workflows/docker.yml
```diff
@@ -88,5 +100,9 @@ jobs:
           push: true
           labels: "gitsha1=${{ github.sha }}"
           file: docker/${{ matrix.dockerfile }}.Dockerfile
-          build-args: "SYTEST_IMAGE_TAG=${{ matrix.sytest_image_tag }}"
+          build-args: ${{ matrix.build_arg }}
```
richvdh, line 103: maybe call the matrix variable `build_args` since there can be more than one?

.github/workflows/docker.yml
```diff
@@ -57,15 +59,25 @@ jobs:
           - sytest_image_tag: focal
```
richvdh, line 59: the old `sytest_image_tag` variables seem to be unused now. Could you get rid of them?

.github/workflows/pipeline.yml
```diff
@@ -32,6 +32,15 @@ jobs:
             postgres: postgres
             workers: workers
 
+          - label: Py 3.10, PG 14, Monolith
+            sytest-tag: bookworm
```
richvdh, line 36: I think these refer to the tag on the `matrixdotorg/sytest-synapse` image, so should be `bookworm-python3.10` if we want python 3.10?

richvdh (commented):
looks good, a few more tweaks!

.github/workflows/docker.yml
H-Shay: I believe they are still being used on line 54?:
```
  name: "Build sytest-${{ matrix.dockerfile }}:${{ matrix.sytest_image_tag }}"
```

PR #1191, opened 2022-02-21
Add POD documentation for some `SyTest::Federation` methods

richvdh:


erikjohnston (approved):

lib/SyTest/Federation/Server.pm
```diff
@@ -65,6 +65,39 @@ sub make_request
...
+match the components of the path in the requested URI.
+
+If no such handler is found, a 404 is returned. Otherwise, the handler can
+return a Future which returns one of:
+
+=over 4
+
+=item C<undef>
+
+If no value is returned, it is asumed the handler has already fully handled the request.
```
clokep, line 87: Sorry for noticing after the merge:

```suggestion
If no value is returned, it is assumed the handler has already fully handled the request.
```

PR #1192, opened 2022-02-21
Handle servers which decide to make `/state` requests

richvdh:
Some server implementations may make `/state` requests when there is a large
amount of state to sync up; add code to handle that.


richvdh:
superceded by #1223

PR #1193, opened 2022-02-21
send_join test: Omit create event from `state`

richvdh:
As of MSC3706, omitting the create event from `auth_chain` is valid (because it
should also be in `state`). But it definitely *should* be there in `state`.


richvdh:
this is needed for https://github.com/matrix-org/synapse/pull/12039

erikjohnston (approved):

PR #1195, opened 2022-02-23
Update tests to be nicer to Dendrite

S7evinK:
[Synapse](https://github.com/matrix-org/synapse/blob/6d14b3dabfe38c6ae487d0f663e294056b6cc056/synapse/handlers/room.py#L120-L133) sets the `power_level_content_override` by default, if it is not set, resulting in non-spec behavior in Dendrite with this test.
This sets the `power_level_content_override` as per the Synapse code, so Dendrite can pass this test (would timeout before, because the inviting user wasn't allowed to invite the creator of the room)

I've also tried to update some of the `repeat_until_true` calls to use `retry_until_success` instead. Also uses `await_sync_timeline_contains` instead of calls to `/events`

### DIFF
--- a/tests/42tags.pl
+++ b/tests/42tags.pl
@@ -185,6 +185,7 @@ sub check_tag_event {
 
 
 test "Tags appear in the v1 /events stream",
+   deprecated_endpoints => 1;
    requires => [ local_user_fixture( with_events => 1 ),
                  qw( can_add_tag can_remove_tag ) ],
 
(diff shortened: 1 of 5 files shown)


neilalexander (approved):

tests/42tags.pl
```diff
@@ -185,6 +185,7 @@ sub check_tag_event {
 
 
 test "Tags appear in the v1 /events stream",
+   deprecated_endpoints => 1;
```
richvdh, line 188: this needs to end with a `,`, not a `;`

PR #1196, opened 2022-02-24
Remove tap-to-junit

richvdh:
This is unused since b225a183, and moreover the installation of
`XML::Generator` seems to be broken currently (yay perl)


erikjohnston (approved):

deepbluev7:
We have been using this script to integrate with our test reports in gitlab, guess we need to look for a different solution now, which is fine, I guess. https://gitlab.com/famedly/company/backend/services/synapse/-/pipelines/477215450/test_report

PR #1197, opened 2022-02-24
Add required fields to POST /keys/upload

S7evinK:
This adds the [required](https://spec.matrix.org/v1.2/client-server-api/#post_matrixclientv3keysupload) fields to `device_keys` if they are missing. (the values probably don't make much sense)


erikjohnston:
As per https://github.com/matrix-org/synapse/issues/12081, this is failing because you're trying to upload a different key with the same key ID :slightly_smiling_face:

squahtx:
@S7evinK could you make the tests happy and also add a sign off (https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off) ?

Also is the "Fix mistake" commit related to key upload?

S7evinK:
I'm actually not sure how to fix this within Sytest. :/

Regarding the "Fix mistake": Not exactly related, but fixes an issue I introduced in https://github.com/matrix-org/sytest/pull/1195

clokep:
> Regarding the "Fix mistake": Not exactly related, but fixes an issue I introduced in #1195

This is likely best in a separate PR then!

I'm going to remove the review flag from Synapse since this:

* Needs to fix the tests (I'm not 100% sure what @erikjohnston is suggesting, but it sounds like you need to generate unique key IDs, maybe one for each test?)
* There's a conflict (probably from the path changes in #1200).

richvdh:
> Regarding the "Fix mistake": Not exactly related, but fixes an issue I introduced in https://github.com/matrix-org/sytest/pull/1195

this is fixed on `develop` already.

S7evinK:
Not going to spend time on Sytest, so closing this.

PR #1199, opened 2022-03-01
Goodbye, synapse's scripts directory

DMRobertson:
In https://github.com/matrix-org/synapse/pull/12118 I am moving scripts into synapse to make CI config neater and tidier. Update Sytest's glue scripts to no longer require the scripts directory to be present.

### DIFF
--- a/scripts/synapse_sytest.sh
+++ b/scripts/synapse_sytest.sh
@@ -122,7 +122,7 @@ fi
 if [ -d "$SYNAPSE_SOURCE" ]; then
     echo "Creating tarball from synapse source"
     tar -C "$SYNAPSE_SOURCE" -czf /tmp/synapse.tar.gz \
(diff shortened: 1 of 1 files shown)


clokep:
Are any of these scripts even used in sytest or was it just for consistency?

DMRobertson:
> Are any of these scripts even used in sytest or was it just for consistency?

I suspect just for consistency. But let's see how the sytest jobs on #12118 fair.

clokep (approved):

DMRobertson:
Many thanks Patrick!

DMRobertson:
> But let's see how the sytest jobs on [matrix-org/synapse#12118](https://github.com/matrix-org/synapse/pull/12118) fair.

They passed. I'm confident that we don't use the scripts in sytest anywhere and a quick grep shows no evidence to the contrary.

PR #1200, opened 2022-03-02
Use `/v3` endpoints

neilalexander:
This PR moves client API calls onto `/v3` endpoints instead of `/r0`, `/unstable` etc.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -1284,10 +1284,10 @@ sub generate_haproxy_map
    # haproxy. Also, the routing for GET requests below takes precedence over
    # these routes.
    my $haproxy_map = <<'EOCONFIG';
-^/_matrix/client/(v2_alpha|r0)/sync$                  synchrotron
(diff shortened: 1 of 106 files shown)


kegsay (approved):

lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -1284,10 +1284,10 @@ sub generate_haproxy_map
    # haproxy. Also, the routing for GET requests below takes precedence over
    # these routes.
    my $haproxy_map = <<'EOCONFIG';
-^/_matrix/client/(v2_alpha|r0)/sync$                  synchrotron
```
richvdh, line 1287: this stuff really needs to match what's at https://matrix-org.github.io/synapse/develop/workers.html, but keeping them in sync is a bigger topic (https://github.com/matrix-org/synapse/issues/12139). This is fine for now, though I would have taken the opportunity to push `v2_alpha` and `api/v1` into a burning pit.

richvdh (approved):
lgtm!

PR #1202, opened 2022-03-02
synctl is now an entrypoint

DMRobertson:
More generally: I'm a bit surprised that we have synapse- and dendrite- specific stuff in sytest.

https://github.com/matrix-org/synapse/pull/12140


richvdh (approved):
yay.

PR #1204, opened 2022-03-03
Ensure pushers exist before asking `/notifications` which notifications were sent

neilalexander:
Without pushers, the notifications might be dropped by some implementations.


kegsay:
Unsure if these semantics are correct. Can you get a synapse guy to chime in here?

neilalexander:
Closing and will do matrix-org/dendrite#2244 instead.

PR #1209, opened 2022-03-09
Undeprecate and rewrite `repeat_until_true`

richvdh:
`repeat_until_true` is a useful function, on occasions where you want to retry certain result values, whilst failing immediately on any exceptions.

Let's
 * extend `retry_until_success` with a few extra parameters
 * rewrite `repeat_until_true` to use it, hence gaining an iteration limit
 * Undeprecate it.

Fixes some deprecation warnings.

### DIFF
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -514,8 +514,6 @@ sub retry_until_success(&%)
 # The final result is either the result of the code block on success, or the
 # most recent failure.
 #
-# Will fail if the code block fails `max_iterations` (default 7) times.
-#
 sub repeat_until_true(&%)
 {
    my ( $code, %params ) = @_;
--- a/run-tests.pl
+++ b/run-tests.pl
@@ -414,11 +414,25 @@ sub delay
 }
 
 
-# Retries a code block until it returns a successful Future.
+# Retries a code block until it succeeds, with a limited number of retries.
 #
 # Includes a delay between each call, and logs the reason for failure.
 #
-# The code block is called with the iteration number.
+# The code block is called with the iteration number, and must return a Future.
+#
+# By default, if the resulting Future fails, the code block will be retried
+# (until the iteration count is exhausted). This behaviour can be changed by
+# setting the `retry_fails` parameter to 0, which will cause the failure to
+# propagate immediately.
+#
+# A CODE reference may be passed in the `check` parameter. If so, it is
+# called with the successful result of the Future. If it fails (ie, calls `die`),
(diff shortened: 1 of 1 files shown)


run-tests.pl
```diff
@@ -414,70 +414,121 @@ sub delay
...
 #
 # Includes a delay between each call, and logs the reason for failure.
 #
-# The code block is called with the iteration number.
+# The code block is called with the iteration number, and must return a Future.
+#
+# By default, if the resulting Future fails, the code block will be retried
+# (until the iteration count is exhausted). This behaviour can be changed by
+# setting the `retry_fails` parameter to 0, which will cause the failure to
+# propagate immediately.
```
clokep, line 426: ~~I think I'm not groking what `retry_fails` does -- it seems to skip the behavior of `retry_until_success`, which doesn't seem useful? I might just not understand failure propagation in Perl Futures well enough though.~~

From reading the code below it seems to let you skip retrying if the `Future` fails, but keep retrying if the `check` function fails?

run-tests.pl
```diff
@@ -414,70 +414,121 @@ sub delay
...
+#      ...
+#   }, max_iterations => 20,
+#      initial_delay => 0.01;
+#
+# Will fail if the code block fails `max_iterations` (default 7) times.
+#
+# The final result is either the result of the code block on success, or the
+# most recent failure.
+#
+# Will fail if the code block fails `max_iterations` (default 7) times.
```
clokep, line 517: This line is repeated

```suggestion
```

run-tests.pl
richvdh: yes, that's right.

It would be pretty silly to set `retry_fails=0` without also changing `check`, since that would mean there would be no circumstances under which we'd repeat the action. Which I think is what you meant in your ~~struckthrough~~ text.

Any suggestions for a more intuitive API here? Maybe setting `check` should change the default for `retry_fails=0`, but that feels a bit confusing to me. Or we could just have a single boolean rather than both `check` and `retry_fails`, but then it felt like having `check` be more general might be useful.

clokep: Maybe just updating the comment to explicitly say this is useful with `check`?

```suggestion
# By default, if the resulting Future fails, the code block will be retried
# (until the iteration count is exhausted). This behaviour can be changed by
# setting the `retry_fails` parameter to 0, which will cause the failure to
# propagate immediately; this is only useful when also providing a `check`
# parameter.
```

PR #1210, opened 2022-03-10
Prevent GHA workflow failing for Dendrite

richvdh:
Sytest appears to fail against Dendrite, so let's disable it for now.


kegsay (approved):

clokep (approved):
I suspect @kegsay's review is enough, but if CI is happy and Dendrite folks are happy than I'm happy. Have a ✅ !

PR #1211, opened 2022-03-12
Outlier/rejected tests for `/state` and `/state_ids`

richvdh:
Extend the tests for `/state` and `/state_ids` to check the behaviour on outliers and rejected events.

The rejected-event behaviour has always been to return a 404, but we add a test to make sure it stays that way.

The outlier behaviour is changed by https://github.com/matrix-org/synapse/pull/12087.

### DIFF
--- a/tests/50federation/00prepare.pl
+++ b/tests/50federation/00prepare.pl
@@ -136,7 +136,7 @@ I<$main::OUTBOUND_CLIENT>, which is used to send the event.
 
 I<$room> should be the L<SyTest::Federation::Room> in which to send the event.
 
-I<$user> should be a L<User> which will poll for receiving the event.
+I<$server_user> should be a L<User> which will poll for receiving the event.
 
 The remainder of the arguments (I<%fields>) are passed into
 L<SyTest::Federation::Room/create_and_insert_event>. They should include at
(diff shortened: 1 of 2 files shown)


tests/50federation/00prepare.pl
```diff
@@ -169,6 +169,135 @@ sub send_and_await_event {
...
+
+   my ( $outlier_event_Q, $outlier_event_id_Q ) = $room->create_and_insert_event(
+      type => 'm.room.member',
+      sender => $sending_user_id,
+      state_key => $sending_user_id,
+      content => { membership => 'join' },
+      auth_events => $room->make_event_refs(
+         $room->get_current_state_event( "m.room.create" ),
+         $room->get_current_state_event( "m.room.power_levels" ),
+         $room->get_current_state_event( "m.room.member", $sending_user_id ),
```
clokep, line 222: Looking at `Room->create_event`, I think these are essentially the default `auth_events` (although that seems to also specify `m.room.join_rules`). Am I missing something or are we just being specific here?

clokep (commented):
Seems reasonable, but I had one question!

tests/50federation/00prepare.pl
```diff
@@ -136,7 +136,7 @@ =head2 send_and_await_event
 
 I<$room> should be the L<SyTest::Federation::Room> in which to send the event.
 
-I<$user> should be a L<User> which will poll for receiving the event.
+I<$server_user> should be a L<User> which will poll for receiving the event.
```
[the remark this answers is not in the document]
richvdh: yes it should, thank you.

tests/50federation/00prepare.pl
richvdh: yeah, I don't know why I did this.

PR #1215, opened 2022-03-16
De-flake `Read markers appear in initial v2 /sync`

neilalexander:
It was not `await_sync`ing.


kegsay (approved):
CI has not run though.

PR #1216, opened 2022-03-20
Resolve room alias in `matrix_create_room`

gnieto:
The spec does not require `/createRoom` request to return room aliases
(https://spec.matrix.org/v1.2/client-server-api/#post_matrixclientv3createroom),
but `matrix_create_room` tries to read it from the response. This field
is deprecated and does not appear on the spec.

~~For those implementations which are not returning the alias, we will
issue a new request to load room's aliases in order to make this
function compatible.~~

We calculate what the alias should be, and return that instead.

Fixes: #1071 
Closes: https://github.com/matrix-org/sytest/pull/1112

Signed-off-by: Guillem Nieto <gnieto.talo@gmail.com>

### DIFF
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -317,15 +317,12 @@ sub matrix_create_room
       my ( $body ) = @_;
       my $room_id = $body->{room_id};
 
-      do_request_json_for( $user,
-         method => "GET",
-         uri    => "/v3/rooms/$room_id/aliases",
-      )-> then( sub {
-         my ( $aliases_body ) = @_;
-         my $room_alias = first { index($_, $opts{room_alias_name}) >= 0 } @{$aliases_body->{aliases}};
-
+      if (defined $opts{room_alias_name}) {
+         my $room_alias = sprintf( '#%s:%s', $opts{room_alias_name}, $user->server_name );
          Future->done($body->{room_id}, $room_alias);
-      })
+      } else {
+         Future->done($body->{room_id}, undef);
+      }
    });
 }
 
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -315,22 +315,17 @@ sub matrix_create_room
       content => \%opts,
    )->then( sub {
       my ( $body ) = @_;
+      my $room_id = $body->{room_id};
 
-      if (defined $body->{room_alias}) {
-         Future->done( $body->{room_id}, $body->{room_alias});
-      } else {
-         my $room_id = $body->{room_id};
+      do_request_json_for( $user,
+         method => "GET",
+         uri    => "/v3/rooms/$room_id/aliases",
+      )-> then( sub {
+         my ( $aliases_body ) = @_;
+         my $room_alias = first { index($_, $opts{room_alias_name}) >= 0 } @{$aliases_body->{aliases}};
 
-         do_request_json_for( $user,
-            method => "GET",
-            uri    => "/v3/rooms/$room_id/aliases",
-         )-> then( sub {
-            my ( $aliases_body ) = @_;
-            my $room_alias = first { index($_, $opts{room_alias_name}) >= 0 } @{$aliases_body->{aliases}};
-
-            Future->done($body->{room_id}, $room_alias);
-         })
-      }
+         Future->done($body->{room_id}, $room_alias);
+      })
    });
 }
 
--- a/tests/10apidoc/30room-create.pl
+++ b/tests/10apidoc/30room-create.pl
@@ -317,12 +317,11 @@ sub matrix_create_room
       my ( $body ) = @_;
       my $room_id = $body->{room_id};
 
-      if (defined $opts{room_alias_name}) {
-         my $room_alias = sprintf( '#%s:%s', $opts{room_alias_name}, $user->server_name );
-         Future->done($body->{room_id}, $room_alias);
-      } else {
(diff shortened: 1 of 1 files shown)


tests/10apidoc/30room-create.pl
```diff
@@ -287,8 +288,10 @@ =head2 matrix_create_room
 commandline.
 
 The resultant future completes with two values: the room_id from the
-/createRoom response; the room_alias from the /createRoom response (which is
-non-standard and its use is deprecated).
+/createRoom response; and the room_alias. room_alias will be loaded from 
+/createRoom (if it exists, which is non-standard and its use is deprecated) or
+it will find the alias by loading room's aliases, in case that it's missing
+from /crateRoom.
```
DMRobertson, line 294:
```suggestion
from /createRoom.
```

gnieto:
@DMRobertson My understanding is that this fields has never been on the spec. 

There's a couple of open issues on Synapse about this:

https://github.com/matrix-org/synapse/issues/10321
https://github.com/matrix-org/synapse/issues/6879

tests/10apidoc/30room-create.pl
```diff
@@ -313,7 +316,21 @@ sub matrix_create_room
    )->then( sub {
       my ( $body ) = @_;
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      if (defined $body->{room_alias}) {
+         Future->done( $body->{room_id}, $body->{room_alias});
```
reivilibre, line 320: If this field is non-standard, I'm wondering if we should just cut it out of SyTest altogether? Slightly easier to maintain (less branching), less Synapse-specific code in SyTest, everybody's a winner.

clokep: That sounds like the thing to do to me. 👍

gnieto: I've removed it on https://github.com/matrix-org/sytest/pull/1216/commits/490735b1e1b41b6df4040f2000ad8b5987a40e39

tests/10apidoc/30room-create.pl
```diff
@@ -312,8 +315,17 @@ sub matrix_create_room
       content => \%opts,
    )->then( sub {
       my ( $body ) = @_;
+      my $room_id = $body->{room_id};
 
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      do_request_json_for( $user,
```
gnieto, line 320: I've seen in other places (for example, https://github.com/matrix-org/sytest/blob/97be5056b762a720b9adcb9c5866c6b9aa92ab40/tests/32room-versions.pl#L54) where the alias is generated via a sprintf instead of loading it from the server.

Should I use a sprintf here (and avoid a round trip to the server) as well?

richvdh: yes, I think that would be better. I don't think this function should be implicitly making extra requests to the server.

gnieto: Removed the extra request here: https://github.com/matrix-org/sytest/pull/1216/commits/5fe2531aa0366eefa149a4809b40e87034c519a4

tests/10apidoc/30room-create.pl
```diff
@@ -1,4 +1,5 @@
 use JSON qw( decode_json );
+use List::Util qw( first );
```
richvdh, line 2: I don't think this is used any more?

tests/10apidoc/30room-create.pl
```diff
@@ -287,8 +288,10 @@ =head2 matrix_create_room
 commandline.
 
 The resultant future completes with two values: the room_id from the
-/createRoom response; the room_alias from the /createRoom response (which is
-non-standard and its use is deprecated).
+/createRoom response; and the room_alias. room_alias will be loaded from 
+/createRoom (if it exists, which is non-standard and its use is deprecated) or
+it will find the alias by loading room's aliases, in case that it's missing
+from /createRoom.
```
richvdh, line 294: this is no longer correct.

tests/10apidoc/30room-create.pl
```diff
@@ -312,8 +315,14 @@ sub matrix_create_room
...
-
-      Future->done( $body->{room_id}, $body->{room_alias} );
+      my $room_id = $body->{room_id};
+
+      if (defined $opts{room_alias_name}) {
+         my $room_alias = sprintf( '#%s:%s', $opts{room_alias_name}, $user->server_name );
+         Future->done($body->{room_id}, $room_alias);
+      } else {
+         Future->done($body->{room_id}, undef);
+      }
```
richvdh, line 325: personally I would find this easier to read:

```suggestion
      my $room_alias;
      if( defined $opts{room_alias_name} ) {
         $room_alias = sprintf( '#%s:%s', $opts{room_alias_name}, $user->server_name );
      }
      Future->done( $room_id, $room_alias );
```

richvdh (changes requested):
thanks for this - a couple more tweaks please!

tests/10apidoc/30room-create.pl
gnieto: Your are right. Removed.

tests/10apidoc/30room-create.pl
gnieto: Reworded here: https://github.com/matrix-org/sytest/pull/1216/commits/05aa8459beee07fe803b32420e8aa87b6bc04fba

richvdh (approved):
thank you very much!

PR #1219, opened 2022-03-24
Fix Synapse container failing in offline mode due to missing test deps

squahtx:

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -20,4 +20,8 @@ RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2 coverage codecov
 # and test
 RUN /venv/bin/pip uninstall -q --no-cache-dir -y matrix-synapse
 
+# Pre-install test dependencies installed by `scripts/synapse_sytest.sh`.
+RUN /venv/bin/pip install -q --no-cache-dir \
+        lxml psycopg2 coverage codecov tap.py coverage_enable_subprocess
+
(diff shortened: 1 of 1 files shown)


docker/synapse.Dockerfile
```diff
@@ -20,4 +20,8 @@ RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2 coverage codecov
 # and test
 RUN /venv/bin/pip uninstall -q --no-cache-dir -y matrix-synapse
 
+# Pre-install test dependencies installed by `scripts/synapse_sytest.sh`.
+RUN /venv/bin/pip install -q --no-cache-dir \
+        lxml psycopg2 coverage codecov tap.py coverage_enable_subprocess
+
```
squahtx, line 26: The relevant part of `scripts/synapse_sytest.sh` is
https://github.com/matrix-org/sytest/blob/20521395ee42104edb6ed76aa33a9ce9329c08ed/scripts/synapse_sytest.sh#L132-L151

DMRobertson (approved):
LGTM. To check I understand: this should just mean that we cache these packages in the docker image, rather than having to pip install them every time?

squahtx:
> LGTM. To check I understand: this should just mean that we cache these packages in the docker image, rather than having to pip install them every time?

That sounds right. We install these packages in the Docker image. The later pip install in `scripts/synapse_sytest.sh` becomes a no-op (up until there is an upgrade available?). Offline mode doesn't attempt to pip install them in `scripts/synapse_sytest.sh`, but still expects the packages (or at least `coverage`) to be available.

PR #1220, opened 2022-03-24
Respect poetry lockfile in Synapse container, if present

squahtx:
Install pinned dependencies using poetry if a `poetry.lock` file is
present in the Synapse source directory, otherwise fall back to pip.

The Synapse container installs Synapse dependencies twice, once at
Docker build time and once at test time. The former installation is
intended to speed up the latter. To make this poetry-compatible, we
have to convince poetry to use the same virtual env both times. By
using the `virtualenvs.in-project` config option, we can get poetry to
use a virtual env at `./.venv`. The source directory is going to be a
mounted volume and writing into it is impolite, so we have to make a
copy of it. For simplicity, we ditch the tarball for non-poetry
installs and always make a copy.

To make offline mode work, we have to make a cached copy of a
`poetry-core` wheel available to pip. For simplicity, we do the same
thing for `setuptools` and `wheel` for pre-poetry Synapses.

Once Synapse develop has been converted to poetry, the TODO in
`docker/synapse.Dockerfile` can be resolved.

------------------------------------------------------------

NB: This blocks progress on poetry-izing Synapse, since some of the CI
steps can't be updated until the Synapse container includes poetry.

Can be reviewed commit by commit. The first commit is from #1219.

To test the container changes locally:
```sh
sudo docker build -f docker/synapse.Dockerfile docker
sudo docker run -it --env SYTEST_BRANCH=squah/pyproject-poetry-contribs --volume /home/squah/repos/synapse:/src ${hex-string-from-docker-build}
sudo docker run -it --env SYTEST_BRANCH=squah/pyproject-poetry-contribs --env OFFLINE=1 --volume /home/squah/repos/synapse:/src ${hex-string-from-docker-build}
sudo docker run -it --env SYTEST_BRANCH=squah/pyproject-poetry-contribs --volume /home/squah/repos/synapse-poetry:/src ${hex-string-from-docker-build}
sudo docker run -it --env SYTEST_BRANCH=squah/pyproject-poetry-contribs --env OFFLINE=1 --volume /home/squah/repos/synapse-poetry:/src ${hex-string-from-docker-build}
```

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -20,4 +20,8 @@ RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2 coverage codecov
 # and test
 RUN /venv/bin/pip uninstall -q --no-cache-dir -y matrix-synapse
 
+# Pre-install test dependencies installed by `scripts/synapse_sytest.sh`.
+RUN /venv/bin/pip install -q --no-cache-dir \
+        lxml psycopg2 coverage codecov tap.py coverage_enable_subprocess
+
 ENTRYPOINT [ "/bin/bash", "/bootstrap.sh", "synapse" ]
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -9,6 +9,20 @@ RUN apt-get -qq update && apt-get -qq install -y \
 
 RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.1.12
 
+# As part of the Docker build, we attempt to pre-install Synapse's dependencies
+# in the hope that it speeds up the real install of Synapse. To make this work,
+# we have to reuse the same virtual env both times. There are three ways to do
+# this with poetry:
+#  1. Ensure that the Synapse source directory lives in the same path both
+#     times. Poetry creates and reuses a virtual env based off the package name
+#     ("matrix-synapse") and directory path.
+#  2. Configure `virtualenvs.in-project` to `true`. This makes poetry create or
+#     use a virtual env at `./.venv` in the source directory.
+#  3. Run poetry with a virtual env already active. Poetry will use the active
+#     virtual env, if there is one.
+# We use the second option and make `.venv` a symlink.
+RUN poetry config virtualenvs.in-project true
+
 # /src is where we expect Synapse to be
 RUN mkdir /src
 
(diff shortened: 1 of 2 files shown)


squahtx:
CI is failing because it's running against old containers. If desired, I could split this into two PRs, one for the Dockerfile and one for the scripts, but neither makes much sense in isolation.

scripts/synapse_sytest.sh
```diff
@@ -114,34 +114,32 @@ fi
...
     exit 1
 fi
 
+if [ ! -d "$SYNAPSE_SOURCE" ]; then
+    echo "$SYNAPSE_SOURCE must be a source directory" >&2
+    exit 1
+fi
+
+# Make a copy of the source directory to avoid writing changes to the source
+# volume outside the container.
```
DMRobertson, line 128: To check: we're not using this yet, because we're not yet using poetry?

DMRobertson: We might be using this already: this is another way to avoid writing to the host's source code directory, see the deleted comment on line -117.

scripts/synapse_sytest.sh
```diff
@@ -114,34 +114,54 @@ fi
...
+        poetry install --extras redis
+        popd
+    else
+        # Install Synapse and dependencies using pip. As of pip 20.1, this will
+        # try to build Synapse in-tree, which means writing changes to the source
+        # directory.
+        # The virtual env will already be populated with dependencies from the
+        # Docker build.
+        echo "Installing Synapse using pip..."
+        /venv/bin/pip install -q --upgrade --no-cache-dir /synapse[redis]
```
DMRobertson, line 162: I think this should be `all` rather than `redis`---a bug from before. (If any extra other than `redis` changed since the container was built, that won't be picked up in this script---I think?)

docker/synapse.Dockerfile
```diff
@@ -4,12 +4,53 @@ FROM matrixdotorg/sytest:${SYTEST_IMAGE_TAG}
...
 RUN mkdir /src
 
+# Download a cache of build dependencies to support offline mode.
+# `setuptools` and `wheel` are only required for pre-poetry Synapse versions.
+# These version numbers are arbitrary and were the latest at the time.
+RUN ${PYTHON_VERSION} -m pip download --dest /pypi-offline-cache \
+        poetry-core==1.0.8 setuptools==60.10.0 wheel==0.37.1
+
+# TODO: Once poetry lands in the develop branch of Synapse, uncomment these
+#       lines and delete the pip version of the virtual env preparation.
```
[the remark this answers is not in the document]
squahtx: +54 to +62

+64 to +66 comes from #1219 which makes offline mode work.

scripts/synapse_sytest.sh
squahtx: (to be addressed in a PR I will insert after #1219 but before this one)

squahtx: Addressed in #1224

scripts/synapse_sytest.sh
```diff
@@ -114,34 +114,54 @@ fi
...
+        poetry install --extras all
+        popd
+    else
+        # Install Synapse and dependencies using pip. As of pip 20.1, this will
+        # try to build Synapse in-tree, which means writing changes to the source
+        # directory.
+        # The virtual env will already be populated with dependencies from the
+        # Docker build.
+        echo "Installing Synapse using pip..."
+        /venv/bin/pip install -q --upgrade --no-cache-dir /synapse[all]
```
DMRobertson, line 162: I guess we will end up nuking this once poetry has landed?

DMRobertson (approved):
LGTM. (Keen to see this in CI to make sure we haven't missed anything or broken the pre-poetry world!)

scripts/synapse_sytest.sh
squahtx: Not entirely sure. I'm half thinking that we should keep compatibility for pre-poetry versions of synapse. Then again you can just specify a pre-poetry branch of sytest?

DMRobertson: I'd be inclined to leave it for a few releases and then do the nuking. One for the future, I suppose.

Issue #1180, opened 2021-12-10
`/whois` is flakey(?)


DMRobertson:
https://github.com/matrix-org/synapse/runs/4484673437?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/synapse/runs/5007508334?check_suite_focus=true on https://github.com/matrix-org/synapse/pull/11861 seems to have hit this

PR #1222, opened 2022-03-25
Retry `/whois` to allow for propagation/persistence delay

reivilibre:
Might solve https://github.com/matrix-org/sytest/issues/1180?

Motivated by https://github.com/matrix-org/synapse/pull/12251, which moves the client IP persistence to a different worker than the one that handles /whois, so there is a propagation delay (when it's on the same worker, the information is enriched with unpersisted IPs).

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -31,8 +31,6 @@ sub await_purge_complete {
 test "/whois",
    requires => [ $main::API_CLIENTS[0] ],
 
-   timeout => 10,
-
    do => sub {
       my ( $http ) = @_;
 
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -51,8 +51,8 @@ test "/whois",
          # need to keep checking because the IP address won't appear for a few
          # seconds (unless the worker that flushes the IP addresses is the same
          # as the one that handles /whois).
-         return retry_until_success sub {
-            return do_request_json_for( $user,
+         repeat_until_true sub {
+            do_request_json_for( $user,
                method => "GET",
                uri    => "/v3/admin/whois/".$user->user_id,
             )->then( sub {
@@ -62,28 +62,29 @@ test "/whois",
                assert_eq( $body->{user_id}, $user->user_id, "user_id" );
                assert_json_object( $body->{devices} );
 
+               # Whether we've found a connection with the right keys
+               # (ip, last_seen, user_agent).
+               my $found_connections = 0;
+
                foreach my $value ( values %{ $body->{devices} } ) {
                   assert_json_keys( $value, "sessions" );
                   assert_json_list( $value->{sessions} );
                   assert_json_keys( $value->{sessions}[0], "connections" );
                   assert_json_list( $value->{sessions}[0]{connections} );
-                  assert_json_object( $value->{sessions}[0]{connections}[0] );
+                  foreach my $connection ( @{ $value->{sessions}[0]{connections} } ) {
+                     assert_json_object( $connection );
+                     assert_json_keys(
+                        $connection,
+                        qw( ip last_seen user_agent )
+                     );
+
+                     $found_connections = 1;
+                  }
                }
 
-               Future->done( $body->{devices} );
+               Future->done( $found_connections );
             });
-         }, initial_delay => 0.5,
-            check => sub {
-               # Once the connections object is ready, check that it contains
-               # the right keys.
-               my ( $devices ) = @_; 
-               foreach my $value ( values %$devices ) {
-                  assert_json_keys(
-                     $value->{sessions}[0]{connections}[0],
-                     qw( ip last_seen user_agent )
-                  );
-               }
-            };
+         }, initial_delay => 0.5;
       });
    };
 
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -40,30 +40,34 @@ test "/whois",
       # tightly control the actions taken by that user.
       # Conceivably this API may change based on the number of API calls the
       # user made, for instance.
+
       matrix_register_user( $http, "admin" )
       ->then( sub {
          ( $user ) = @_;
 
-         do_request_json_for( $user,
-            method => "GET",
-            uri    => "/v3/admin/whois/".$user->user_id,
-         )
-      })->then( sub {
-         my ( $body ) = @_;
+         retry_until_success {
+            sleep 1;
+            do_request_json_for( $user,
+               method => "GET",
+               uri    => "/v3/admin/whois/".$user->user_id,
+            )->then( sub {
+               my ( $body ) = @_;
 
-         assert_json_keys( $body, qw( devices user_id ) );
-         assert_eq( $body->{user_id}, $user->user_id, "user_id" );
-         assert_json_object( $body->{devices} );
+               assert_json_keys( $body, qw( devices user_id ) );
+               assert_eq( $body->{user_id}, $user->user_id, "user_id" );
+               assert_json_object( $body->{devices} );
 
-         foreach my $value ( values %{ $body->{devices} } ) {
-            assert_json_keys( $value, "sessions" );
-            assert_json_list( $value->{sessions} );
-            assert_json_keys( $value->{sessions}[0], "connections" );
-            assert_json_list( $value->{sessions}[0]{connections} );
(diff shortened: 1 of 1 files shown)


tests/48admin.pl
```diff
@@ -40,30 +40,34 @@ sub await_purge_complete {
...
          ( $user ) = @_;
 
-         do_request_json_for( $user,
-            method => "GET",
-            uri    => "/v3/admin/whois/".$user->user_id,
-         )
-      })->then( sub {
-         my ( $body ) = @_;
+         retry_until_success {
+            sleep 1;
```
richvdh, line 49: no no no. you can't do a synchronous `sleep` in async code!

richvdh (commented):
is there any way for us to tighten this test - for example, let's say that as soon as `/whois` returns a 200, the body must be correct?

reivilibre:
> is there any way for us to tighten this test - for example, let's say that as soon as `/whois` returns a 200, the body must be correct?

There is, but I'm not sure what it's buying us:

```
assert_json_list( $value->{sessions}[0]{connections} );
assert_json_keys( $value->{sessions}[0]{connections}[0], qw( ip last_seen user_agent ) );`

File: tests/48admin.pl [hunk 2]
```diff
@@ -40,30 +42,55 @@ sub await_purge_complete {
...
+                  assert_json_keys( $value, "sessions" );
+                  assert_json_list( $value->{sessions} );
+                  assert_json_keys( $value->{sessions}[0], "connections" );
+                  assert_json_list( $value->{sessions}[0]{connections} );
+                  assert_json_object( $value->{sessions}[0]{connections}[0] );
+               }
+
+               Future->done( $body->{devices} );
+            });
+         }, initial_delay => 0.5,
```
richvdh, line 81: ugh. Turns out you need to omit this comma:

```suggestion
         } initial_delay => 0.5,
```

or, possibly less weirdly, stick a `sub` in front of the `{`.

tests/48admin.pl
```diff
@@ -31,6 +31,8 @@ sub await_purge_complete {
 test "/whois",
    requires => [ $main::API_CLIENTS[0] ],
 
+   timeout => 10,
+
```
richvdh, line 35: 10 seconds is the default timeout I think?

```suggestion
```

tests/48admin.pl
```diff
@@ -40,30 +42,47 @@ sub await_purge_complete {
...
+               }
 
-         Future->done( 1 );
+               Future->done( $body->{devices} );
+            });
+         }, initial_delay => 0.5,
+            check => sub {
+               # Once the connections object is ready, check that it contains
+               # the right keys.
+               foreach my $value ( values %{ $_[0] } ) {
```
richvdh, line 79: this is a bit opaque imho.  Better:

```suggestion
               my ( $devices ) = @_; 
               foreach my $value ( values %$devices ) {
```

tests/48admin.pl
```diff
@@ -40,30 +42,47 @@ sub await_purge_complete {
...
-            assert_json_keys( $value->{sessions}[0], "connections" );
-            assert_json_list( $value->{sessions}[0]{connections} );
-            assert_json_keys( $value->{sessions}[0]{connections}[0], qw( ip last_seen user_agent ) );
-         }
+               foreach my $value ( values %{ $body->{devices} } ) {
+                  assert_json_keys( $value, "sessions" );
+                  assert_json_list( $value->{sessions} );
+                  assert_json_keys( $value->{sessions}[0], "connections" );
+                  assert_json_list( $value->{sessions}[0]{connections} );
+                  assert_json_object( $value->{sessions}[0]{connections}[0] );
```
richvdh, line 70: so what's going on here? an entry appears in the list, but it doesn't have all of the `ip`, `last_seen`, `user_agent` properties at first? Is it empty?

reivilibre: This was to try and tighten the test as you suggested: if the entry appears in the list, we should stop retrying. But we then check (in the check block) that it has the keys it should have.

richvdh: I'm afraid I'm not following. Can you give examples of the responses?

reivilibre: Not ready:
```json
  HTTP/1.1 200 OK
  Cache-Control: no-cache, no-store, must-revalidate
  Date: Wed, 30 Mar 2022 15:30:36 GMT
  Transfer-Encoding: chunked
  Server: Synapse/1.55.0rc1
  Content-Type: application/json
  Access-Control-Allow-Headers: X-Requested-With, Content-Type, Authorization, Date
  Access-Control-Allow-Methods: GET, HEAD, POST, PUT, DELETE, OPTIONS
  Access-Control-Allow-Origin: *
  
  {"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[]}]}}}
```

Ready:
```json
  HTTP/1.1 200 OK
  Cache-Control: no-cache, no-store, must-revalidate
  Date: Wed, 30 Mar 2022 15:30:42 GMT
  Transfer-Encoding: chunked
  Server: Synapse/1.55.0rc1
  Content-Type: application/json
  Access-Control-Allow-Headers: X-Requested-With, Content-Type, Authorization, Date
  Access-Control-Allow-Methods: GET, HEAD, POST, PUT, DELETE, OPTIONS
  Access-Control-Allow-Origin: *
  
  {"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[{"ip":"127.0.0.1","last_seen":1648654236155,"user_agent":"Perl + Net::Async::HTTP/0.48"}]}]}}}
```

And you can imagine this would be an example of something to reject:
```json

  HTTP/1.1 200 OK
  Cache-Control: no-cache, no-store, must-revalidate
  Date: Wed, 30 Mar 2022 15:30:42 GMT
  Transfer-Encoding: chunked
  Server: Synapse/1.55.0rc1
  Content-Type: application/json
  Access-Control-Allow-Headers: X-Requested-With, Content-Type, Authorization, Date
  Access-Control-Allow-Methods: GET, HEAD, POST, PUT, DELETE, OPTIONS
  Access-Control-Allow-Origin: *
  
  {"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[{}]}]}}}
```

richvdh: ok, so, it surprises me that `assert_json_object( $value->{sessions}[0]{connections}[0] )` is passing, for 
```json
{"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[]}]}}}
````

... because in that case, `connections[0]` should be `undef`, which would fail the assertion.

Hence, I'm a bit confused about what you're trying to test for here.

reivilibre: `{"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[]}]}}}` wouldn't have passed either before or after this change, because indeed `connections[0]` wouldn't be set.

To step back, the goal of this PR is to retry `/whois` to give chance for the data to propagate in the homeserver.

Originally `{"user_id":"@admin:localhost:8840","devices":{"":{"sessions":[{"connections":[]}]}}}` (the response given when the information is not ready) was rejected outright by this test.
Specifically, this was rejected by `assert_json_keys( $value->{sessions}[0]{connections}[0], qw( ip last_seen user_agent ) );`, since `connections[0]` is unset.

This request has now been moved to a `retry_until_successful` block.

You then suggested in https://github.com/matrix-org/sytest/pull/1222#pullrequestreview-921685760 that we should tighten the test so that as soon as \<something happens\>, it resolves to a pass or a fail (rather than retrying until the timeout is up).

My interpretation of this given the context is that `"connections":[]` should **continue** retrying (because no `connection`s have become available yet), **but what actually gets tightened** is that any of these should stop retrying and fail immediately:
- `"connections":[{}]`
- `"connections":[{"someotherkey": 4242}]`
- (et cetera; any `connections` object that is present but missing the specified keys)

As a result:
- `assert_json_object( $value->{sessions}[0]{connections}[0] );` has been introduced in the main block, so that retries happen when `connections` is an empty list.
- `assert_json_keys( $value->{sessions}[0]{connections}[0], qw( ip last_seen user_agent ) );` moved to the check block, so that once the retries have finished (i.e. `connections` is non-empty), we check the desired keys are present. This prevents the test passing if the response starts off as invalid (object present with absent keys) and then later becomes valid.

PR #1224, opened 2022-03-25
Install correct extras in Synapse container at test time

squahtx:
Instead of relying on the version of extras installed at Docker build time
or through explicit `pip install`s, ensure that appropriate versions of
all extras are installed at test time. In particular, the url_preview
(which implies lxml), postgres (which implies psycopg2) and redis extras
are wanted.

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -14,7 +14,6 @@ RUN mkdir /src
 # Manually upgrade pip to ensure it can locate Cryptography's binary wheels
 RUN ${PYTHON_VERSION} -m venv /venv && /venv/bin/pip install -U pip
 RUN /venv/bin/pip install -q --no-cache-dir matrix-synapse[all]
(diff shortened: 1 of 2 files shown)


docker/synapse.Dockerfile
```diff
@@ -14,14 +14,13 @@ RUN mkdir /src
 # Manually upgrade pip to ensure it can locate Cryptography's binary wheels
 RUN ${PYTHON_VERSION} -m venv /venv && /venv/bin/pip install -U pip
 RUN /venv/bin/pip install -q --no-cache-dir matrix-synapse[all]
-RUN /venv/bin/pip install -q --no-cache-dir lxml psycopg2 coverage codecov
```
squahtx, line 17: Turns out #1219 wasn't really needed and I was just bad at reading.

DMRobertson (approved):
Sorry I didn't spot this. Many levels of nesting and indirection with all of this gubbins!

LGTM and I think this makes things a little clearer.

squahtx:
Thank you for the speedy review!

PR #1225, opened 2022-03-25
Fix Synapse Dockerfile trying to install non-existant python3.x-pip

squahtx:
Unlike `-venv` and `-dev`, there are no `python3.{7,8,9,10}-pip` Debian
or Ubuntu packages. There is, however, always a `python3-pip` package.

For Debian, see https://packages.debian.org/search?keywords=python3-pip
and https://packages.debian.org/search?keywords=python3.8-pip

For Ubuntu, see https://packages.ubuntu.com/search?keywords=python3-pip
and https://packages.ubuntu.com/search?keywords=python3.8-pip

Fixes the Dockerfile change from #1220.

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -5,7 +5,7 @@ FROM matrixdotorg/sytest:${SYTEST_IMAGE_TAG}
 ARG PYTHON_VERSION=python3
 RUN apt-get -qq update && apt-get -qq install -y \
     ${PYTHON_VERSION} ${PYTHON_VERSION}-dev ${PYTHON_VERSION}-venv \
-    ${PYTHON_VERSION}-pip eatmydata redis-server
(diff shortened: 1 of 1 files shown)


DMRobertson (approved):
It's _weird_ to me that pip is packaged separately... but what do I know?

DMRobertson:
> There is, however, always a `python3-pip` package

We use Ubuntu as well as Debian to run the Python 3.8 jobs. But there is always a python3-pip there too: https://packages.ubuntu.com/search?keywords=python3-pip

squahtx:
> > There is, however, always a `python3-pip` package
> 
> We use Ubuntu as well as Debian to run the Python 3.8 jobs. But there is always a python3-pip there too: https://packages.ubuntu.com/search?keywords=python3-pip

Gooid catch, I missed that. I'll update the commit message.

PR #1229, opened 2022-03-30
Document a syntactic footgun around `retry_until_success` and fix tests that fell into the trap

reivilibre:
Whilst working on #1222 , I found out that specifying additional arguments to `retry_until_success` won't do what you expect if your code reference isn't using the `sub` keyword and you write a comma after it.

This PR:

- makes note of this in the documented example code
- favours keeping the comma but writing `sub` (leaving off a comma felt way too obscure)
- fixes 2 instances of the problem that I was able to locate


reivilibre:
Sadly it seems like some of those tests fail if you actually check their result :-(

squahtx (approved):

PR #1231, opened 2022-03-31
Accept `/state` requests in `create_outlier_event`

richvdh:
A follow-up to #1223: Another place that we can get `/state` requests.

### DIFF
--- a/tests/50federation/00prepare.pl
+++ b/tests/50federation/00prepare.pl
@@ -136,7 +136,7 @@ and, when it arrives, sends a response with the given state.
 =cut
 
 sub await_and_handle_request_state {
-   my ($inbound_server, $room, $event_id, $state_events ) = @_;
+   my ( $inbound_server, $room, $event_id, $state_events ) = @_;
 
    my @auth_chain = map { $inbound_server->datastore->get_auth_chain_events( $room->id_for_event( $_ )) } @$state_events;
 
--- a/tests/50federation/00prepare.pl
+++ b/tests/50federation/00prepare.pl
@@ -232,6 +232,8 @@ sub send_and_await_outlier {
 
    log_if_fail "create_outlier_event: events Q, R, S", [ $outlier_event_id_Q, $backfilled_event_id_R, $sent_event_id_S ];
 
+   my $state_req_fut;
+
    Future->needs_all(
       # send S
       $outbound_client->send_event(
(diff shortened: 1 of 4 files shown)


tests/50federation/00prepare.pl
```diff
@@ -279,14 +279,41 @@ sub send_and_await_outlier {
...
+      )->then( sub {
+         my ( $req, @params ) = @_;
+         log_if_fail "/state request", \@params;
+
+         my $resp = {
+            pdus => [ values( %initial_room_state ) ],
+            auth_chain => [
+               map { $inbound_server->datastore->get_auth_chain_events( $room->id_for_event( $_ )) }
+                  values( %initial_room_state )
+               ],
```
squahtx, line 295:
```suggestion
            ],
```

tests/50federation/00prepare.pl
```diff
@@ -279,14 +279,41 @@ sub send_and_await_outlier {
...
+
+         log_if_fail "/state response", $resp;
+         $req->respond_json( $resp );
+
+         # return a future which never completes, so that wait_any is not
+         # satisfied.
+         return Future->new();
+      });
+
       # wait for S to turn up in /sync
```
squahtx, line 306: Out of interest, does the `/state` request only happen during the `/sync`, or can it happen before?

squahtx (approved):
lgtm!

tests/50federation/00prepare.pl
richvdh: hrm. It can happen before. Which probably means I've got this in the wrong place.

richvdh: (ftr: it's unclear if there's actually time for the `/state` request to arrive between us sending the response to `/state_ids` and creating the listener for `/state`, but it's probably best to move the listener setup earlier to be on the safe side)

richvdh: Right, I've now updated this. PTAL?

tests/50federation/36state.pl
```diff
@@ -835,45 +822,31 @@ sub get_state_ids_from_server {
...
                };
 
                log_if_fail "/state_ids response", $resp;
 
+               # once we respond to `/state_ids`, the server may send a /state request;
+               # be prepared to answer that.  (it may, alternatively, send individual
+               # /event requests)
+               $state_req_fut = await_and_handle_request_state(
+                  $inbound_server, $room, $missing_event_y->{event_id}, \@state
+               );
```
squahtx, line 840: Does sending the real auth chain here defeat the purpose of the test?

richvdh: errm I don't *think* so, because synapse doesn't actually take a whole lot of notice of the auth_events in the response.

That said, I'll override the auth events here.

tests/50federation/00prepare.pl
```diff
@@ -281,36 +316,26 @@ sub send_and_await_outlier {
...
+            $receiving_user, $room_id, check => sub {
+               my ( $event ) = @_;
+               log_if_fail "create_outlier_event: Got event", $event;
+               my $event_id = $event->{event_id};
+               return $event_id eq $sent_event_id_S;
+            },
+         ),
       );
    })->then( sub {
       # cancel the /state handler, if it was unused
```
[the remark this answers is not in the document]
squahtx: Ah, I didn't realise that `wait_any` would cancel it.

squahtx (approved):
lgtm!

richvdh:
Thanks!

PR #1233, opened 2022-03-31
Remove groups/communities tests.

clokep:
Part of matrix-org/synapse#11584, required for matrix-org/synapse#12344.

I think this is safe to land at any point since it has been declared this code is deprecated.


DMRobertson (approved):
:fire:

PR #1234, opened 2022-03-31
Wait for joined room to come down sync

S7evinK:
Should de-flake `Local device key changes get to remote servers` on Dendrite.

### DIFF
--- a/tests/50federation/40devicelists.pl
+++ b/tests/50federation/40devicelists.pl
@@ -17,12 +17,8 @@ test "Local device key changes get to remote servers",
          alias   => $room_alias,
       );
 
-      do_request_json_for( $user,
-         method => "POST",
-         uri    => "/v3/join/$room_alias",
-
-         content => {},
-      )->then( sub {
+      matrix_join_room_synced( $user, $room_alias)
+      ->then( sub {
          Future->needs_all(
             $inbound_server->await_edu( "m.device_list_update", sub {1} )
             ->then( sub {


tests/41end-to-end-keys/06-device-lists.pl
```diff
@@ -419,7 +419,7 @@ sub sync_until_user_in_device_list_id
          # sure that remote_leaver *doesn't* appear in the meantime.
 
          my $wait_for_id = $remote2->user_id;
-         retry_until_success sub {
+         retry_until_success {
```
DMRobertson, line 422: Is this a style thing, or is there a semantic meaning to removing `sub` here?

DMRobertson (commented):
Seems reasonable to me.

tests/41end-to-end-keys/06-device-lists.pl
S7evinK: Not really a style thing I guess. I'm not too familiar with Perl, but with it the parameter `max_iterations` is ignored - resulting in flakiness in Dendrite.

S7evinK: [Example of only iterating a few times](https://github.com/matrix-org/dendrite/runs/5777053987?check_suite_focus=true), even if it should iterate 20 times.

reivilibre: Yeah, since there's `, max_iterations = 20;` after  the code reference, it's crucial that you write `sub` here (or remove the comma, but that seems uglier).
The reason is that as it stands, `max_iterations = 20;` becomes a new statement and so the result of `retry_until_success` will be ignored (and it won't receive that parameter either).

> Not really a style thing I guess. I'm not too familiar with Perl, but with it the parameter `max_iterations` is ignored - resulting in flakiness in Dendrite.

I'm a Perl noob so take it with a grain of salt, but the conclusion we came to is that writing it this way means that `retry_until_success` is basically entirely ignored. As such, the reason it passes is because it's ignoring the result :-); you can imagine the reason why it's flakey when you make it actually check the result...

This test is flakey on Synapse too so I'm very keen to fix it, but I had a stare at it and scratched my head and didn't figure it out yet!

S7evinK: Ok, but then I don't get how I can pass the `max_iterations => 20;` to `retry_until_success`.
With `sub`, we get the linked result (that should be after #1229 was merged), where `retry_until_success` only iterates the default of 7 times.

reivilibre: The reason it only iterated 6 times is because the test timed out.

```
  # Started: 2022-03-31 19:23:34.768
  # Ended: 2022-03-31 19:23:44.972
  # Timed out waiting for test at ./run-tests.pl line 864.
```

I think it needs `timeout => 20,` or something on the test?

reivilibre (changes requested):
(for the reasons described, removing `sub` causing the test to be ignored is not the right solution here :-))

S7evinK:
Closed in favor of https://github.com/matrix-org/sytest/pull/1249

PR #1235, opened 2022-04-03
PUT redaction endpoints

gnieto:
Sytest uses an endpoint for redaction which uses POST (instead
of PUT) and does not contain a transaction id. I've been reviewing the
latest spec version and I could only find the PUT version
(https://spec.matrix.org/v1.2/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid).

I've changed the endpoints to use the speced endpoint (both synapse and
dendrite seems to support both). I'm not certain about if this is an
issue on the spec or in sytest: Feel free to close this PR if the
previous behaviour is the expected one.

--- 

Additionally, I've added a test which checks that two redactions with the same transaction id generates a
single redaction event.

### DIFF
--- a/tests/30rooms/10redactions.pl
+++ b/tests/30rooms/10redactions.pl
@@ -223,3 +223,41 @@ test "Redaction of a redaction redacts the redaction reason",
          Future->done;
       });
    };
+
+test "PUT /rooms/:room_id/redact/:event_id/:txn_id is idempotent",
+   requires => [ local_user_fixtures( 2 ),
+                 qw( can_send_message )],
+
+   do => sub {
+      my ( $creator, $sender ) = @_;
+      my $txn_id = random_transaction_id();
+      my ( $event_id_1, $event_id_2 );
+
+      make_room_and_message( [ $creator, $sender ], $sender )
+      ->then( sub {
+         my ( $room_id, $to_redact ) = @_;
+
+         $to_redact = uri_escape( $to_redact );
+
+         do_request_json_for( $creator,
+            method => "PUT",
+            uri    => "/v3/rooms/$room_id/redact/$to_redact/$txn_id",
+            content => {},
+         )->then( sub {
+            my ( $body ) = @_;
+            $event_id_1 = $body->{event_id};
+
+            do_request_json_for( $creator,
+               method => "PUT",
+               uri    => "/v3/rooms/$room_id/redact/$to_redact/$txn_id",
+               content => {},
+            )->then(sub {
+               my ( $body ) = @_;
+               $event_id_2 = $body->{event_id};
+               assert_eq( $event_id_1, $event_id_2 );
+
+               Future->done(1);
+            })
+         });
+      })
+   };
--- a/tests/30rooms/10redactions.pl
+++ b/tests/30rooms/10redactions.pl
@@ -28,16 +28,22 @@ Makes a /redact request
 
 =cut
 
+sub random_transaction_id
+{
+   join "", map { chr 65 + rand 26 } 1 .. 20;
+}
+
 sub matrix_redact_event
 {
    my ( $user, $room_id, $event_id, %params ) = @_;
 
    $room_id = uri_escape( $room_id );
    my $esc_event_id = uri_escape( $event_id );
+   my $txn_id = random_transaction_id();
 
    do_request_json_for( $user,
-      method => "POST",
-      uri    => "/v3/rooms/$room_id/redact/$esc_event_id",
+      method => "PUT",
+      uri    => "/v3/rooms/$room_id/redact/$esc_event_id/$txn_id",
       content => \%params,
    )->then( sub {
       my ( $body ) = @_;
@@ -87,7 +93,7 @@ sub matrix_redact_event_synced
 
 push @EXPORT, qw( matrix_redact_event_synced );
 
-test "POST /rooms/:room_id/redact/:event_id as power user redacts message",
+test "PUT /rooms/:room_id/redact/:event_id/:txn_id as power user redacts message",
    requires => [ local_user_fixtures( 2 ),
                  qw( can_send_message )],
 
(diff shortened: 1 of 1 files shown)


tests/30rooms/10redactions.pl
```diff
@@ -28,16 +28,22 @@ =head2 matrix_redact_event
 
 =cut
 
+sub random_transaction_id
+{
+   join "", map { chr 65 + rand 26 } 1 .. 20;
```
richvdh, line 33:
```suggestion
   join "", map { chr( 65 + rand 26 )} 1 .. 20;
```

tests/30rooms/10redactions.pl
```diff
@@ -215,3 +223,41 @@ sub matrix_redact_event_synced
...
+            )->then(sub {
+               my ( $body ) = @_;
+               $event_id_2 = $body->{event_id};
+               assert_eq( $event_id_1, $event_id_2 );
+
+               Future->done(1);
+            })
+         });
+      })
+   };
```
richvdh, line 263: missing NL:

```suggestion
   };
```

tests/30rooms/10redactions.pl
```diff
@@ -215,3 +223,41 @@ sub matrix_redact_event_synced
...
+      ->then( sub {
+         my ( $room_id, $to_redact ) = @_;
+
+         $to_redact = uri_escape( $to_redact );
+
+         do_request_json_for( $creator,
+            method => "PUT",
+            uri    => "/v3/rooms/$room_id/redact/$to_redact/$txn_id",
+            content => {},
+         )->then( sub {
```
richvdh, line 246: No need to nest all these `->then` calls. This "callback hell" is the whole reason Futures exist!

```suggestion
         );
      })->then( sub {
```

likewise after the other call to `do_request_json_for`.

richvdh (changes requested):
yeah, looks like the POST format has never been specced.

richvdh:
have filed https://github.com/matrix-org/synapse/issues/12362

tests/30rooms/10redactions.pl
gnieto: I copied the code from: https://github.com/matrix-org/sytest/blob/fbc57e787703e7fcdbd3ab1b622a773c7bc7cc19/tests/52user-directory/01public.pl#L456

Should I change it there as well? There are a few more occurrences in the codebase with the same pattern.

tests/30rooms/10redactions.pl
gnieto: :+1: Thanks for the suggestion, I'm not much familiar with Perl.

tests/30rooms/10redactions.pl
gnieto: Do you mean New Line with NL?

richvdh: yes.

tests/30rooms/10redactions.pl
richvdh: I don't know if it's necessary to go and change existing code. But I think it's better to strive for clarity rather than cargo-cult existing unclear code.

tests/30rooms/10redactions.pl
```diff
@@ -28,16 +28,22 @@ =head2 matrix_redact_event
 
 =cut
 
+sub random_transaction_id
+{
+   join "", map { chr (65 + rand 26) } 1 .. 20;
```
richvdh, line 33:
```suggestion
   join "", map { chr( 65 + rand 26 )} 1 .. 20;
```

tests/30rooms/10redactions.pl
```diff
@@ -215,3 +223,42 @@ sub matrix_redact_event_synced
...
+            );
+         })
+         ->then(sub {
+            my ( $body ) = @_;
+            $event_id_2 = $body->{event_id};
+            assert_eq( $event_id_1, $event_id_2 );
+         })
+      })
+   };
+
```
richvdh, line 264: too many NLs !

```suggestion
```

tests/30rooms/10redactions.pl
```diff
@@ -215,3 +223,42 @@ sub matrix_redact_event_synced
...
+         my ( $room_id, $to_redact ) = @_;
+
+         $to_redact = uri_escape( $to_redact );
+
+         do_request_json_for( $creator,
+            method => "PUT",
+            uri    => "/v3/rooms/$room_id/redact/$to_redact/$txn_id",
+            content => {},
+         )
+         ->then( sub {
```
richvdh, line 247:
```suggestion
         );
      })->then( sub {
```

(and corresponding unnesting below)

tests/30rooms/10redactions.pl
```diff
@@ -215,3 +223,42 @@ sub matrix_redact_event_synced
...
+            my ( $body ) = @_;
+            $event_id_1 = $body->{event_id};
+
+            do_request_json_for( $creator,
+               method => "PUT",
+               uri    => "/v3/rooms/$room_id/redact/$to_redact/$txn_id",
+               content => {},
+            );
+         })
+         ->then(sub {
```
richvdh, line 257:
```suggestion
         })->then( sub {
```

richvdh (changes requested):
thanks for continuing to work on this!

tests/30rooms/10redactions.pl
gnieto: I should have said that just after pushing my second set of changes, but I think that's not possible because we need `$room_id` and `$to_redact` (which are declared on line 238) in the scope on these 2 last blocks. Is there any way of structure it in a better way so these blocks can be unnested?

richvdh: oh right. you can declare the variables at the function level with

```
my $room_id;
my $to_redact;
```

... and then remove the `my` at line 238. But it may not be worth it (and it looks like other tests in this file follow the same pattern as you).

gnieto: I decided to apply your suggestion here: https://github.com/matrix-org/sytest/pull/1235/commits/3536eae46a82659eab3ffe4b2738889768f56836

I've also formatted the `then`s as you suggested on the other comment. 

Sorry for this back and forth about the formatting. Hope it's fine now!

gnieto:
@richvdh I think all feedback has been addressed. Do you mind to check it again when you have some time, please? Thanks!

richvdh (approved):
looks great now, thank you!

PR #1237, opened 2022-04-06
Use the `dendrite` branch by default for Dendrite

neilalexander:
That way we can choose when to merge the latest `develop` changes into our branch rather than find CI turns into a sea of red randomly.


squahtx (approved):
lgtm!

docker/dendrite.Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN tar xf go.tar.gz -C /goroot --strip-components=1
 ENV GOROOT=/goroot
 ENV GOPATH=/gopath
 ENV PATH="/goroot/bin:${PATH}"
+ENV SYTEST_DEFAULT_BRANCH dendrite
```
richvdh, line 14: a comment here would be good. What does this do, and why is it useful? (I shouldn't have to go digging around in a bunch of other files to figure it out)

richvdh (changes requested):

S7evinK:
I've added a comment about the ENV and removed building `sytest-dendrite` on pull to `develop`. We're going to build the image on-demand from the `dendrite` branch later on.

kegsay (approved):

PR #1240, opened 2022-04-08
Work around a race condition when installing Synapse with poetry

squahtx:
Poetry runs multiple pip operations in parallel. Unfortunately this
results in race conditions when a dependency is installed while another
dependency is being up/downgraded in `scripts/synapse_sytest.sh`.
Configure poetry to serialize `pip` operations.

See https://github.com/matrix-org/synapse/issues/12419


DMRobertson (approved):
It's unfortunate that we keep having to workaround things like this. :(

Let's  get it in and see if this does the trick.

Is there anywhere else we're likely to see this? I _think_ CI always installs into a blank environment everywhere else.

PR #1241, opened 2022-04-11
Fix poetry config not taking effect inside GitHub Actions 

squahtx:
Makes the workaround in #1240 to actually work, thus fixing https://github.com/matrix-org/synapse/issues/12419.

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -9,6 +9,14 @@ RUN apt-get -qq update && apt-get -qq install -y \
 
 RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.1.12
 
+# According to https://python-poetry.org/docs/configuration/,
+# poetry stores its config in `$XDG_CONFIG_HOME/pypoetry/config.toml` or
+# `$HOME/.config/pypoetry/config.toml` if `XFG_CONFIG_HOME` is undefined.
+# GitHub Actions changes `$HOME` to `/github/home` at runtime, which makes
+# poetry look for its config in the wrong place. Explicitly set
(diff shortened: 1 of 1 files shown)


callahad:
Would it be more reliable to use Poetry's environment variables (https://python-poetry.org/docs/configuration/#using-environment-variables) than to deal with `poetry config` and setting `XDG_CONFIG_HOME`?

squahtx:
I implemented @callahad's suggestion, which is a lot nicer.

docker/synapse.Dockerfile
```diff
@@ -67,10 +67,11 @@ RUN /venv/bin/pip install -q --no-cache-dir \
...
 # in race conditions when a dependency is installed while another dependency
 # is being up/downgraded in `scripts/synapse_sytest.sh`. Configure poetry to
-# serialize `pip` operations.
+# serialize `pip` operations by setting `experimental.new-installer` to a falsy
+# value.
 # See https://github.com/matrix-org/synapse/issues/12419
 # TODO: Once poetry 1.2.0 has been released, use the `installer.max-workers`
 #       config option instead.
-RUN poetry config experimental.new-installer false
+ENV POETRY_EXPERIMENTAL_NEW_INSTALLER ""
```
squahtx, line 75: poetry has a bug where this environment variable is not converted to a boolean, so we choose a falsy string value for it.
It's only fixed in poetry 1.2.0: https://github.com/python-poetry/poetry/commit/d4c2fe44f43d426ecf963bbce9325c45dbc40fc8

DMRobertson: Great find---can we stick this comment in the dockerfile though?

DMRobertson (approved):
Other than the comment, I'm happy. Let's land it and see if sytest runs are happier?

PR #1243, opened 2022-04-21
Pre-install Synapse dependencies in container using poetry

squahtx:
Now that Synapse develop uses poetry, we can build the sytest-synapse
Docker image using poetry. 


DMRobertson:
I could have sworn there was somewhere I used the pip-only version recently... let me see if I can find it.

squahtx:
> Would that be broken by this change?

It shouldn't. This PR only affects container build, which runs against develop and happens way before that CI step.

DMRobertson (approved):
I'm happy if CI is happy. (Assuming that sytest's CI is a reasonable representation of Synapse's CI?)

Issue #1130, opened 2021-09-06
`If remote user leaves room, changes device and rejoins we see update in /keys/changes` is flaky


richvdh:
https://github.com/matrix-org/synapse/pull/10745/checks?check_run_id=3524341005

DMRobertson:
Wasn't able to reproduce this locally. I think the example in question got stuck waiting for matrix_leave_room_synced but I can't see any clues there.

It would be really helpful if we could match up request ids from sytest logs to synapse logs!

For completeness, I was trying to reproduce with this:

```shell
while docker run --rm -v /home/dmr/workspace/synapse:/src:ro -v /home/dmr/workspace/sytest/:/sytest -v /tmp/logs:/logs -e POSTGRES=1 -e WORKERS=1 -e MULTI_POSTGRES=1 matrixdotorg/sytest-synapse:testing --torture-replication 50 -s tests/41end-to-end-keys/06-device-lists.pl; do :; done
```

richvdh:
and https://github.com/matrix-org/synapse/runs/4106909748?check_suite_focus=true

DMRobertson:
and https://github.com/matrix-org/synapse/runs/4431172752?check_suite_focus=true

richvdh:
and https://github.com/matrix-org/synapse/runs/4621032498?check_suite_focus=true

DMRobertson:
and https://github.com/matrix-org/synapse/runs/4719905893?check_suite_focus=true

DMRobertson:
https://github.com/matrix-org/sytest/runs/6111956286?check_suite_focus=true

richvdh:
https://github.com/matrix-org/synapse/runs/6588295124?check_suite_focus=true

PR #1244, opened 2022-04-21
Don't install dependencies from git when running Synapse in offline mode

squahtx:
Previously we used `--no-index` to prevent use of PyPI in offline mode,
but that did not apply to git dependencies. Add the `--no-deps` flag
to prevent pip from re-cloning those.

At this time, the only git dependency we might have is twisted trunk
during a daily CI run.

Fixes https://github.com/matrix-org/synapse/issues/12517, hopefully (untested).


DMRobertson (approved):
Sgtm, assuming CI is happy!

DMRobertson:
sytest failure appears to be https://github.com/matrix-org/sytest/issues/1130 ?

PR #1245, opened 2022-04-21
Fix incorrect `pip` invocation when checking deps in offline mode

squahtx:
Fixes bug introduced in #1244. CI didn't trip because it only happens in offline mode.


DMRobertson (approved):
Oops. Sorry I missed this.

PR #1246, opened 2022-04-25
Join room synced for room summary tests

S7evinK:
This ensures the user is actually joined to the room and doesn't flake on Dendrite.


DMRobertson (approved):
I'm happy if CI is.

(Maybe we should just get rid of <del>`matrix_join_room_synced`</del> Edit: `matrix_join_room` altogether?)

S7evinK:
I'd be in favor of doing so, but not sure if it actually makes sense to have both. I'll create an issue for this.

PR #1248, opened 2022-04-28
Enable registration again in Dendrite

S7evinK:
We are disabling registration by default with https://github.com/matrix-org/dendrite/pull/2402, this PR ensures Sytest can still run as expected.

### DIFF
--- a/lib/SyTest/Homeserver/Dendrite.pm
+++ b/lib/SyTest/Homeserver/Dendrite.pm
@@ -344,6 +344,7 @@ sub _start_monolith
       '--api-bind-address', $self->{bind_host} . ':1' . $self->unsecure_port,
       '--tls-cert', $self->{paths}{tls_cert},
       '--tls-key', $self->{paths}{tls_key},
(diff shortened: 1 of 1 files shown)


neilalexander (changes requested):
We'll need to set `-really-enable-open-registration` too in `_start_monolith` in `lib/SyTest/Homeserver/Dendrite.pm`.

S7evinK:
> We'll need to set `-really-enable-open-registration` too in `_start_monolith` in `lib/SyTest/Homeserver/Dendrite.pm`.

Since `registration_shared_secret` is set, we don't necessarily need the parameter. Will add it, just to make it clear.

neilalexander:
Oh, if `registration_shared_secret` is being used then let's not, since it'll just come up as an unknown flag if we run it against old Dendrite versions.

neilalexander:
We will need `-really-enable-open-registration` after all because the shared secret bit here doesn't do what we think it does.

neilalexander (approved):

Issue #1247, opened 2022-04-26
Get rid of non-synced room joins


S7evinK:
As mentioned by @DMRobertson in https://github.com/matrix-org/sytest/pull/1246#pullrequestreview-952250932, it might make sense to remove the non synced version of `matrix_join_room` (and most likely other non synced actions, like `matrix_create_room` etc.)

The reason: Dendrite often flakes due to it's async nature and events might not be persisted yet when hitting `/sync` or other endpoints.

I'm not sure if we can simply replace it or if the non-synced versions have their use cases.

richvdh:
Sounds good - we're not aware of any important usecase for the non-synced versions.

S7evinK:
Solved by #1249

PR #1249, opened 2022-04-29
Replace most non-synced actions

S7evinK:
As explained in https://github.com/matrix-org/sytest/issues/1247, this PR changes most non-synced versions with the synced version.
`matrix_create_room_synced` now also returns the room alias.

~~Removes `tests/60app-services/04asuser.pl`, as support for AS users to use `/sync` was removed in https://github.com/matrix-org/synapse/pull/2948.~~


tests/60app-services/04asuser.pl
```diff
@@ -1,43 +0,0 @@
...
-            method => "POST",
-            uri    => "/v3/join/$room_id",
-
-            params => {
-               user_id => $as_user->user_id,
-            },
-            content => {},
-         );
-      });
-   };
```
richvdh, line 43: surely we still need to test these things? I don't really understand why these tests need to be removed?

tests/10apidoc/30room-create.pl
```diff
@@ -461,6 +461,6 @@ sub matrix_create_room_synced
                $_[0]->{event_id} eq $event_id
             });
          },
-      )->then_done( $room_id );
+      )->then_done( $room_id, $room_alias );
```
richvdh, line 464: please keep the documentation on the function definition in sync with this change.

tests/30rooms/03members-remote.pl
```diff
@@ -255,15 +255,15 @@
...
          room_alias_name  => $room_alias_name,
          creation_content => {
             "m.federate" => JSON::false,
          },
       )->then( sub {
          my ( undef, $room_alias ) = @_;
 
-         matrix_join_room( $remote_user, $room_alias )
+         matrix_join_room_synced( $remote_user, $room_alias )
             ->main::expect_http_403;
```
richvdh, line 267: these things (there are quite a lot of them) which check for a 403 response should probably not be synced. Obviously if we're expecting the operation to fail, then there is no point polling `/sync`, but moreover: we want to be sure it is the `/join` that is returning 403, not the `/sync`.

tests/50federation/40devicelists.pl
```diff
@@ -338,17 +334,7 @@
...
-         my $iter = 0;
-         retry_until_success {
-            $iter++;
-            matrix_join_room( $user, $room->room_id,
-               server_name => $inbound_server->server_name,
-            )->on_fail( sub {
-               my ( $exc ) = @_;
-               chomp $exc;
-               log_if_fail "Room join iteration $iter failed: $exc";
-            });
```
richvdh, line 350: isn't this loop important?

richvdh (changes requested):
this is mildly terrifying, but seems generally plausible!

tests/60app-services/04asuser.pl
S7evinK: Restored, as it was `/sync` failing (like you mentioned below).

tests/30rooms/03members-remote.pl
S7evinK: Good point, didn't think about that. Restored the normal version where it makes sense.

tests/50federation/40devicelists.pl
S7evinK: At least I'm not seeing why it should be important.
We're leaving the room synced, wait for the federated server to receive the leave and re-join the room.

richvdh: hrm you're probably right!

tests/10apidoc/30room-create.pl
```diff
@@ -426,7 +426,9 @@ =head2 matrix_create_room_synced
 
 The parameters are passed through to C<matrix_create_room>.
 
-The resultant future completes with the room_id.
+The resultant future completes with the room_id and a room_alias.
+If room_alias_name is present in %params, an alias will be built with the given 
+alias and user's server name. Othwerwise, it will return undef.
```
richvdh, line 431: Odd line wrapping (and a typo) here:

```suggestion
The resultant future completes with the room_id and a room_alias. If
room_alias_name is present in %params, an alias will be built with the given 
alias and user's server name. Otherwise, it will return undef.
```

Ideally the symbols in that text (`room_alias_name`, `%params`, `undef`) would be wrapped with a `C<...>` [formatting code](https://perldoc.perl.org/perlpod#Formatting-Codes), but it's not critical.

richvdh: Please can you update the example on line 422 as well?

tests/30rooms/06invite.pl
```diff
@@ -396,7 +396,7 @@ sub invited_user_can_reject_invite_for_empty_room
    do => sub {
       my ( $creator, $room_id ) = @_;
 
-      matrix_invite_user_to_room( $creator, $creator, $room_id )
+      matrix_invite_user_to_room_synced( $creator, $creator, $room_id )
          ->main::expect_http_403;
```
richvdh, line 400: this is expecting a 403.

S7evinK:
Not sure if `Tags appear in the v1 /initialSync` is flakey in general on Synapse, looks like?

richvdh:
> Not sure if Tags appear in the v1 /initialSync is flakey in general on Synapse, looks like?

doesn't look like it, I'm afraid! Generally we open issues when we see flakey tests, and there is nothing around this one.

richvdh (changes requested):
looks generally good, but there is a failing test!

tests/42tags.pl
```diff
@@ -148,7 +148,7 @@ sub create_room_and_add_tag
 {
    my ( $user ) = @_;
 
-   matrix_create_room_synced( $user )->then( sub {
+   matrix_create_room( $user )->then( sub {
```
richvdh, line 151: ok. Any idea why this is needed?

richvdh (approved):
lgtm!

PR #1250, opened 2022-04-30
Add support for building arm64 docker test images

brianathere:
Build both the AMD64 and ARM64 flavors
Allows faster, more battery efficient local test runs during development on M1 Mac

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,7 +26,7 @@ jobs:
           - base_image: debian:bookworm
             tag: bookworm
     steps:
-      - name: Set up QEMU
+        - name: Set up QEMU
         id: QEMU
         uses: docker/setup-qemu-action@v1
 
@@ -52,7 +52,7 @@ jobs:
           labels: "gitsha1=${{ github.sha }}"
           file: docker/base.Dockerfile
           build-args: "BASE_IMAGE=${{ matrix.base_image }}"
-          tags: herenotthere/sytest:${{ matrix.tag }}
+          tags: matrixdotorg/sytest:${{ matrix.tag }}
 
   build-dependent-images:
     needs: build-sytest-images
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -76,10 +81,14 @@ jobs:
               PYTHON_VERSION=python3.10
           - sytest_image_tag: buster
             dockerfile: dendrite
-            tags: "matrixdotorg/sytest-dendrite:go113,matrixdotorg/sytest-dendrite:latest"
+            tags: "matrixdotorg/sytest-dendrite:go118,matrixdotorg/sytest-dendrite:latest"
```
DMRobertson, line 84: What's the story here---is this necessary or a drive-by bump?

docker/base.Dockerfile
```diff
@@ -6,6 +6,7 @@ ENV DEBIAN_FRONTEND noninteractive
 
 # Install base dependencies that Python or Go would require
 RUN apt-get -qq update && apt-get -qq install -y \
+    apt-utils \
```
DMRobertson, line 9: Can you explain why apt-utils is now necessary?

docker/synapse.Dockerfile
```diff
@@ -3,9 +3,12 @@ ARG SYTEST_IMAGE_TAG=buster
 FROM matrixdotorg/sytest:${SYTEST_IMAGE_TAG}
 
 ARG PYTHON_VERSION=python3
+
+ENV DEBIAN_FRONTEND noninteractive
```
DMRobertson, line 7: What does this do, and/or why wasn't it needed before?

DMRobertson (commented):
This seems reasonable, but there are some other changes mixed in here (go 1.18, apt-utils, formatting changes) which I don't understand. Go 1.18 seems like it might merit its own change, and maybe even the formatting to. Could you explain a few things for me?

.github/workflows/docker.yml
brianathere: I don't have an explanation of this one, I couldn't find the go113 tag used anywhere so I bumped it to align with the go version.

docker/base.Dockerfile
brianathere: With the QEMU build building for multiple flavors (adding arm64 in this change), I ran into an issue where apt-utils needed to be re-installed in order to work.

docker/synapse.Dockerfile
brianathere: Aligning the two dockerfiles, and also related to the addition of apt-utils (as above).

brianathere:
> This seems reasonable, but there are some other changes mixed in here (go 1.18, apt-utils, formatting changes) which I don't understand. Go 1.18 seems like it might merit its own change, and maybe even the formatting to. Could you explain a few things for me?

go 1.18, I can split this into another change, I tested it all together, but it shouldn't be necessary, just nice.
Formatting was just consistency. 
The others (apt-utils) we related to using buildex and qemu to build alternative flavors.

.github/workflows/docker.yml
brianathere: I've reverted the bump to Go 1.18 and confirmed 1.16 runs OK on ARM64. This should be updated in another change to a supported version of Go at this time, I'll make another change to bump to 1.18.

DMRobertson:
> > This seems reasonable, but there are some other changes mixed in here (go 1.18, apt-utils, formatting changes) which I don't understand. Go 1.18 seems like it might merit its own change, and maybe even the formatting to. Could you explain a few things for me?
> 
> go 1.18, I can split this into another change, I tested it all together, but it shouldn't be necessary, just nice. Formatting was just consistency. The others (apt-utils) we related to using buildex and qemu to build alternative flavors.

Many thanks. I've kicked off another CI run. I'm a little busy right now, so I've put this back in the review queue.

brianathere:
> > > This seems reasonable, but there are some other changes mixed in here (go 1.18, apt-utils, formatting changes) which I don't understand. Go 1.18 seems like it might merit its own change, and maybe even the formatting to. Could you explain a few things for me?
> > 
> > 
> > go 1.18, I can split this into another change, I tested it all together, but it shouldn't be necessary, just nice. Formatting was just consistency. The others (apt-utils) we related to using buildex and qemu to build alternative flavors.
> 
> Many thanks. I've kicked off another CI run. I'm a little busy right now, so I've put this back in the review queue.

When you say "review queue", is there a queue or workflow system I should be looking at and working with here?

DMRobertson (approved):
No queue for yourself to worry about. I just mean that I'd marked it for review by the synapse and dendrite teams.

Thanks for your comments and for pulling out the go 1.18 change. This LGTM!

PR #1261, opened 2022-06-17
Pick auth events more intelligently

richvdh:
Previously, we were always including the `join_rules` event. We should only do so for join and invite events.


squahtx (approved):
lgtm, seems to match the spec!

PR #1263, opened 2022-06-17
More fixes to auth events

richvdh:
A followup to #1261: we should not be including `m.room.join_rules` in the auth
events for most events.


squahtx (approved):
lgtm

reivilibre (approved):

PR #1264, opened 2022-06-27
Allow /notifications API to return read push actions

erikjohnston:
This is in line with the spec.

Fixes the tests in #13118

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -241,7 +241,7 @@ sub configure_logger
 
       loggers => {
          synapse => {
-            level => "INFO"
+            level => "DEBUG"
          }
       },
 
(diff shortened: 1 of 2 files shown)


lib/SyTest/Homeserver.pm
```diff
@@ -241,7 +241,7 @@ sub configure_logger
 
       loggers => {
          synapse => {
-            level => "INFO"
+            level => "DEBUG"
```
DMRobertson, line 244: Is this debug that crept in?

DMRobertson (approved):
Aside from the comment, seems reasonable to me.

lib/SyTest/Homeserver.pm
erikjohnston: I HATE MY LIFE

PR #1265, opened 2022-06-28
Don't require `end` on `/messages`

S7evinK:
As per the spec, the `end` key is not required if there are no more events for the user.

> If no further events are available (either because we have reached the start of the timeline, or because the user does not have permission to see any more events), this property is omitted from the response.

### DIFF
--- a/tests/10apidoc/34room-messages.pl
+++ b/tests/10apidoc/34room-messages.pl
@@ -158,7 +158,7 @@ test "GET /rooms/:room_id/messages returns a message",
       })->then( sub {
          my ( $body ) = @_;
 
-         # We should still get events and a "end" key, check it is actually there
+         # We should still get events and a "end" key: check they are actually there
          assert_json_keys( $body, qw( start end chunk ));
          assert_json_list( $body->{chunk} );
          $token = $body->{end};
@@ -166,13 +166,10 @@ test "GET /rooms/:room_id/messages returns a message",
          scalar @{ $body->{chunk} } > 0 or
             die "Expected some messages but got none at all\n";
       })->then( sub {
-
-         # Do another call to /messages, this time we don't expect to receive a "end" key
+         # Do another call to /messages. This time we don't expect to receive a "end" key
          do_request_json_for( $user,
             method => "GET",
             uri    => "/v3/rooms/$room_id/messages",
-
-            # With no params this does "forwards from END"; i.e. nothing useful
             params => {
                 dir => "b",
                 from => $token,
@@ -220,7 +217,7 @@ test "GET /rooms/:room_id/messages lazy loads members correctly",
 
          log_if_fail "Body", $body;
          
-         # We should still get events and a "end" key, check it is actually there
+         # We should still get events and a "end" key: check they are actually there
          assert_json_keys( $body, qw( start end state chunk ));
          assert_json_list( $body->{chunk} );
          assert_json_list( $body->{state} );
@@ -234,12 +231,10 @@ test "GET /rooms/:room_id/messages lazy loads members correctly",
             die "Expected some messages but got none at all\n";
 
       })->then( sub {
-         # Do another call to /messages, this time we don't expect to receive a "end" key
+         # Do another call to /messages. This time we don't expect to receive a "end" key
          do_request_json_for( $user,
             method => "GET",
             uri    => "/v3/rooms/$room_id/messages",
-
-            # With no params this does "forwards from END"; i.e. nothing useful
             params => {
                 dir => "b",
                 from => $token,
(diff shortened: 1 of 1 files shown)


DMRobertson:
I think the spec's wording is actually stricter here: I think it's saying that `end` *must* be omitted if there are no more messages to retrieve.

Should we be asserting that `end` is _not_ present in the json body?

S7evinK:
Makes sense, yea.
Just tested that locally, Synapse isn't yet happy with that, though.

richvdh:
I thought that recently got fixed in synapse.

DMRobertson:
> I thought that recently got fixed in synapse.

matrix-org/synapse#12903

S7evinK:
Hm, what happens if the returned `chunk` count is less than the requested limit? Shouldn't that also result in `end` being omitted?
Or would that be [this](https://github.com/matrix-org/synapse/pull/12903/files#diff-a3b3a9ae833ad473bff3bb3825e66d7ebdf0ef33cc1f59e4ee5fd7c030c522ecR535-R536)?

If so, we probably can't assert that `end` is actually missing in these tests, without extending the tests.

tests/10apidoc/34room-messages.pl
```diff
@@ -158,12 +158,33 @@ sub matrix_send_room_text_message
...
 
+         # We should still get events and a "end" key, check it is actually there
          assert_json_keys( $body, qw( start end chunk ));
          assert_json_list( $body->{chunk} );
+         $token = $body->{end};
 
          scalar @{ $body->{chunk} } > 0 or
             die "Expected some messages but got none at all\n";
+      })->then( sub {
+
```
richvdh, line 169:
```suggestion
```

tests/10apidoc/34room-messages.pl
```diff
@@ -158,12 +158,33 @@ sub matrix_send_room_text_message
...
+         # Do another call to /messages, this time we don't expect to receive a "end" key
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/v3/rooms/$room_id/messages",
+
+            # With no params this does "forwards from END"; i.e. nothing useful
```
richvdh, line 175: I don't think this needs repeating here

```suggestion
```

tests/10apidoc/34room-messages.pl
```diff
@@ -158,12 +158,33 @@ sub matrix_send_room_text_message
...
+         # We should still get events and a "end" key, check it is actually there
          assert_json_keys( $body, qw( start end chunk ));
          assert_json_list( $body->{chunk} );
+         $token = $body->{end};
 
          scalar @{ $body->{chunk} } > 0 or
             die "Expected some messages but got none at all\n";
+      })->then( sub {
+
+         # Do another call to /messages, this time we don't expect to receive a "end" key
```
richvdh, line 170:
```suggestion
         # Do another call to /messages. This time we don't expect to receive a "end" key
```

tests/10apidoc/34room-messages.pl
```diff
@@ -158,12 +158,33 @@ sub matrix_send_room_text_message
       })->then( sub {
          my ( $body ) = @_;
 
+         # We should still get events and a "end" key, check it is actually there
```
richvdh, line 161:
```suggestion
         # We should still get events and a "end" key: check they are actually there
```

tests/10apidoc/34room-messages.pl
```diff
@@ -197,10 +219,12 @@ sub matrix_send_room_text_message
          my ( $body ) = @_;
 
          log_if_fail "Body", $body;
-
+         
+         # We should still get events and a "end" key, check it is actually there
```
richvdh, line 223:
```suggestion
         # We should still get events and a "end" key: check they are actually there
```

tests/10apidoc/34room-messages.pl
```diff
@@ -209,6 +233,25 @@ sub matrix_send_room_text_message
          scalar @{ $body->{chunk} } > 0 or
             die "Expected some messages but got none at all\n";
 
+      })->then( sub {
+         # Do another call to /messages, this time we don't expect to receive a "end" key
```
richvdh, line 237:
```suggestion
         # Do another call to /messages. This time we don't expect to receive a "end" key
```

tests/10apidoc/34room-messages.pl
```diff
@@ -209,6 +233,25 @@ sub matrix_send_room_text_message
...
+         do_request_json_for( $user,
+            method => "GET",
+            uri    => "/v3/rooms/$room_id/messages",
+
+            # With no params this does "forwards from END"; i.e. nothing useful
```
richvdh, line 242:
```suggestion
```

richvdh (approved):
lgtm otherwise

PR #1266, opened 2022-07-07
Use generate-config for Dendrite

S7evinK:
This should allow us to introduce new config entries without having to create extra PRs for SyTest.
It also re-enables CI runs for Dendrite, as it shouldn't be as flakey as before.


kegsay (approved):
CI is green, soooo LGTM?!

PR #1270, opened 2022-07-15
Does bumping poetry to 1.1.14 help?

DMRobertson:


DMRobertson:
Oh, I thought that CI would build the docker images but it doesn't. Whoops.

Testing locally with `docker buildx build --build-arg SYTEST_IMAGE_TAG=buster --file docker/synapse.Dockerfile --pull .` I still get the same failure.

PR #1271, opened 2022-07-19
Work around Synapse Docker build failures by using a `--cache-dir`

### LINKED ISSUE #1269: Unexplained sytest poetry build failure weirdness
https://github.com/matrix-org/sytest/runs/7250780423?check_suite_focus=true#step:6:929

I wonder if this is fixed by bumping to poetry 1.1.14 (see https://github.com/matrix-org/synapse/pull/13285 and https://github.com/python-poetry/poetry/pull/5973)

squahtx:
When the `--no-cache-dir` option is given to pip, setuptools>=48.0.0
installs msgpack to `/usr/lib/python3.7/site-packages/` instead of
`/usr/local/lib/python3.7/dist-packages/`, where Python can't find it.

---

Fixes #1269.

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -10,7 +10,13 @@ RUN apt-get -qq update && apt-get -qq install -y \
         apt-utils ${PYTHON_VERSION} ${PYTHON_VERSION}-dev ${PYTHON_VERSION}-venv \
         python3-pip eatmydata redis-server
 
-RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.1.12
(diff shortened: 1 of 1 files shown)


DMRobertson (commented):
Thank you so much for digging into this. It looks like a Herculean effort to me!

Should we be using poetry 1.1.14 here, as in https://github.com/matrix-org/synapse/pull/13285? Perhaps not---I don't think we're generating lockfiles, just reading from them?

No objections otherwise.

DMRobertson (approved):
Thanks! SGTM. I assume the dendrite failures are a flake or known problem; they can't be caused by this.

squahtx:
Agreed, let's merge it.

PR #1272, opened 2022-07-19
Update to poetry 1.14 in the Synapse Docker container

squahtx:
The recommended poetry version for Synapse is now 1.1.14.
Update it here too.

---

Synapse's recommended poetry version was changed in https://github.com/matrix-org/synapse/pull/13285


DMRobertson (approved):
Thanks (for consistency's sake :) )

PR #1274, opened 2022-07-20
Upgrade pip instead of using `--cache-dir` workaround in Synapse image

squahtx:
In #1271, we added a workaround for #1269, where msgpack was not
installed correctly. Remove the workaround and upgrade pip instead.

---

As suggested in https://github.com/pypa/setuptools/issues/3457#issuecomment-1190125849, upgrading pip does the trick.

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get -qq update && apt-get -qq install -y \
         apt-utils ${PYTHON_VERSION} ${PYTHON_VERSION}-dev ${PYTHON_VERSION}-venv \
         python3-pip eatmydata redis-server
 
-RUN ${PYTHON_VERSION} -m pip install -q --upgrade pip
+RUN ${PYTHON_VERSION} -m pip install -q pip==22.1.2
 RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.1.14
 
 # As part of the Docker build, we attempt to pre-install Synapse's dependencies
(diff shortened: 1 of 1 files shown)


docker/synapse.Dockerfile
```diff
@@ -10,13 +10,8 @@ RUN apt-get -qq update && apt-get -qq install -y \
...
         python3-pip eatmydata redis-server
 
-RUN ${PYTHON_VERSION} -m pip install --cache-dir /pip-cache poetry==1.1.14 && \
-    rm -rf /pip-cache
-# Instead of using `--no-cache-dir`, we specify a `--cache-dir` and delete it
-# afterwards. This works around a bug where `msgpack` gets installed to
-# `site-packages`, where Python can't find it, instead of `dist-packages`, when
-# using `setuptools>=48.0.0` and the `--no-cache-dir` option.
-# https://github.com/matrix-org/sytest/issues/1269
+RUN ${PYTHON_VERSION} -m pip install -q --upgrade pip
```
DMRobertson, line 13: Should we be _pinning_ the version of pip here?

docker/synapse.Dockerfile
```diff
@@ -10,13 +10,8 @@ RUN apt-get -qq update && apt-get -qq install -y \
...
+RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.1.12
```
DMRobertson, line 14: Conflict with #1272 here?

squahtx: Oops, fixed!

docker/synapse.Dockerfile
squahtx: I can't convince myself that we should pin or not pin pip either way.
The dependencies of poetry are equally free to update underneath us. So perhaps we should always upgrade pip, in case one of them starts relying on some new pip behaviour?

Adding a comment sounds like a good idea.

squahtx:
Thanks for the speedy review!

PR #1278, opened 2022-07-27
Enable race detection for dendrite server used to run tests

brianathere:
This change builds and installs the dendrite server and tools enabling the race golang race detector and puts the race logs into the same location as other logs so they can be reviewed if tests introduce races.

This will fail sytest on Dendrite as there are multiple race failures on startup. I have another PR to Dendrite in progress that will allow this to pass, putting this PR up as a placeholder to land after the fixes are in Dendrite.

Signed-off-by: Brian Meek <brian@hntlabs.com>

### DIFF
--- a/docker/dendrite.Dockerfile
+++ b/docker/dendrite.Dockerfile
@@ -1,7 +1,7 @@
 ARG SYTEST_IMAGE_TAG=buster
 FROM matrixdotorg/sytest:${SYTEST_IMAGE_TAG}
 
-ARG GO_VERSION=1.18.3
+ARG GO_VERSION=1.19
 ARG TARGETARCH
 ENV GO_DOWNLOAD https://dl.google.com/go/go${GO_VERSION}.linux-${TARGETARCH}.tar.gz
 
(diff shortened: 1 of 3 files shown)


brianathere:
@richvdh This PR should now pass against Dendrite with the other fixes to race conditions that have landed on that side. I'd like to get it merged upstream so all future changes also test for races.

neilalexander:
In our testing, running the integration tests with the race detector slows down the runs significantly (sometimes doubling the time taken if not more). I think that may not be desirable for all CI runs. We may need to think about what makes sense here.

brianathere:
> In our testing, running the integration tests with the race detector slows down the runs significantly (sometimes doubling the time taken if not more). I think that may not be desirable for all CI runs. We may need to think about what makes sense here.

Do you have additional tiers of tests elsewhere that this could be added to if every run isn't appropriate?

S7evinK:
We could use an environment variable to enable race detection in specific CI runs, like when merging to `main`.
So we'd have a copy of the current `sytest` CI step in Dendrite with enabled race detection, executed when we initially passed the tests. IMHO it shouldn't fail the whole run, so shouldn't be added to `integration-tests-done`. (We still see it failed/are notified and can take action)

S7evinK:
Race detection is now possible with an optional `RACE_DETECTION` environment variable, see https://github.com/matrix-org/sytest/pull/1308

PR #1281, opened 2022-08-26
Remove `user_id` from expected keys in `/search` results

S7evinK:
As per the [spec](https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3search), the `user_id` field doesn't exist on `/search` results.

### DIFF
--- a/tests/43search.pl
+++ b/tests/43search.pl
@@ -45,7 +45,7 @@ test "Can search for an event by body",
 
          assert_json_keys( $result, qw( rank result ) );
          assert_json_keys( $result->{result}, qw(
-            event_id room_id user_id content type
+            event_id room_id content type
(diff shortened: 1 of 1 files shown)


richvdh:
does this suggest there is a bug in existing HS implementations that they return this field? It would be good to have issues filed if so.

S7evinK:
Probably, I've just checked older spec versions and didn't find that field in any of them. Will file a bug for Synapse.

S7evinK:
Thinking about it, it could also be a spec bug.

squahtx:
Synapse bug is https://github.com/matrix-org/synapse/issues/7925

squahtx (approved):
lgtm!

PR #1282, opened 2022-08-26
Always use Redis with Synapse workers.

clokep:
Prep work for matrix-org/synapse#11728.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -33,8 +33,6 @@ The following environment variables can be set with `-e` to control the test run
   PostgreSQL databases where the main store and state store are split.
 - `WORKERS`: set non-empty to test a worker-mode deployment rather than a
   monolith. Requires `POSTGRES`.
-- `REDIS`: set non-empty to use redis replication rather than old
-  TCP. Requires `WORKERS`.
 - `OFFLINE`: set non-empty to avoid updating the python or perl dependencies.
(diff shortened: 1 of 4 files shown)


docker/README.md
```diff
@@ -33,32 +33,20 @@ The following environment variables can be set with `-e` to control the test run
...
 - `WORKERS`: set non-empty to test a worker-mode deployment rather than a
   monolith. Requires `POSTGRES`.
-- `REDIS`: set non-empty to use redis replication rather than old
-  TCP. Requires `WORKERS`.
 - `OFFLINE`: set non-empty to avoid updating the python or perl dependencies.
 - `BLACKLIST`: set non-empty to change the default blacklist file to the
   specified path relative to the Synapse directory
 - `TIMEOUT_FACTOR`: sets a number that test timeouts are multiplied by.
 
 Some examples of running Synapse in different configurations:
```
H-Shay, line 41: Since there is now only one example here perhaps it makes sense to change this, i.e. "An example running Synapse in worker mode"? or similar?

clokep: Good idea!

H-Shay (approved):
This looks good to me with the caveat that my perl is not fluent so my approval is about 50% "I can decipher what you're doing and it seems reasonable" and 50% "the tests work and I trust you" wrt to the ha proxy changes

PR #1284, opened 2022-09-01
Bump poetry version in Synapse docker image

erikjohnston:


docker/synapse.Dockerfile
```diff
@@ -34,10 +34,9 @@ ENV POETRY_VIRTUALENVS_IN_PROJECT true
 RUN mkdir /src
 
 # Download a cache of build dependencies to support offline mode.
-# `setuptools` and `wheel` are only required for pre-poetry Synapse versions.
```
erikjohnston, line 37: I don't think this is true? For a start we use pkg_resources from setuptools, and anyway we'll need it for the rust support.

erikjohnston:
I wonder if we should also skip the TCP replication worker tests that are duplicated by Redis?

reivilibre (approved):

PR #1286, opened 2022-09-04
Do not check optional fields in sync tests

gnieto:
Do not check fields which are optional and may not exist in sync
responses. Some optional fields are kept because they are relevant to
that specific test.


DMRobertson (approved):
SGTM, thanks!

PR #1287, opened 2022-09-08
Fix offline mode

erikjohnston:
We added a new build dependency to Synapse so we have to add it to the docker image.

Hopefully fixes https://github.com/matrix-org/synapse/issues/13740

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -69,6 +69,15 @@ jobs:
         with:
           path: sytest
 
+      # There should already be a Rust compiler in the docker image, but we
+      # pick a particular version here.
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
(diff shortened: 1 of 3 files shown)


DMRobertson:
CI is failing---is that expected? I can't remember if we have to build docker images first to get Sytest to be happy

erikjohnston:
@DMRobertson turns out we weren't installing rust. Have now added it to the docker image, but since that isn't rebuilt for PRs I also added it to the GHA job steps. Not sure if we want to remove that before we land it?

DMRobertson:
> Not sure if we want to remove that before we land it?

I'd be happy with docker-image only or GHA-only, with a mild preference for docker-image so that it doesn't have to get installed on every run. But also happy if you just want to merge as-is

erikjohnston:
I'll merge as is and then rip out the GHA installation in a separate PR, i think

PR #1288, opened 2022-09-08
Remove unnecessary installation of Rust from CI step

erikjohnston:
Follows on from https://github.com/matrix-org/sytest/pull/1287

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -68,16 +68,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: sytest
-
-      # There should already be a Rust compiler in the docker image, but we
-      # pick a particular version here.
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: 1.61.0
-            override: true
-      - uses: Swatinem/rust-cache@v2
-
+          
       # TODO the shell script below is nicked from complement. We use this pattern
       # in a few places. Can we make this an Action so it's easier to reuse?
       - name: Fetch corresponding synapse branch


DMRobertson:
Err is _this_ meant to fail CI?

(Tempted to just leave it as it is and not ask too many questions)

erikjohnston:
I am *extremely* confused as to wtf is going on now.

squahtx:
The docker build step where we install the rust toolchain only adds 32B to the image(????)
https://hub.docker.com/layers/matrixdotorg/sytest-synapse/focal/images/sha256-2206b265dc62720753ee1dbe43b03002f293455a905ee8e8e2a5cd0fc67c2bae?context=explore

Looks like the install command failed, but the GHA still passed?
https://github.com/matrix-org/sytest/runs/8253226528?check_suite_focus=true#step:6:957

erikjohnston:
Aha, that's annoying. Why on earth did the docker image build not fail :/

erikjohnston:
Hopefully https://github.com/matrix-org/sytest/pull/1290 fixes the docker image

DMRobertson:
> Aha, that's annoying. Why on earth did the docker image build not fail :/

See https://github.com/docker/docker.github.io/blob/master/develop/develop-images/dockerfile_best-practices.md#using-pipes

https://stackoverflow.com/questions/69684133/dl4006-warning-set-the-shell-option-o-pipefail-before-run-with-a-pipe-in-it seems to suggest that a `SHELL` directive can be used to set this globally in the dockerfile.

https://github.com/hadolint/hadolint might be interesting to pick this up?

erikjohnston:
This is now fixed!

squahtx (approved):
lgtm

PR #1289, opened 2022-09-10
Prevent using deprected endpoints for non-deprecated endpoint tests

gnieto:
Until now, members tests were indirectly using deprecated endpoints even in those tests which are not marked with `deprecated_endpoints`. These tests initialization were using `flush_events_for`, which is calling `/r0/events`, but this initialization is not strictly needed because there's no further call to `/r0/events` (so it does not need to track an `eventstream_token`).

One side effect of calling the `/r0/events` endpoint was that presence for the users [was updated ](https://github.com/matrix-org/synapse/blob/4c4889cac0e6f7df4689287b9fddea1bf8b15b7f/synapse/handlers/events.py#L70-L74), which enabled some of the test to do some assertions regarding presence events. Because of the removal of `flush_events_for` during the test set up, the tests related to presence have been changed to explicitly set the presence of the corresponding users.

### DIFF
--- a/tests/30rooms/02members-local.pl
+++ b/tests/30rooms/02members-local.pl
@@ -122,11 +122,10 @@ test "All room members see all room members' presence in global initialSync",
       my ( $first_user, $local_user, $room_id ) = @_;
       my @all_users = ( $first_user, $local_user );
 
-      Future->needs_all( map {
-         my $user = $_;
-
-         flush_events_for( $user )
-         ->then(sub {
+      Future->needs_all( map { matrix_set_presence_status($_, "online") } @all_users)
+      ->then(sub {
+         Future->needs_all( map {
+            my $user = $_;
             matrix_initialsync( $user )->then( sub {
                my ( $body ) = @_;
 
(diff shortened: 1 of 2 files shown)


gnieto:
Apparently there's something which is not compatible with synapse in workers mode. I will take a look to check if I can find the cause.

gnieto:
I _think_ that Dendrite's failure might be a false negative, because this test is totally unrelated to the tests I've changed. I do not have permissions to re-run that job. Could someone re-run it to check if it was a transient issue, please?

DMRobertson:
> I _think_ that Dendrite's failure might be a false negative, because this test is totally unrelated to the tests I've changed. I do not have permissions to re-run that job. Could someone re-run it to check if it was a transient issue, please?

Rerun of all jobs in progress at https://github.com/matrix-org/sytest/actions/runs/3028940997

DMRobertson:
Dendrite failed again, but in a different way on a different job. There is likely to be a flakey test somewhere :(

I've kicked off a third run for the hell of it. https://github.com/matrix-org/sytest/actions/runs/3028940997 lets you navigate between the three runs.

babolivier (approved):
Looks good to me, thanks for your contribution! :)

PR #1290, opened 2022-09-12
Fix Rust installation in Synapse Docker image

erikjohnston:
I tried to add this in #1287

### DIFF
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -6,6 +6,9 @@ ARG PYTHON_VERSION=python3
 
 ENV DEBIAN_FRONTEND noninteractive
 
+# Ensire we die correctly when using pipes in RUN.
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
 RUN apt-get -qq update && apt-get -qq install -y \
         apt-utils ${PYTHON_VERSION} ${PYTHON_VERSION}-dev ${PYTHON_VERSION}-venv \
         python3-pip eatmydata redis-server curl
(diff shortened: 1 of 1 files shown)


docker/synapse.Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get -qq update && apt-get -qq install -y \
         apt-utils ${PYTHON_VERSION} ${PYTHON_VERSION}-dev ${PYTHON_VERSION}-venv \
-        python3-pip eatmydata redis-server
+        python3-pip eatmydata redis-server cur
```
DMRobertson, line 11: is this meant to be `curl`?

erikjohnston: SIGH, yes thank you. My crankness continues

DMRobertson (changes requested):
Can we add something which would detect the failure to invoke curl? See https://github.com/matrix-org/sytest/pull/1288#issuecomment-1243505026

docker/synapse.Dockerfile
```diff
@@ -6,9 +6,12 @@ ARG PYTHON_VERSION=python3
 
 ENV DEBIAN_FRONTEND noninteractive
 
+# Ensire we die correctly when using pipes in RUN.
```
DMRobertson, line 9:
```suggestion
# Ensure we die correctly when using pipes in RUN.
```

Otherwise LGTM, I'm happy if CI is

DMRobertson (approved):

erikjohnston:
Ermh, not sure why a single Dendrite test is broken, but given nothing in this PR has changed anything to do with them I'm going to merge as is.

PR #1293, opened 2022-09-25
Remove `origin` from events and assertions

gnieto:
Related to: https://github.com/matrix-org/matrix-spec/pull/998

Setting `origin` field in events/PDUs is not required any more, so I've removed assertions and assignments to this field.

### DIFF
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -240,7 +240,6 @@ sub join_room
             auth_events content depth prev_events room_id sender
             state_key type ) ),
 
-         origin           => $store->server_name,
          origin_server_ts => $self->time_ms,
       );
 
--- a/lib/SyTest/Federation/Client.pm
+++ b/lib/SyTest/Federation/Client.pm
@@ -240,6 +240,9 @@ sub join_room
             auth_events content depth prev_events room_id sender
             state_key type ) ),
 
+         # The spec does not require the origin field, but Dendrite does. This can be removed
+         # once this issue gets solved: https://github.com/matrix-org/dendrite/issues/2736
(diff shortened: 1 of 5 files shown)


gnieto:
Apparently, Dendrite is no happy with this change. 
I can just remove the assertions, alternatively, if you are fine with that.

DMRobertson:
> Apparently, Dendrite is no happy with this change.

Probably worth opening a Dendrite bug (if there isn't one for this already).

> I can just remove the assertions, alternatively, if you are fine with that.

Sounds reasonable.

richvdh:
>  I can just remove the assertions, alternatively, if you are fine with that.

If you take this path, please can you comment the places we set `origin` with a link to the dendrite bug?

lib/SyTest/Federation/Client.pm
```diff
@@ -240,6 +240,8 @@ sub join_room
             auth_events content depth prev_events room_id sender
             state_key type ) ),
 
+         # The spec does not require the origin field, but Dendrite does. This can be removed
```
gnieto, line 243: @richvdh Is this similar to what you were expecting? Let me know if you want me to change the message. Thanks!

neilalexander:
matrix-org/dendrite#2737
matrix-org/gomatrixserverlib#341

Please hold off on doing anything until we get these landed.

neilalexander (approved):
Right, this should be OK now that we've fixed Dendrite. I have reverted aed2b10d9cbfb4ea10ac6d6dcbbf11d276ed7a33 — LGTM if CI is green.

DMRobertson (approved):
Thanks Neil A for the prompt response!

PR #1296, opened 2022-09-28
Assert default field is present in pushrules

deepbluev7:
Relates to https://github.com/matrix-org/synapse/pull/13904#issuecomment-1261387124

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>


deepbluev7:
I guess that depends on https://github.com/matrix-org/synapse/pull/13904

clokep (approved):
LGTM if builds are green now that the synapse fix has merged.

PR #1300, opened 2022-10-05
Fix no servers having historical state after purging events

MadLittleMods:
Fix no servers having historical state after purging events when running with `POSTGRES=1 -e MULTI_POSTGRES=1 -e WORKERS=1`

As seen in https://github.com/matrix-org/synapse/pull/13816#discussion_r987522070 (details and more context). The reason we're now only noticing this problem is because with the introduction of https://github.com/matrix-org/synapse/pull/13816, we stop trying to fetch missing `prev_events` if we already tried recently (backoff). The `/state_ids` for the invite event 404's which means we can't verify the rest of the DAG all the way down.

Sytest failure in CI:

https://github.com/matrix-org/synapse/actions/runs/3185784983/jobs/5198427610
```
FAILURE: #588: Can backfill purged history
```

---

The reason this isn't seen in monolith is because TODO

### Dev notes

```
docker run --rm -it -e POSTGRES=1 -e MULTI_POSTGRES=1 -e WORKERS=1 \
    -v ~/Documents/github/element/synapse:/src:ro \
    -v ~/Documents/github/element/synapse/logs:/logs \
    -v ~/Documents/github/element/sytest:/sytest:ro matrixdotorg/sytest-synapse:buster tests/48admin.pl
```

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -274,10 +274,10 @@ test "Can backfill purged history",
       ->then( sub {
          matrix_join_room_synced( $remote_user, $room_id )
       })->then( sub {
-         # Force the remote server to backfill all of the events to the beginning of the
-         # room. We want to make sure to de-outlier all of the state (like the invite
-         # event) before the join so it's available later when `server-0` asks about it
-         # again.
+         # Force the remote server (`server-1`) to backfill all of the events to the
+         # beginning of the room. We want to make sure to de-outlier all of the state
+         # (like the invite event) before the join so it's available later when
+         # `server-0` asks about it again.
          matrix_get_room_messages( $remote_user, $room_id, limit => 100 )
       })->then( sub {
          matrix_put_room_state( $user, $room_id,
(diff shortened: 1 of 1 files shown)


tests/48admin.pl
```diff
@@ -273,6 +273,12 @@ sub await_purge_complete {
       matrix_invite_user_to_room_synced( $user, $remote_user, $room_id )
       ->then( sub {
          matrix_join_room_synced( $remote_user, $room_id )
+      })->then( sub {
+         # Force the remote server (`server-1`) to backfill all of the events to the
+         # beginning of the room. We want to make sure to de-outlier all of the state
+         # (like the invite event) before the join so it's available later when
+         # `server-0` asks about it again.
+         matrix_get_room_messages( $remote_user, $room_id, limit => 100 )
```
MadLittleMods, line 281: Is there a more canonical semantic way we should be forcing `server-1` to get the whole state-chain from `server-0`?

MadLittleMods, line 281: Closing as there may be a better way forward, https://github.com/matrix-org/synapse/pull/13816#discussion_r987711019

PR #1301, opened 2022-10-06
Allow running Synapse with the asyncioreactor enabled

sandhose:
This adds a flag `--asyncio-reactor`, which sets the `SYNAPSE_ASYNCIO_REACTOR` environment variable to `1`.
It also adds a `ASYNCIO_REACTOR` environment variable to the Docker image to set this flag.

Note that it currently also sets `SYNAPSE_ASYNCIO_REACTOR` to `0` when the flag is not set, which currently enables the asyncioreactor anyway, so this will require some changes in Synapse first: https://github.com/matrix-org/synapse/pull/14092

Related: https://github.com/matrix-org/synapse/issues/14091

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -33,6 +33,7 @@ The following environment variables can be set with `-e` to control the test run
   PostgreSQL databases where the main store and state store are split.
 - `WORKERS`: set non-empty to test a worker-mode deployment rather than a
   monolith. Requires `POSTGRES`.
+- `ASYNCIO_REACTOR`: set non-empty to test with the Twisted asyncio reactor enabled
 - `OFFLINE`: set non-empty to avoid updating the python or perl dependencies.
 - `BLACKLIST`: set non-empty to change the default blacklist file to the
   specified path relative to the Synapse directory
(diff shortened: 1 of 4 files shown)


DMRobertson (commented):
It seems a bit odd to accept an env var `ASYNCIO_REACTOR`, convert that to a command-line flag `--asyncio-reactor` and then use the flag to set a different env var `SYNAPSE_ASYNC_IO_REACTOR`.  Does `SYNAPSE_ASYNC_IO_REACTOR=1 synapse_sytest.sh` pass through the environment var to the synapse processes?

Well, I took a quick look at the perl and I have no idea if the environment will pass through when we spawn the process. https://github.com/matrix-org/sytest/blob/4fa32d0314585cb1684410d0185e8a466cd3430e/lib/SyTest/Homeserver/ProcessManager.pm#L98-L103

DMRobertson:
In the interests of getting asyncio tested sooner rather than later, let's roll with this. Note that https://github.com/matrix-org/synapse/pull/14092 has already landed.

DMRobertson (approved):

PR #1304, opened 2022-10-11
Synapse: use minimal Rust profile

DMRobertson:
See matrix-org/synapse#14139 and matrix-org/synapse#14141.


reivilibre:
urr... I was meant to approve that, but hopefully I just saved you a click ;-p

PR #1307, opened 2022-10-18
Allow HSes to omit device display names

DMRobertson:
Hopefully gets https://github.com/matrix-org/synapse/pull/10015 over the line.

I don't speak perl and it's not easy to test this change against that PR because it comes from a fork.

I suggest we merge this if CI is green and move on with our lives.

PR #1311, opened 2022-10-27
Make device display name check optional for `federation/v1/user/keys/query`

anoadragon453:
Some homeservers (Synapse) have started [omitting device's display name over federation](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#allow_device_name_lookup_over_federation) by default. The spec [doesn't mandate that this field be available](https://spec.matrix.org/v1.4/server-server-api/#post_matrixfederationv1userkeysquery), so Sytest shouldn't either.

This PR modifies the check so that it is optional - only if the field is present do we check the value.

### DIFF
--- a/tests/41end-to-end-keys/04-query-key-federation.pl
+++ b/tests/41end-to-end-keys/04-query-key-federation.pl
@@ -30,9 +30,8 @@ multi_test "Can query remote device keys using POST",
 
          # Device display names are optional for POST /user/keys/query responses.
          # If one exists, ensure it's the one we expected.
-         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"};
-         (!defined $device_display_name) or ($device_display_name == "test display name") or
-            croak "Unexpected device_display_name: $device_display_name";
+         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"} // "";
+         assert_eq "test display name", $device_display_name, "device display name";
 
          Future->done(1)
       });
--- a/tests/41end-to-end-keys/04-query-key-federation.pl
+++ b/tests/41end-to-end-keys/04-query-key-federation.pl
@@ -28,8 +28,11 @@ multi_test "Can query remote device keys using POST",
 
          # TODO: Check that the content matches what we uploaded.
 
-         assert_eq( $alice_device_keys->{"unsigned"}->{"device_display_name"},
-                    "test display name" );
+         # Device display names are optional for POST /user/keys/query responses.
+         # If one exists, ensure it's the one we expected.
+         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"};
+         (!defined $device_display_name) or ($device_display_name == "test display name") or
+            croak "Unexpected device_display_name: $device_display_name";
 
          Future->done(1)
       });
--- a/tests/41end-to-end-keys/04-query-key-federation.pl
+++ b/tests/41end-to-end-keys/04-query-key-federation.pl
@@ -30,8 +30,12 @@ multi_test "Can query remote device keys using POST",
 
          # Device display names are optional for POST /user/keys/query responses.
          # If one exists, ensure it's the one we expected.
-         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"} // "";
-         assert_eq "test display name", $device_display_name, "device display name";
+         if (exists( $alice_device_keys->{"unsigned"}->{"device_display_name"} )) {
+            assert_eq(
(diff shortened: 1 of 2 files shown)


tests/41end-to-end-keys/04-query-key-federation.pl
```diff
@@ -28,8 +28,11 @@
 
          # TODO: Check that the content matches what we uploaded.
 
-         assert_eq( $alice_device_keys->{"unsigned"}->{"device_display_name"},
-                    "test display name" );
+         # Device display names are optional for POST /user/keys/query responses.
+         # If one exists, ensure it's the one we expected.
+         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"};
+         (!defined $device_display_name) or ($device_display_name == "test display name") or
+            croak "Unexpected device_display_name: $device_display_name";
```
DMRobertson, line 35: FWIW I have no idea if this is idomatic perl (from #1307).

IIRC I manually inlined `assert_eq` to get the `a == b or croak "wrong"` form, and then shoved an extra defined check at the start.

DMRobertson (approved):

tests/41end-to-end-keys/04-query-key-federation.pl
richvdh: I'd say in this situation it would be more idiomatic to use `||` between the conditions:

```
         (!defined $device_display_name || $device_display_name eq "test display name") or
            croak "Unexpected device_display_name: $device_display_name";
```

Note also the use of `eq` instead of `==`: the latter is for *numeric* comparison:

```
$ perl -e '"a" == "b" and print "true\n";'
true
```

A simpler approach here would be to do something like:
```
         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"} // "";
         assert_eq "test display name", $device_display_name, "device display name";
```

(`//` is like an `||` operator, but whereas `||` returns its right-hand operand if the left-hand one is falsey, `//` only returns its right-hand operand if the left-hand operand is `undef`. See https://perldoc.perl.org/perlop#Logical-Defined-Or)

richvdh: > Note also the use of eq instead of ==: the latter is for numeric comparison:

(and yes, this is a massive footgun)

DMRobertson: Ahh, yes, `assert_eq` does indeed use `eq`.

https://github.com/matrix-org/sytest/blob/32129d98c33ff85e6e73e67147ea3e4cd04f630d/lib/SyTest/Assertions.pm#L61-L67

Presumably this is bogus with `==` then. Can you fix that up here Andrew---and the test from #1307?

DMRobertson (changes requested):
== -> eq (aka fix my mistakes pls)

DMRobertson:
> == -> eq (aka fix my mistakes pls)

(Or better: use Rich's formulation)

(Or better better: invent a time machine and prevent perl from being created)

tests/41end-to-end-keys/04-query-key-federation.pl
anoadragon453: Sure, done in ad8ad254.

anoadragon453: I actually wonder if my original code would be easier to read for the average programmer?

```perl
         if (exists( $alice_device_keys->{"unsigned"}->{"device_display_name"} )) {
            assert_eq(
               $alice_device_keys->{"unsigned"}->{"device_display_name"},
               "test display name",
            );
         }
```

but either way I'm happy as long as it works.

tests/41end-to-end-keys/04-query-key-federation.pl
```diff
@@ -28,8 +28,10 @@
 
          # TODO: Check that the content matches what we uploaded.
 
-         assert_eq( $alice_device_keys->{"unsigned"}->{"device_display_name"},
-                    "test display name" );
+         # Device display names are optional for POST /user/keys/query responses.
+         # If one exists, ensure it's the one we expected.
+         my $device_display_name = $alice_device_keys->{"unsigned"}->{"device_display_name"} // "";
+         assert_eq "test display name", $device_display_name, "device display name";
```
DMRobertson, line 34: Errr on second thoughts: is this correct? if the `device_display_name` field is missing we'll then `assert_eq "test display name" "" "device_display_name"`, which will fail the test?

anoadragon453: Oop, shows me right for blindly copying. Have hopefully fixed in https://github.com/matrix-org/sytest/pull/1311/commits/c19f460d99107d0322f925230f644493880c85ef.

PR #1312, opened 2022-10-27
Use `true` and not `"true"` as JSON value in AS public room test

S7evinK:


DMRobertson (approved):
Oh right: perl doesn't have a built-in boolean type. Urgh.

PR #1315, opened 2022-11-11
Try until success for membership event

S7evinK:
Due to Dendrites async nature, the membership event might not be persisted yet when hitting `/state`. This now retries until we have the expected result or the test times out.

### DIFF
--- a/tests/30rooms/13guestaccess.pl
+++ b/tests/30rooms/13guestaccess.pl
@@ -141,8 +141,6 @@ test "Guest users are kicked from guest_access rooms on revocation of guest_acce
 
          matrix_set_room_guest_access( $user, $room_id, "forbidden" );
       })->then( sub {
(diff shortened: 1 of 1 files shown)


tests/30rooms/13guestaccess.pl
```diff
@@ -143,11 +143,14 @@
       })->then( sub {
          matrix_get_room_membership( $user, $room_id, $guest_user );
```
DMRobertson, line 144: I think these lines are these now redundant(?)

DMRobertson (changes requested):

tests/30rooms/13guestaccess.pl
S7evinK: Yep, this is redundant. Removed :)

S7evinK:
Yep, looks like an unrelated flake. Going to investigate this.

PR #1320, opened 2022-11-30
Add manual step to build Dendrite image

S7evinK:
Adds a manual step to build the Dendrite Sytest image.
Defines some tests as implementation specific, as they are not in the spec (`/deactivate` with a `erase` parameter) and only implemented in Synapse.


.github/workflows/docker.yml
```diff
@@ -1,84 +1,24 @@
 # GitHub actions workflow which builds and publishes the docker images.
 
-name: Build and deploy docker images
+name: Build and deploy Dendrite docker images
 
 on:
-  push:
-    branches: [develop]
```
kegsay, line 7: Surely `branches: [dendrite]`?

S7evinK: Don't we want to manually build the image? Otherwise we might pull in package updates we don't know, causing unwanted issues.

kegsay: Hmm, because perl doesn't pin dependencies?

S7evinK: Both, Perl and distribution packages.

kegsay (approved):

PR #1323, opened 2022-12-06
A missing trailing slash should be a 404.

clokep:
MSC3743 says that unknown endpoints are a 404, not a 400.

The other related tests to me I think a bit harder to decide if they should be a 400 or not, but I think the initial slash missing is clearly just an unknown endpoint.


clokep:
(This is for matrix-org/synapse#14621)

PR #1325, opened 2022-12-19
Remove deprecated `home_server` from checks

S7evinK:
The `home_server` field in responses to `/register` and `/login` is deprecated since [r0.4.0](https://matrix.org/docs/spec/client_server/r0.4.0.html)


clokep (approved):
Seems reasonable. 👍

PR #1326, opened 2023-01-16
Bump poetry-version to 1.2.2

DMRobertson:
Should be backwards-compatible, see
https://github.com/matrix-org/setup-python-poetry/pull/14 for rationale


DMRobertson:
Since CI was green I'm going to merge this as-is to get the synpase-sytest docker images building.

PR #1330, opened 2023-01-20
Retry getting unread counts

S7evinK:
This should fix flakes like https://github.com/matrix-org/sytest/actions/runs/3949863262/jobs/6761624592#step:6:5


DMRobertson (approved):
SGTM, many thanks!

PR #1331, opened 2023-02-22
Add /filter to client_reader

reivilibre:
Matches with https://github.com/matrix-org/synapse/pull/15134.


clokep (approved):
Seems reasonable from the synapse POV. Not sure if dendrite is expected to be failing?

reivilibre:
Dendrite is failing on develop, so guess so :/

PR #1332, opened 2023-02-23
Add knock to `event_creator`

dklimpel:
Matches with: https://github.com/matrix-org/synapse/pull/15133 (https://github.com/matrix-org/synapse/pull/15133#discussion_r1114454118)


clokep (approved):
Fine if matrix-org/synapse#15133 is OK.

PR #1333, opened 2023-02-24
Add federation senders to instance map

DMRobertson:
For https://github.com/matrix-org/synapse/pull/15121


reivilibre (approved):

DMRobertson:
I'm going to drop this as it seems like we're not happy with the  design of https://github.com/matrix-org/synapse/pull/15121

PR #1335, opened 2023-02-24
Build docker images overnight

DMRobertson:
Even if there are no sytest changes, there are often Synapse dependency changes which we spend time `poetry install`ing. By rebuilding the docker image we are more likely to have Synapse's latest set of dependencies installed when we come to use the sytest docker image.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,6 +5,8 @@ name: Build and deploy docker images
 on:
   push:
     branches: [develop]
+  schedule:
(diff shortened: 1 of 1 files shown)


H-Shay (approved):
lgtm

H-Shay:
I'm assuming the sytest failure is a flake.

DMRobertson:
> I'm assuming the sytest failure is a flake.

Agreed---in any case, it shouldn't be caused by this yaml change.

PR #1336, opened 2023-02-28
Change binary for Dendrite

S7evinK:
@DMRobertson: This should fix the issues with Dendrite failing to run Sytest at the moment.


DMRobertson (approved):
I'm happy if dendrite CI is!

PR #1353, opened 2023-06-01
Fix unsafe assertions for media api tests

joshqou:
Fix assertions for unsafe behaviour blocking [matrix-org/synapse/#15680](https://github.com/matrix-org/synapse/pull/15680).

PR #1355, opened 2023-06-15
Change expectations of filenameless `Content-Disposition`

reivilibre:
Follows #1353 

For the 'no filename' test, SyTest was expecting no `Content-Disposition` header. Given what we want to achieve, I propose we instead expect a value of `attachment` (without a filename part).

### DIFF
--- a/tests/51media/02nofilename.pl
+++ b/tests/51media/02nofilename.pl
@@ -24,7 +24,7 @@ sub test_using_client
 
       # Require `attachment` `Content-Disposition` without a filename portion.
       # See `get_media`: {} means attachment with no key-value components in the header value.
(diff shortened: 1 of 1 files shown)


erikjohnston:
@reivilibre are the failures because we have a branch on Synapse with the same branch name?

joshqou:
@erikjohnston @reivilibre Seems to be. Branch isn't up to date with develop, so pre-hotserve pr

reivilibre:
Silly me not cutting the branch on the latest develop :facepalm: — it passes in https://github.com/matrix-org/synapse/pull/15784 so I'm happy to believe this is OK now.

PR #1334, opened 2023-02-24
Synapse: fix the build and reduce logging noise

DMRobertson:
> a42e75a3 origin/dmr/sytest-tweaks Limit number of haproxy connections

Suppresses big scary warnings of the form 

```
WARN: server process exited 1
[server]: [NOTICE]   (2457) : haproxy version is 2.6.7-1
[server]: [NOTICE]   (2457) : path to executable is /usr/sbin/haproxy
[server]: [ALERT]    (2457) : Not enough memory to allocate 1073741815 entries for fdtab!
[server]: [ALERT]    (2457) : No polling mechanism available.
[server]:   It is likely that haproxy was built with TARGET=generic and that FD_SETSIZE
[server]:   is too low on this platform to support maxconn and the number of listeners
[server]:   and servers. You should rebuild haproxy specifying your system using TARGET=
[server]:   in order to support other polling systems (poll, epoll, kqueue) or reduce the
[server]:   global maxconn setting to accommodate the system's limitation. For reference,
[server]:   FD_SETSIZE=1024 on this system, global.maxconn=536870884 resulting in a maximum of
[server]:   1073741815 file descriptors. You should thus reduce global.maxconn by 536870396. Also,
[server]:   check build settings using 'haproxy -vv'.
[server]: 
WARN: Error starting server-0: Process died without becoming connectable at tests/05homeserver.pl line 126.
WARN: Aborting test run due to failure to start test server at tests/05homeserver.pl line 130.
WARN: server process exited 1
[server]: [NOTICE]   (2466) : haproxy version is 2.6.7-1
[server]: [NOTICE]   (2466) : path to executable is /usr/sbin/haproxy
[server]: [ALERT]    (2466) : Not enough memory to allocate 1073741815 entries for fdtab!
[server]: [ALERT]    (2466) : No polling mechanism available.
[server]:   It is likely that haproxy was built with TARGET=generic and that FD_SETSIZE
[server]:   is too low on this platform to support maxconn and the number of listeners
[server]:   and servers. You should rebuild haproxy specifying your system using TARGET=
[server]:   in order to support other polling systems (poll, epoll, kqueue) or reduce the
[server]:   global maxconn setting to accommodate the system's limitation. For reference,
[server]:   FD_SETSIZE=1024 on this system, global.maxconn=536870884 resulting in a maximum of
[server]:   1073741815 file descriptors. You should thus reduce global.maxconn by 536870396. Also,
[server]:   check build settings using 'haproxy -vv'.
[server]: 
```

[...]

PR #1358, opened 2023-07-07
Drop old oldstable buster and sub new oldstable bullseye

H-Shay:
As part of dropping support for python 3.7, drop debian buster (python version 3.7) and sub in debian bullseye (python 3.9). 

Corresponding Synapse PR's: https://github.com/matrix-org/synapse/pull/15892, https://github.com/matrix-org/synapse/pull/15893

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,8 +25,8 @@ jobs:
         include:
           - base_image: ubuntu:focal
             tag: focal
-          - base_image: debian:buster
-            tag: buster
+          - base_image: debian:bullseye
+            tag: bullseye
           - base_image: debian:testing
             tag: testing
     steps:
@@ -69,10 +69,10 @@ jobs:
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:focal"
             build_args: "SYTEST_IMAGE_TAG=focal"
-          - sytest_image_tag: buster
+          - sytest_image_tag: bullseye
             dockerfile: synapse
-            tags: "matrixdotorg/sytest-synapse:buster"
-            build_args: "SYTEST_IMAGE_TAG=buster"
+            tags: "matrixdotorg/sytest-synapse:bullseye"
+            build_args: "SYTEST_IMAGE_TAG=bullseye"
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
--- a/docker/synapse.Dockerfile
+++ b/docker/synapse.Dockerfile
@@ -1,4 +1,4 @@
-ARG SYTEST_IMAGE_TAG=buster
+ARG SYTEST_IMAGE_TAG=bullseye
 
 FROM matrixdotorg/sytest:${SYTEST_IMAGE_TAG}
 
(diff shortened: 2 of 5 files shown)


docker/synapse.Dockerfile
```diff
@@ -21,13 +21,9 @@ RUN mkdir /rust /cargo
 
 RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain stable --profile minimal
 
-# Use the latest version of pip. This pulls in fixes not present in the
-# pip version provided by Debian Buster. See
-# https://github.com/pypa/setuptools/issues/3457#issuecomment-1190125849
 # For now, we need to tell Debian we don't care that we're editing the system python
 # installation.
 # Some context in https://github.com/pypa/pip/issues/11381#issuecomment-1399263627
-RUN ${PYTHON_VERSION} -m pip install -q --upgrade pip ${SYSTEM_PIP_INSTALL_SUFFIX}
```
H-Shay, line 30: My thinking on dropping this my reading of the issue mentioned indicates that the version of pip shipped in bullseye (20.3.4-4) should be sufficient as it is greater than 19?

H-Shay:
I did a manual run of the build pipeline (on this new branch) to verify that everything builds properly: https://github.com/matrix-org/sytest/actions/runs/5489675592/jobs/10005024592

.github/workflows/docker.yml
```diff
@@ -69,10 +69,10 @@ jobs:
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:focal"
             build_args: "SYTEST_IMAGE_TAG=focal"
-          - sytest_image_tag: buster
+          - sytest_image_tag: bullseye
             dockerfile: synapse
-            tags: "matrixdotorg/sytest-synapse:buster"
-            build_args: "SYTEST_IMAGE_TAG=buster"
+            tags: "matrixdotorg/sytest-synapse:bullseye"
+            build_args: "SYTEST_IMAGE_TAG=bullseye"
```
DMRobertson, line 75: I'm somewhat surprised that we don't need `
SYSTEM_PIP_INSTALL_SUFFIX=--break-system-packages
` here, but if it works then LGTM.

docker/synapse.Dockerfile
DMRobertson: The change seems sane to me (even though the sytest docker machinery does not).

I don't follow what is meant by

> For now, we need to tell Debian we don't care that we're editing the system python installation. Some context in https://github.com/pypa/pip/issues/11381#issuecomment-1399263627

which is amusing because I wrote the comment. https://github.com/matrix-org/sytest/pull/1334 and https://github.com/matrix-org/sytest/pull/1339 shed some light.

I think it's referring to `SYSTEM_PIP_INSTALL_SUFFIX`, and I think we probably need to leave this in place for now (unless every distribution is happy with the `--break-system-packages` option).

DMRobertson (approved):

docker/synapse.Dockerfile
H-Shay: Just to be clear, when you refer to "I think we probably need to leave this in place for now", do you mean the `SYSTEM_PIP_INSTALL_SUFFIX` on line 31/27, or are you referring to the removal on line 30? (My assumption is that you mean the former but just wanted to be sure)

.github/workflows/docker.yml
H-Shay: My assumption is that it works? The dockerfiles seemed to build but I could be missing something here?

docker/synapse.Dockerfile
DMRobertson: >  do you mean the SYSTEM_PIP_INSTALL_SUFFIX on line 31/27, or are you referring to the removal on line 30?

Sorry: the former, keeping SYSTEM_PIP_INSTALL_SUFFIX as the diff is currently.

.github/workflows/docker.yml
DMRobertson: IIRC I added the `SYSTEM_PIP_INSTALL_SUFFIX` stuff to get sytest working under debian testing. I had assumed bullseye was brand-new latest release and so would need this too, but that's not the case: [bullseye is nearly two years old](https://endoflife.date/debian).

So yeah, given that the docker images build this is fine. Sorry for the noise.

docker/synapse.Dockerfile
H-Shay: Ok great thanks!

.github/workflows/docker.yml
H-Shay: Not at all I appreciate the insight!

H-Shay:
Merged #1358 into develop, thanks for the review!

PR #1359, opened 2023-07-21
Allow processing of device lists to be async

erikjohnston:
c.f. https://github.com/matrix-org/synapse/pull/15975


clokep (approved):
Dendrite failures look unrelated?

erikjohnston:
> Dendrite failures look unrelated?

Yup!

PR #1361, opened 2023-07-26
Add test to ensure push/notifications work after room purge

erikjohnston:
c.f. https://github.com/matrix-org/synapse/issues/13476#issuecomment-1551009569

### DIFF
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -557,7 +557,7 @@ test "After /purge_history users still get pushed for new messages",
          # Test that /sync with an existing token still works.
          matrix_sync_again( $admin )
       })->then( sub {
-         # Check that the notifiication count increases if we see a new message.
+         # Check that the notification count increases if we see a new message.
          matrix_send_room_text_message_synced( $user, $room_id,
             body => "New Message",
          )
@@ -575,7 +575,7 @@ test "After /purge_history users still get pushed for new messages",
          # the first being $last_event_id and the second the "new message"
          assert_eq( $room->{unread_notifications}{"notification_count"}, 2 );
 
-         pass( "Notificaiton count was increased on new message" );
+         pass( "Notification count was increased on new message" );
 
          # Check that we get pushed messages
          do_request_json_for(
--- a/tests/48admin.pl
+++ b/tests/48admin.pl
@@ -498,3 +498,121 @@ sub await_message_in_room
       log_if_fail "Found event $event_id for $user_id";
    })
 }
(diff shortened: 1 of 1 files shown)


tests/48admin.pl
```diff
@@ -498,3 +498,121 @@ sub await_message_in_room
...
+         my $purge_id = $body->{purge_id};
+         await_purge_complete( $admin, $purge_id );
+      })->then( sub {
+         my ( $purge_status ) = @_;
+         assert_eq( $purge_status, 'complete' );
+
+         # Test that /sync with an existing token still works.
+         matrix_sync_again( $admin )
+      })->then( sub {
+         # Check that the notifiication count increases if we see a new message.
```
clokep, line 560:
```suggestion
         # Check that the notification count increases if we see a new message.
```

tests/48admin.pl
```diff
@@ -498,3 +498,121 @@ sub await_message_in_room
...
+         my ( $body ) = @_;
+
+         assert_json_keys( $body->{rooms}{join}, $room_id );
+         my $room =  $body->{rooms}{join}{$room_id};
+
+         log_if_fail( "Room", $room );
+
+         # We expect notification count to be 2, as we'll still see two events:
+         # the first being $last_event_id and the second the "new message"
+         assert_eq( $room->{unread_notifications}{"notification_count"}, 2 );
```
clokep, line 576: So do we expect a purge to force re-calculation of notification counts?

tests/48admin.pl
```diff
@@ -498,3 +498,121 @@ sub await_message_in_room
...
+
+         pass( "Notificaiton count was increased on new message" );
```
clokep, line 578:
```suggestion
         pass( "Notification count was increased on new message" );
```

tests/48admin.pl
erikjohnston: Hmm, good question. I guess what I really want to test is that the notification count *increases*. Though given this is a Synapse only test I'm not hugely bothered if we test Synapse specific stuff here.

clokep: I guess that makes sense. I think if you still have a event push summary on a purged event then it will increase properly? Not 100% sure on that though.

clokep (approved):

clokep:
Looks like a test might have flaked though?

erikjohnston:
> Looks like a test might have flaked though?

Ugh, yes. I'm failing to track that down.

PR #1362, opened 2023-07-27
Update appservice tests to check for access token in header vs query params

H-Shay:
Synapse v1.81.0 deprecated application service authorization via query parameters - authorization should now be sent via Authorization header. This PR (I hope, Perl is not my native language) changes two appservice tests to stop checking the query params for the access token and to check for them instead in the Authorization header.

### DIFF
--- a/tests/60app-services/05lookup3pe.pl
+++ b/tests/60app-services/05lookup3pe.pl
@@ -147,13 +147,15 @@ test "HS will proxy request for 3PU mapping",
             my ( $request ) = @_;
             
             my $access_token = $appserv1->info->hs2as_token;
+            my $auth_header = $request->header("Authorization");
+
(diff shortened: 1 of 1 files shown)


H-Shay:
Seems that the Dendrite failures are not related?

DMRobertson:
Sanity check: the spec [says](https://spec.matrix.org/v1.7/application-service-api/#identity-assertion)

> The access_token should be supplied through the Authorization header where possible to prevent the token appearing in HTTP request logs by accident.

DMRobertson:
> Seems that the Dendrite failures are not related?

I concur, looks like Dendrite is failing on the main branch with the same set of tests.

DMRobertson (approved):
Seems sane, and I can see similar snippets in sytest elsewhere.

H-Shay:
Merged #1362 into develop, thanks for the review!

PR #1365, opened 2023-08-04
Related to #1362, the access token is in both query params and headers

MatMaul:


DMRobertson:
I'm not following this---we don't want to be sending access tokens as query parameters?

MatMaul:
🤦 this is Indeed useless, after rebasing my PR all is fine.

PR #1367, opened 2023-08-09
Remove media tests supported by complement

Half-Shot:
As discussed in https://github.com/matrix-org/sytest/pull/1360


clokep (approved):
Looks like matrix-org/complement#279 added the remainder of these.

PR #1372, opened 2023-11-23
Build a `sytest-synapse` docker image with 3.12

erikjohnston:
Mainly to see if it speeds up synapse CI when using asyncio


DMRobertson (approved):
Assuming the CI failures are flakes

PR #1374, opened 2024-02-20
Fix lazy-loading test which was looking for the wrong thing

richvdh:
As the comment says, this test was looking for the wrong thing, and actually enforcing buggy behaviour.


richvdh:
Looks like sytest's CI is broken :(

S7evinK:
Looks like it is still trying to download the archive from the matrix-org/synapse repository:
https://github.com/matrix-org/sytest/blob/e2c4b0ea8136e754ec0423a3ccdbe5650475d204/.github/workflows/pipeline.yml#L92

erikjohnston (approved):

PR #1378, opened 2024-04-12
Download Synapse from correct repo in Dockerfile

erikjohnston:
This will fix pre-installing of dependencies, which will improve how quickly the tests are run.


anoadragon453 (approved):
Oops, that went unnoticed for a while. Nice catch.

PR #1379, opened 2024-05-14
Send `/{send,make}_knock` to Synapse workers

anoadragon453:
These endpoints should be handled by workers correctly.

To test this configuration, before it's recommended in https://github.com/element-hq/synapse/pull/17058.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -1240,7 +1240,7 @@ sub generate_haproxy_map
    return <<'EOCONFIG';
 
 ^/_matrix/client/(v2_alpha|r0|v3)/sync$                  synchrotron
-^/_matrix/client/(api/v1|v2_alpha|r0)/events$         synchrotron
+^/_matrix/client/(api/v1|v2_alpha|r0)/events$            synchrotron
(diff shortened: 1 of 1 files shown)


lib/SyTest/Homeserver/Synapse.pm
```diff
@@ -1257,6 +1257,8 @@ sub generate_haproxy_map
 ^/_matrix/federation/v1/make_leave/                   federation_reader
 ^/_matrix/federation/v1/send_join/                    federation_reader
 ^/_matrix/federation/v1/send_leave/                   federation_reader
+^/_matrix/federation/(v1|v2)/make_knock/              federation_reader
+^/_matrix/federation/(v1|v2)/send_knock/              federation_reader
```
clokep, line 1261: I think only a v1 version exists? (Same goes for complement and the docs btw.)

anoadragon453: D'oh, good catch. Copy/paste fail.

clokep: Although shouldn't the join, leave, etc ones include v2? 🤔

anoadragon453: Another good catch! I've corrected in https://github.com/matrix-org/sytest/pull/1382.

erikjohnston (approved):

PR #1380, opened 2024-05-14
Update Synapse Admin API quarantine path

anoadragon453:
This path was deprecated nearly 4 years ago in https://github.com/matrix-org/synapse/pull/6681.

### DIFF
--- a/tests/51media/48admin-quarantine.pl
+++ b/tests/51media/48admin-quarantine.pl
@@ -44,7 +44,7 @@ multi_test "Can quarantine media in rooms",
       })->then( sub {
          do_request_json_for( $admin,
             method   => "POST",
-            full_uri => "/_synapse/admin/v1/quarantine_media/$room_id",
+            full_uri => "/_synapse/admin/v1/room/$room_id/media/quarantine",
             content  => {},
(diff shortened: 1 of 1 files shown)


erikjohnston (approved):

anoadragon453:
That test failure exists on `develop` as well, and isn't related to this PR. However we should try to fix it separately.

anoadragon453:
The problem is that the Admin API request is going to the main process, whereas it should be going to the media worker. As we disable media functionality, including media-related Admin API functionality, in Sytest's Synapse when workers are enabled:

https://github.com/matrix-org/sytest/blob/7d3f92d4c2083a3efd16ad5c4046f6c91acfc5e8/lib/SyTest/Homeserver/Synapse.pm#L265

https://github.com/element-hq/synapse/blob/1e68b56a623a0d7f98106482bfed223c22193050/synapse/rest/admin/__init__.py#L351-L352

anoadragon453:
> The problem is that the Admin API request is going to the main process, whereas it should be going to the media worker.

Should be fixed by https://github.com/matrix-org/sytest/pull/1381.

PR #1390, opened 2024-12-02
Update configurations to recognize `dbname` instead of `database`

realtyem:
In preparation of a new database driver for Synapse, let's stop mutating the configuration after it's generated. Since `dbname` is what is already used by libpq and is what is expected by Synapse's configuration parsing, just use this. In Dendrite's case, it's not used directly but is parsed and turned into a DSN to feed to it's command line invocation.

This does not modify the pre-existing behavior of adding an additional attribute to that part of the yaml block as it did before: `type: pg` or `type: sqlite`, depending. It appears this is only used to identify which method to invoke to pre-clear the database before testing starts and is otherwise harmless.

### DIFF
--- a/lib/SyTest/Homeserver.pm
+++ b/lib/SyTest/Homeserver.pm
@@ -409,7 +409,7 @@ sub _check_db_config
 
    my $db_type = $db_config{type};
    if( $db_type eq 'pg' ) {
-      foreach (qw( database )) {
+      foreach (qw( dbname )) {
          if( !$db_config{args}->{$_} ) {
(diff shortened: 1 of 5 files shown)


clokep:
This is essentially matrix-org/synapse#16618, but for sytest configs.

anoadragon453 (approved):
Thanks for cleaning this up!

erikjohnston:
(Heads up this broke sqlite synapse CI: https://github.com/matrix-org/sytest/pull/1392)

PR #1391, opened 2024-12-12
Drop invalid PDUs instead of erroring. Implements element-hq/synapse#7543

morguldir:
Relevant Synapse PR: https://github.com/element-hq/synapse/pull/17893


anoadragon453:
Taking this off the review queue until the changes mentioned at https://github.com/element-hq/synapse/pull/17893#issuecomment-3530202423 are addressed.

PR #1392, opened 2024-12-20
Fix SQLite Synapse runs

erikjohnston:
Broke in https://github.com/matrix-org/sytest/pull/1390. This PR adds back in some deleted code, as well as enables SQLite CI.


anoadragon453 (approved):
Thanks for fixing this and adding a test for Synapse SQLite.

LGTM assuming CI passes.

PR #1394, opened 2025-01-24
Synapse: increase presence ratelimit for tests

anoadragon453:
This ratelimit is introduced in https://github.com/element-hq/synapse/pull/18000 and causes some existing Sytests to fail for Synapse.

Increase the ratelimit in Sytest's Synapse config so tests pass again.

---

Note: I've re-pushed https://github.com/element-hq/synapse/pull/18000's branch to https://github.com/element-hq/synapse/tree/set-presence-ratelimit so that the Sytest PR's CI will test that branch instead of `develop`.

### DIFF
--- a/lib/SyTest/Homeserver/Synapse.pm
+++ b/lib/SyTest/Homeserver/Synapse.pm
@@ -229,6 +229,13 @@ sub start
            },
         },
 
+        rc_presence => {
+           per_user => {
+              per_second => 1000,
+              burst_count => 1000,
+           },
+        },
+
         enable_registration => "true",
         enable_registration_without_verification => "true",
         databases => \%db_configs,


devonh (approved):
Shoud these be the same values as used for complement tests though?
ie. 9999 (https://github.com/element-hq/synapse/pull/18000/commits/62b5e967b8b7104338a829fdb7227c6355a3cdaf)

anoadragon453:
> Shoud these be the same values as used for complement tests though? ie. 9999 ([element-hq/synapse@62b5e96](https://github.com/element-hq/synapse/commit/62b5e967b8b7104338a829fdb7227c6355a3cdaf))

Nah, 1000/second is more than enough for these tests (that typically only send <5 presence events). it may be nice to unify with Complement for consistency - but that should be a separate PR.

PR #1399, opened 2025-04-01
Check that invalid PDUs are ignored, instead of erroring out

sandhose:
This is similar to #1391, but it doesn't assert on the /send body response, and rather looks at events persisted through /sync and /events

### DIFF
--- a/tests/50federation/51transactions.pl
+++ b/tests/50federation/51transactions.pl
@@ -58,6 +58,8 @@ test "Server correctly handles transactions that break edu limits",
 test "Server discards events with invalid JSON in a version 6 room",
    requires => [ $main::OUTBOUND_CLIENT,
                  federated_rooms_fixture( room_opts => { room_version => "6" } ) ],
+   # This behaviour has only been changed in Synapse, not Dendrite
+   implementation_specific => ['synapse'],
 
    do => sub {
       my ( $outbound_client, $creator, $user_id, @rooms ) = @_;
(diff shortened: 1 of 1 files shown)


tests/50federation/51transactions.pl
```diff
@@ -55,31 +55,62 @@
 # sent. It is unclear if this is the correct behavior or not.
```
anoadragon453, line 55: Could you update this comment as well? Perhaps say that the transaction should succeed, but it's unclear whether the response body should contain the invalid event.

tests/50federation/51transactions.pl
```diff
@@ -55,31 +55,62 @@
...
+             body    => "Bad event",
              # Insert a "bad" value into the PDU, in this case a float.
              bad_val => 1.1,
           },
       );
 
-      my @pdus = ( $bad_event );
+      my @pdus = ( $good_event, $bad_event );
 
       # Send the transaction to the client and expect a fail
```
anoadragon453, line 90:
```suggestion
      # Send the transaction to the client and expect a 200
```

tests/50federation/51transactions.pl
```diff
@@ -55,31 +55,62 @@
...
           pdus => \@pdus,
           destination => $creator->server_name,
-      )->main::expect_m_bad_json;
+      )->then(sub {
+         # Wait for the good event to be sent down through sync
+         await_sync_timeline_contains( $creator, $room_id, check => sub {
+            my ( $event ) = @_;
+            $event->{type} eq "m.room.message" &&
+               $event->{content}{body} eq "Good event"
+         })
```
anoadragon453, line 100: Can we also check that we *don't* receive the bad event down sync?

anoadragon453 (changes requested):
Thanks for taking this on. I'm curious if this test now fails for other homeserver implementations...

tests/50federation/51transactions.pl
```diff
@@ -55,31 +55,62 @@
...
+             uri     => "/v3/rooms/$room_id/event/$event_id",
+         )
+      })->then(sub {
+         # Check that we have ignored the bad event PDU
+         my $event_id = $room->id_for_event( $bad_event );
+         do_request_json_for( $creator,
+             method  => "GET",
+             uri     => "/v3/rooms/$room_id/event/$event_id",
+         )->main::expect_m_not_found
+      });
```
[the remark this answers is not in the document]
anoadragon453: That's an open question; the sending homeserver technically can't do anything with that information.

tests/50federation/51transactions.pl
sandhose: I couldn't find utilities to make sure that the sync timeline *doesn't* contain something :(

tests/50federation/51transactions.pl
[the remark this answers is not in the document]
sandhose: Fixed in 5359ab14

tests/50federation/51transactions.pl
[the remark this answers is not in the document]
sandhose: This is why we don't yet assert on the /send response body

tests/50federation/51transactions.pl
sandhose: Done in https://github.com/matrix-org/sytest/commit/5359ab1401bbc114419068512f05b5c25d82a499 and in 842e5b71

sandhose:
> I'm curious if this test now fails for other homeserver implementations...

Dendrite indeed fails. I've made implementation specific tests in [a10eebc](https://github.com/matrix-org/sytest/pull/1399/commits/a10eebc5b794ecb66719b0570479fbbea6ca3ca9)

tests/50federation/51transactions.pl
anoadragon453: Wouldn't it be just adding `&& $event->{content}{body} neq "Good event"`?

anoadragon453 (commented):
This LGTM other than potentially fixing the nit in https://github.com/matrix-org/sytest/pull/1399#discussion_r2022690419.

Dendrite appears to still be failing, as it's failing to find this test? https://github.com/matrix-org/sytest/actions/runs/14194392227/job/39766068369?pr=1399#step:6:9

tests/50federation/51transactions.pl
sandhose: It only checks if one event in the timeline checks that constraint

sandhose:
@anoadragon453 fixed the test, that was a typo [7e2a47e](https://github.com/matrix-org/sytest/pull/1399/commits/7e2a47ee63487745438a5d222555c83568d54ca8)

anoadragon453 (approved):

PR #1408, opened 2025-07-04
Run Synapse tests on Debian Bullseye (Python 3.9, PG 13)

sandhose:
The images we build (the `docker.yml` workflow) and the ones we use (the `pipeline.yml` workflow) were out of sync, which broke in recent Synapse, as it had old versions of rustc in them.


reivilibre (approved):
Just noting it's EOL in 3 months

PR #1411, opened 2025-10-07
Update uses of `matrix_put_e2e_keys` to contain valid data

anoadragon453:
After adding validation to `/keys/upload` on Synapse, uses of this function with mostly-empty bodies were no longer valid.

This change adds fake test data to them where necessary.

---

This is paired with https://github.com/element-hq/synapse/pull/17097, which was accidentally merged before these tests were.


reivilibre (approved):

anoadragon453:
This commit has been backported to `release-v1.138` and `release-v1.139` branches.

PR #1412, opened 2025-10-20
Fix request body structure for "Should reject keys claiming to belong to a different user"

anoadragon453:
This was causing Synapse to respond with validation errors for these tests:

```
2025-10-18 09:24:32,660 - synapse.http.server - 135 - INFO - POST-6877 - <SynapseRequest at 0x7f071c58d9d0 method='POST' uri='/_matrix/client/v3/keys/upload?access_token=<redacted>' clientproto='HTTP/1.1' site='8800'> SynapseError: 400 - 3 validation errors for KeyUploadRequestBody
device_keys.algorithms
  Field required [type=missing, input_value={'user_id': '@anon-202510...evice_id': 'TRIISNQZFQ'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
device_keys.keys
  Field required [type=missing, input_value={'user_id': '@anon-202510...evice_id': 'TRIISNQZFQ'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
device_keys.signatures
  Field required [type=missing, input_value={'user_id': '@anon-202510...evice_id': 'TRIISNQZFQ'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
```

...which caused the test to fail (which we were expecting). Yet, we were looking for failure due to the wrong user ID being passed, not the request body's structure.

This PR fixes up the structure of the request body to avoid the validation error. We now see the correct error in the logs:

```
2025-10-20 17:09:20,357 - synapse.http.server - 135 - INFO - POST-3 - <SynapseRequest at 0x7f091c1570d0 method='POST' uri='/_matrix/client/v3/keys/upload?access_token=<redacted>' clientproto='HTTP/1.1' site='8800'> SynapseError: 400 - Provided `user_id` in `device_keys` does not match that of the authenticated user
```

Ideally we would differentiate the errors based on `errcode`, but [the spec](https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3keysupload) doesn't currently define any for this endpoint.

### DIFF
--- a/tests/41end-to-end-keys/01-upload-key.pl
+++ b/tests/41end-to-end-keys/01-upload-key.pl
@@ -91,6 +91,12 @@ test "Should reject keys claiming to belong to a different user",
             device_keys => {
                user_id => "\@50-e2e-alice:localhost:8480",
                device_id => "alices_first_device",
+               algorithms => ["m.olm.curve25519-aes-sha256", "m.megolm.v1.aes-sha"],
+               keys => {
+                 "curve25519:".$user->device_id => "curve25519+key",
+                 "ed25519:".$user->device_id => "ed25519+key",
+               },
+               signatures => {},
             },
          }
       )->main::expect_http_4xx;


tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -91,6 +91,12 @@
             device_keys => {
                user_id => "\@50-e2e-alice:localhost:8480",
                device_id => "alices_first_device",
+               algorithms => ["m.olm.curve25519-aes-sha256", "m.megolm.v1.aes-sha"],
+               keys => {
+                 "curve25519:".$user->device_id => "curve25519+key",
+                 "ed25519:".$user->device_id => "ed25519+key",
+               },
+               signatures => {},
             },
```
MadLittleMods, line 100: This is caused by changes from https://github.com/element-hq/synapse/pull/17097 / https://github.com/element-hq/synapse/pull/19023 ?

I'm a bit confused about how Sytest is passing on Synapse `develop` if it's from those PRs.

Or since it's optional, why it's failing in (unknown, as described in PR title)?

tests/41end-to-end-keys/01-upload-key.pl
```diff
@@ -91,6 +91,12 @@
             device_keys => {
                user_id => "\@50-e2e-alice:localhost:8480",
                device_id => "alices_first_device",
+               algorithms => ["m.olm.curve25519-aes-sha256", "m.megolm.v1.aes-sha"],
+               keys => {
+                 "curve25519:".$user->device_id => "curve25519+key",
+                 "ed25519:".$user->device_id => "ed25519+key",
+               },
+               signatures => {},
```
MadLittleMods, line 99: These fields make sense in terms of [Pydantic model](https://github.com/element-hq/synapse/blob/fc244bb592aa481faf28214a2e2ce3bb4e95d990/synapse/rest/client/keys.py#L124-L224) in Synapse (see `DeviceKeys`)

Feel free to merge if things make sense to you. The fields themselves seem sane.

tests/41end-to-end-keys/01-upload-key.pl
anoadragon453: The test was passing because it expects a `4xx` error:

```
)->main::expect_http_4xx;
```

but, as explained in the PR description, we were getting a 400 due to the Pydantic model validation fail. Whereas the test is trying to check that the homeserver will return a 400 when uploading device keys for a user other than yourself.

anoadragon453:
Dendrite test failures appear unrelated to this PR.

PR #1413, opened 2025-10-20
Remove `Name/topic keys are correct` test

anoadragon453:
This test has been migrated to Complement, see https://github.com/matrix-org/complement/pull/811.

This PR should only be merged once the equivalent Complement test has been.


MadLittleMods (approved):
Approving as a no-brainer once https://github.com/matrix-org/complement/pull/811 is merged

PR #1414, opened 2025-10-20
Use bookworm; target specific Python & PostgreSQL

AndrewFerr:
Update dependencies to comply with Synapse's deprecation policy: https://element-hq.github.io/synapse/latest/deprecation_policy.html

- Upgrade to Debian Bookworm (the current oldstable) to pull in the oldest supported version of SQLite.
- Instead of installing PostgreSQL from Debian repositories, use the PostgreSQL Apt Repository to install its lowest-supported version.
- Instead of installing Python from Debian repositories, use uv to install its lowest-supported version.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -74,7 +74,7 @@ jobs:
             tags: "matrixdotorg/sytest-synapse:testing"
             build_args: |
               SYTEST_IMAGE_TAG=testing
-              PYTHON_VERSION=3.14
+              PYTHON_VERSION=3.13
 
     steps:
       - name: Set up QEMU
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,11 +26,9 @@ jobs:
           - base_image: debian:bookworm
             tag: bookworm
             postgresql_version: 13
-            python_version: python3.9
           - base_image: debian:testing
             tag: testing
             postgresql_version: 18
-            python_version: python3.14
     steps:
       - name: Set up QEMU
         id: QEMU
@@ -74,13 +72,13 @@ jobs:
             tags: "matrixdotorg/sytest-synapse:bookworm"
             build_args: |
               SYTEST_IMAGE_TAG=bookworm
-              PYTHON_VERSION=${{ matrix.python_version }}
+              PYTHON_VERSION=python3.9
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
             build_args: |
               SYTEST_IMAGE_TAG=testing
-              PYTHON_VERSION=${{ matrix.python_version }}
+              PYTHON_VERSION=python3.14
 
     steps:
       - name: Set up QEMU
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,10 +23,14 @@ jobs:
     strategy:
       matrix:
         include:
-          - base_image: debian:bullseye
-            tag: bullseye
+          - base_image: debian:bookworm
+            tag: bookworm
+            postgresql_version: 13
+            python_version: python3.9
           - base_image: debian:testing
             tag: testing
+            postgresql_version: 18
+            python_version: python3.14
     steps:
       - name: Set up QEMU
         id: QEMU
@@ -53,7 +57,9 @@ jobs:
           platforms: linux/amd64,linux/arm64
           labels: "gitsha1=${{ github.sha }}"
           file: docker/base.Dockerfile
-          build-args: "BASE_IMAGE=${{ matrix.base_image }}"
+          build-args: |
+            BASE_IMAGE=${{ matrix.base_image }}
+            POSTGRESQL_VERSION=${{ matrix.postgresql_version }}
           tags: matrixdotorg/sytest:${{ matrix.tag }}
 
   build-dependent-images:
@@ -63,16 +69,18 @@ jobs:
     strategy:
       matrix:
         include:
-          - sytest_image_tag: bullseye
+          - sytest_image_tag: bookworm
             dockerfile: synapse
-            tags: "matrixdotorg/sytest-synapse:bullseye"
-            build_args: "SYTEST_IMAGE_TAG=bullseye"
+            tags: "matrixdotorg/sytest-synapse:bookworm"
+            build_args: |
+              SYTEST_IMAGE_TAG=bookworm
+              PYTHON_VERSION=${{ matrix.python_version }}
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
             build_args: |
               SYTEST_IMAGE_TAG=testing
-              SYSTEM_PIP_INSTALL_SUFFIX=--break-system-packages
+              PYTHON_VERSION=${{ matrix.python_version }}
 
     steps:
       - name: Set up QEMU
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,12 +23,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - base_image: debian:bookworm
-            tag: bookworm
-            postgresql_version: 13
-          - base_image: debian:testing
-            tag: testing
-            postgresql_version: 18
+          - tag: bookworm
+            build_args: "BASE_IMAGE=debian:bookworm"
+          - tag: testing
+            build_args: |
+              BASE_IMAGE=debian:testing
+              POSTGRESQL_VERSION=18
     steps:
       - name: Set up QEMU
         id: QEMU
@@ -55,9 +55,7 @@ jobs:
           platforms: linux/amd64,linux/arm64
           labels: "gitsha1=${{ github.sha }}"
           file: docker/base.Dockerfile
-          build-args: |
-            BASE_IMAGE=${{ matrix.base_image }}
-            POSTGRESQL_VERSION=${{ matrix.postgresql_version }}
+          build-args: ${{ matrix.build_args }}
           tags: matrixdotorg/sytest:${{ matrix.tag }}
 
   build-dependent-images:
@@ -70,15 +68,13 @@ jobs:
           - sytest_image_tag: bookworm
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:bookworm"
-            build_args: |
-              SYTEST_IMAGE_TAG=bookworm
-              PYTHON_VERSION=python3.9
+            build_args: "SYTEST_IMAGE_TAG=bookworm"
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
             build_args: |
               SYTEST_IMAGE_TAG=testing
-              PYTHON_VERSION=python3.14
+              PYTHON_VERSION=3.14
 
     steps:
       - name: Set up QEMU
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,7 +24,9 @@ jobs:
       matrix:
         include:
           - tag: bookworm
-            build_args: "BASE_IMAGE=debian:bookworm"
+            build_args: |
+              BASE_IMAGE=debian:bookworm
+              POSTGRESQL_VERSION=13
           - tag: testing
             build_args: |
               BASE_IMAGE=debian:testing
@@ -68,7 +70,9 @@ jobs:
           - sytest_image_tag: bookworm
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:bookworm"
-            build_args: "SYTEST_IMAGE_TAG=bookworm"
+            build_args: |
+              SYTEST_IMAGE_TAG=bookworm
+              PYTHON_VERSION=3.10
           - sytest_image_tag: testing
             dockerfile: synapse
             tags: "matrixdotorg/sytest-synapse:testing"
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -21,23 +21,26 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - label: Debian Bullseye (Py 3.9, SQLite) Monolith
-            sytest-tag: bullseye
+          - label: Debian Bookworm (Py 3.9, SQLite 3.40) Monolith
+            sytest-tag: bookworm
 
-          - label: Debian Bullseye (Py 3.9, PG 13) Monolith
-            sytest-tag: bullseye
+          - label: Debian Bookworm (Py 3.9, PG 13) Monolith
+            sytest-tag: bookworm
             postgres: postgres
 
-          - label: Debian Bullseye (Py 3.9, PG 13) Workers
-            sytest-tag: bullseye
+          - label: Debian Bookworm (Py 3.9, PG 13) Workers
+            sytest-tag: bookworm
             postgres: postgres
             workers: workers
 
-          - label: Debian Testing (Py 3.13, PG 17), Monolith
+          - label: Debian Testing (Py 3.13, SQLite 3.46) Monolith
+            sytest-tag: testing
+
+          - label: Debian Testing (Py 3.13, PG 18), Monolith
             sytest-tag: testing
             postgres: postgres
 
-          - label: Debian Testing (Py 3.13, PG 17), Workers
+          - label: Debian Testing (Py 3.13, PG 18), Workers
             sytest-tag: testing
             postgres: postgres
             workers: workers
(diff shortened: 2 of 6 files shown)


AndrewFerr:
The CI tests depend on new `sytest` and `sytest-synapse` images to be built, as the tests run in containers of the published `matrixdotorg/sytest-*` images.  Is there a good way to use freshly-built images from the current PR instead?

-> https://github.com/matrix-org/sytest/pull/1414#discussion_r2452838083

docker/base.Dockerfile
```diff
@@ -20,14 +20,19 @@ RUN apt-get -qq update && apt-get -qq install -y \
...
     sqlite3 \
     wget \
     libicu-dev \
     pkg-config \
     && rm -rf /var/lib/apt/lists/*
 
+ARG POSTGRESQL_VERSION=13
+RUN /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y \
+    && apt -qq install -y postgresql-${POSTGRESQL_VERSION} \
+    && rm -rf /var/lib/apt/lists/*
```
MadLittleMods, line 38: > Instead of installing PostgreSQL from Debian repositories, use the PostgreSQL Apt Repository to install its lowest-supported version.

Do we do this anywhere else? I'd be more keen if this magic was already accepted elsewhere

docker/synapse.Dockerfile
```diff
@@ -21,10 +18,13 @@ RUN mkdir /rust /cargo
...
 
-# For now, we need to tell Debian we don't care that we're editing the system python
-# installation.
-# Some context in https://github.com/pypa/pip/issues/11381#issuecomment-1399263627
-RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.3.2 ${SYSTEM_PIP_INSTALL_SUFFIX}
+ARG PYTHON_VERSION=python3
+
+RUN --mount=type=bind,from=ghcr.io/astral-sh/uv:0.9.4,source=/uv,target=/bin/uv \
+        uv python install "$PYTHON_VERSION" && \
+        uv tool install poetry@1.3.2
```
MadLittleMods: > Instead of installing Python from Debian repositories, use uv to install its lowest-supported version.

Do we do this anywhere else? I'd be more keen if this magic was already accepted elsewhere

docker/synapse.Dockerfile
```diff
@@ -21,10 +18,13 @@ RUN mkdir /rust /cargo
...
 RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain stable --profile minimal
 
-# For now, we need to tell Debian we don't care that we're editing the system python
-# installation.
-# Some context in https://github.com/pypa/pip/issues/11381#issuecomment-1399263627
-RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.3.2 ${SYSTEM_PIP_INSTALL_SUFFIX}
+ARG PYTHON_VERSION=python3
+
+RUN --mount=type=bind,from=ghcr.io/astral-sh/uv:0.9.4,source=/uv,target=/bin/uv \
+        uv python install "$PYTHON_VERSION" && \
```
MadLittleMods: > Instead of installing Python from Debian repositories, use uv to install its lowest-supported version.

How does `uv python install python3` end up installing the minimum supported Python version which should be Python 3.9?

.github/workflows/docker.yml
```diff
@@ -23,10 +23,12 @@ jobs:
     strategy:
```
MadLittleMods, line 23: > The CI tests depend on new `sytest` and `sytest-synapse` images to be built, as the tests run in containers of the published `matrixdotorg/sytest-*` images.  Is there a good way to use freshly-built images from the current PR instead?

I'm not sure as I never deal with Sytest.

It looks like @H-Shay did this last -> https://github.com/matrix-org/sytest/pull/1358 - Any insight here?

.github/workflows/docker.yml
```diff
@@ -23,10 +23,12 @@ jobs:
...
       matrix:
         include:
-          - base_image: debian:bullseye
-            tag: bullseye
+          - base_image: debian:bookworm
+            tag: bookworm
+            postgresql_version: 13
```
MadLittleMods: How can I cross-verify these version numbers?

docker/synapse.Dockerfile
MadLittleMods: Ahh, I think `ARG PYTHON_VERSION=python3` is the default and we override in the CI matrix. See `PYTHON_VERSION=python3.9` above for example.

docker/base.Dockerfile
AndrewFerr: To my knowledge, we don't do it elsewhere, but this is the recommended strategy from upstream PostgreSQL docs for installing a specific version in Debian: https://www.postgresql.org/download/linux/debian/

docker/synapse.Dockerfile
AndrewFerr: Not sure if we use this exact strategy elsewhere, but I got the idea of doing this from Synapse's (relatively) recent usage of `uv` for its Docker image builds.

docker/synapse.Dockerfile
AndrewFerr: Exactly.  Since that caused some confusion, I'll refactor where/how this variable is set to make it more clear.

.github/workflows/docker.yml
AndrewFerr: https://endoflife.date/postgresql, which I found from [Synapse's deprecation policy docs](https://element-hq.github.io/synapse/latest/deprecation_policy.html).

The Debian tag used here is the codename of whatever `oldstable` is at the time of writing.  Maybe it'd be more convenient to use `oldstable`, but pinning on a specific version is more reliable.

For the Debian-provided version of PostgreSQL, see https://packages.debian.org/oldstable/postgresql, though that is less relevant when installing from the upstream PostgreSQL repo.

docker/base.Dockerfile
AndrewFerr: Ah, just found that it's also recommended in the Debian Wiki, as well: https://wiki.debian.org/PostgreSql%20%20#PGDG_Repository

.github/workflows/docker.yml
H-Shay: Hmm it's been quite awhile and I didn't do this regularly, but I think there was a way to manually trigger the build - https://github.com/matrix-org/sytest/pull/1358#issuecomment-1626054011 
I don't have the same access I did then so I can't look but I think there was a run button on one of the actions that could trigger this.

docker/synapse.Dockerfile
AndrewFerr: Done with a94a67bb

.github/workflows/docker.yml
MadLittleMods: On https://github.com/matrix-org/sytest/actions/workflows/pipeline.yml, I see a "Build and deploy docker images" workflow to run manually but not for a branch that is in a fork (as far I as I can tell).

I ended up doing this:

 1. `git fetch origin refs/pull/1414/head:pr-1414`
 1. `git checkout pr-1414`
 1. `git push --set-upstream origin pr-1414`
 1. On https://github.com/matrix-org/sytest/actions/workflows/docker.yml, run the "Build and deploy docker images" workflow for the `pr-1414` branch
 1. -> https://github.com/matrix-org/sytest/actions/runs/18729194682
  Looks like it mostly passed:
    - `Build sytest:bookworm` :white_check_mark: 
    - `Build sytest:testing` :white_check_mark: 
    - `Build sytest-synapse:bookworm` (canceled, why? -> `The strategy configuration was canceled because "build-dependent-images.testing_synapse_matrixdot" failed`)
    - `Build sytest-synapse:testing` :x:
    ```
    Error: buildx failed with: ERROR: failed to build: failed to solve: process "/bin/bash -o pipefail -c wget -q https://github.com/element-hq/synapse/archive/develop.tar.gz         -O /synapse.tar.gz &&         mkdir /synapse &&         tar -C /synapse --strip-components=1 -xf synapse.tar.gz &&         ln -s -T /venv /synapse/.venv &&         cd /synapse &&         poetry install -q --no-root --extras all &&         rm -rf `poetry config cache-dir` &&         rm -rf /synapse &&         rm /synapse.tar.gz" did not complete successfully: exit code: 1
    ```
     
 Perhaps we just need to re-run the CI here now - or fix up the builds and run them again.

docker/synapse.Dockerfile
```diff
@@ -21,10 +18,13 @@ RUN mkdir /rust /cargo
 
 RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain stable --profile minimal
 
-# For now, we need to tell Debian we don't care that we're editing the system python
-# installation.
-# Some context in https://github.com/pypa/pip/issues/11381#issuecomment-1399263627
-RUN ${PYTHON_VERSION} -m pip install -q --no-cache-dir poetry==1.3.2 ${SYSTEM_PIP_INSTALL_SUFFIX}
+ARG PYTHON_VERSION=3.9
```
MadLittleMods: I'm thinking we should maybe add a comment about this default being the minimum Python version supported according to the [Synapse deprecation policy](https://element-hq.github.io/synapse/latest/deprecation_policy.html).

Especially since this is documented in another project.

docker/base.Dockerfile
MadLittleMods: Let's cross-link those references in a comment :fast_forward: 

We should also explain why we care about installing that minimum version (the Synapse deprecation policy)

.github/workflows/docker.yml
```diff
@@ -63,16 +65,16 @@ jobs:
     strategy:
       matrix:
         include:
-          - sytest_image_tag: bullseye
+          - sytest_image_tag: bookworm
             dockerfile: synapse
-            tags: "matrixdotorg/sytest-synapse:bullseye"
-            build_args: "SYTEST_IMAGE_TAG=bullseye"
+            tags: "matrixdotorg/sytest-synapse:bookworm"
+            build_args: "SYTEST_IMAGE_TAG=bookworm"
```
MadLittleMods:
```suggestion
            build_args: |
              SYTEST_IMAGE_TAG=bookworm
              # This is the default from the image
              # PYTHON_VERSION=3.9
```

(more obvious)

docker/synapse.Dockerfile
MadLittleMods: I'm not sure if the minimum Python version is the best default to have 🤷 

Since this is just for testing, it's probably fine.

MadLittleMods: As I suspected, we use the [latest supported Python version ](https://github.com/element-hq/synapse/blob/cba3a814c63ad877c482df3bc75570e4a7d61ddb/docker/Dockerfile#L24) as the default in Synapse Docker image. But that's also used in production for people. And then we use that image to build the Complement image.

docker/synapse.Dockerfile
MadLittleMods: How do we handle this in Synapse for other things?

In CI, it looks like we set `python-version` and `matrix-org/setup-python-poetry` takes care of things.

In our Docker images, it looks like we use an image directly that bundles it in: [`FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-${DEBIAN_VERSION}`](https://github.com/element-hq/synapse/blob/cba3a814c63ad877c482df3bc75570e4a7d61ddb/docker/Dockerfile#L68)

Not sure how much effort it would be to just use the direct image that looks like it already using Debian and has the specified Python version bundled in.

It looks relatively sane to do it the way you have it :fast_forward:

AndrewFerr: The `uv:python3.X` images are derived from a base distribution, whereas the `uv:X.Y.Z` ones are distroless images that include just the `uv` binaries ([source](https://docs.astral.sh/uv/guides/integration/docker/#getting-started)).

Synapse's CI uses `uv` during its build steps, so it benefits from using a derived image with `uv` available as a build layer image base.

For sytest's CI, I thought it made more sense to use `uv` solely to "install" a specific Python version.

.github/workflows/docker.yml
AndrewFerr: The failure is from `poetry install`, but the details are suppressed by `-q`. Rebuilding locally without it shows these details:
```
error: the configured Python interpreter version (3.14) is newer than PyO3's maximum supported version (3.13)
= help: please check if an updated version of PyO3 is available. Current version: 0.24.1
= help: set PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 to suppress this check and build anyway using the stable ABI
```

Addressed by 1ab7a300

docker/base.Dockerfile
AndrewFerr: 4e15850c for repo usage

0c73ed7b for min version

docker/synapse.Dockerfile
AndrewFerr: 0c73ed7b

.github/workflows/docker.yml
AndrewFerr: If showing the default `PYTHON_VERSION` here is more obvious, would it be better to not give it a default at all, and rely on it being set in every CI matrix value?  (This applies to the `POSTGRES_VERSION` build arg too)

docker/synapse.Dockerfile
nico-famedly: It might make sense to go to 3.10 directly though, considering that 3.9 is EOL in a week. Unless someone will bump it then. Otherwise you will get conflicts when the minimum version is bumped in Synapse.

AndrewFerr: 9cd8e8b6

.github/workflows/docker.yml
MadLittleMods: Depends what perspective you're coming from. For CI, that makes sense.

What's the usual developer flow for running Sytest locally? If it works without more configuration for local developers, that sounds fine. I recall an option to run Sytest in Docker so it would be good to use the latest Python version in that case.

.github/workflows/docker.yml
MadLittleMods: New rebuild underway -> https://github.com/matrix-org/sytest/actions/runs/18752219369

docker/synapse.Dockerfile
MadLittleMods: I'm not following completely as we currently use Debian and `uv` to install a certain Python version. And it looks like we can use Debian-based images from `uv` with the Python version directly.

In any case, prior art :fast_forward:

.github/workflows/pipeline.yml
```diff
@@ -21,23 +21,26 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - label: Debian Bullseye (Py 3.9, SQLite) Monolith
-            sytest-tag: bullseye
+          - label: Debian Bookworm (Py 3.9, SQLite 3.40) Monolith
```
MadLittleMods: The minimum version was bumped to `3.10` now (https://github.com/matrix-org/sytest/pull/1414#discussion_r2454981332)

We should update these labels

docker/base.Dockerfile
MadLittleMods: We currently only link one of the references. Are both references useful? I haven't inspected them too closely but if I was tackling this task in the future, I would want to at-least look at both of these so having them to jump from quickly seems good to me.

AndrewFerr: The Debian Wiki page references a PostgreSQL wiki page, which has similar information to the docs page on www.postgresql.org, so I figured it was more efficient to link just the Debian Wiki page.

.github/workflows/pipeline.yml
AndrewFerr: Good catch. ebd42a60

.github/workflows/docker.yml
AndrewFerr: Presumably, developers wanting to run Sytest in Docker would just pull our prebuilt Sytest images instead of building them manually.  And anyone developing Sytest itself would probably run it natively.  So it should be rare for developers to need to build Sytest images & have to set missing build args themselves.

Though out of principle, I generally prefer setting good `ARG` defaults, especially since these Dockerfiles already give defaults to some other important `ARG`s, like `BASE_IMAGE` and `SYTEST_IMAGE_TAG`.

...but with that said, CI always sets `BASE_IMAGE` and `SYTEST_IMAGE_TAG` even when their defaults would suffice.  So for consistency's sake, I'll make CI always set the `*_VERSION` args too.

AndrewFerr: 16383797

.github/workflows/docker.yml
AndrewFerr: It passed :partying_face:

MadLittleMods: CI is all green here as well :white_check_mark: :tada: 

@AndrewFerr Should I merge now?

MadLittleMods (approved):

.github/workflows/docker.yml
AndrewFerr: Yes, please!

PR #1415, opened 2025-10-21
Fix DBD::Pg install for debian:testing image build

AndrewFerr:
Installing DBD::Pg via [install-deps.pl](https://github.com/matrix-org/sytest/blob/develop/install-deps.pl) currently fails under Debian Testing, apparently [due to a change in GCC 15](https://github.com/bucardo/dbdpg/issues/135).

An alternative is installing it via Debian repositories, which works even for non-Testing Debian.

### DIFF
--- a/docker/base.Dockerfile
+++ b/docker/base.Dockerfile
@@ -25,7 +25,8 @@ RUN apt-get -qq update && apt-get -qq install -y \
     sqlite3 \
     wget \
     libicu-dev \
-    pkg-config
+    pkg-config \
+    && rm -rf /var/lib/apt/lists/*
 
 # Set up the locales, as the default Debian image only has C, and PostgreSQL needs the correct locales to make a UTF-8 database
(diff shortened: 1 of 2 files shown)


docker/base.Dockerfile
```diff
@@ -62,3 +64,5 @@ RUN for ver in `ls /usr/lib/postgresql | head -n 1`; do \
 
 # configure it not to try to listen on IPv6 (it won't work and will cause warnings)
 RUN echo "listen_addresses = '127.0.0.1'" >> "$PGDATA/postgresql.conf"
+
+RUN rm -rf /var/lib/apt/lists/*
```
anoadragon453: I'd recommend appending `rm -rf /var/lib/apt/lists/*` to both `RUN apt-get` commands instead, otherwise we'll have a bloated docker layer after each command.

anoadragon453 (changes requested):
LGTM otherwise. Thanks for solving this one!

docker/base.Dockerfile
AndrewFerr: Sure. f2815639

anoadragon453 (approved):
Thanks!

PR #1419, opened 2026-03-13
Stop testing unspecced POST /send

tulir:
For https://github.com/element-hq/synapse/pull/19557

### DIFF
--- a/tests/10apidoc/34room-messages.pl
+++ b/tests/10apidoc/34room-messages.pl
@@ -1,25 +1,3 @@
-test "POST /rooms/:room_id/send/:event_type sends a message",
-   deprecated_endpoints => 1,
-   requires => [ local_user_and_room_fixtures() ],
-
-   do => sub {
-      my ( $user, $room_id ) = @_;
-
-      do_request_json_for( $user,
-         method => "POST",
-         uri    => "/v3/rooms/$room_id/send/m.room.message",
-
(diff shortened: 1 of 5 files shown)


sandhose (approved):

tests/10apidoc/34room-messages.pl
```diff
@@ -1,25 +1,3 @@
-test "POST /rooms/:room_id/send/:event_type sends a message",
```
MadLittleMods, line 1: I assume we're removing because an [equivalent test in Complement](https://github.com/matrix-org/complement/blob/12210e3e418b12c62bac5868d06aa72a9cfc5119/tests/csapi/room_messages_test.go#L25-L67) already exists

tulir: This is being removed because it's only testing an endpoint that was removed from the spec in 2015. The test immediately below this one is testing the correct /send endpoint

MadLittleMods: > The test immediately below this one is testing the correct /send endpoint

Perfect :+1:

clokep:
Any reason not to merge this? Seems like a nice little clean-up tying the tests closer to the spec.

MadLittleMods:
Looks good to me. I assume it was just forgotten after approval :fast_forward: 

Thanks for the clean-up on this side @tulir

PR #1423, opened 2026-05-15
Fix tests to work with Rust event

erikjohnston:
The PR https://github.com/element-hq/synapse/pull/19701 ports the event class to Rust, which as a side-effect adds some extra validation when we parse the event. As such, when we create "fake" events in tests we need to make sure they are sufficiently valid to parse, otherwise we don't test what we wanted to test.

### DIFF
--- a/lib/SyTest/Federation/Datastore.pm
+++ b/lib/SyTest/Federation/Datastore.pm
@@ -379,6 +379,11 @@ sub create_room
    my $creator = $args{creator};
    my $room_version = $args{room_version} // 1;
 
+   # Allow override the room version we add to the create event, to test the
+   # case where the create event's room version is different from the room's
+   # actual version.
+   my $room_version_create_event = $args{room_version_create_event} // $room_version;
+
    my $room = SyTest::Federation::Room->new(
       datastore => $self,
       room_version => $room_version,
@@ -386,6 +391,7 @@ sub create_room
 
    $room->create_initial_events(
       creator => $creator,
+      room_version => $room_version_create_event,
    );
 
    $self->{rooms_by_id}{ $room->room_id } = $room;
--- a/lib/SyTest/Federation/Datastore.pm
+++ b/lib/SyTest/Federation/Datastore.pm
@@ -379,9 +379,10 @@ sub create_room
    my $creator = $args{creator};
    my $room_version = $args{room_version} // 1;
 
-   # Allow override the room version we add to the create event, to test the
-   # case where the create event's room version is different from the room's
-   # actual version.
+   # Allow override of the room version declared in the create event.
+   #
+   # Events will be formatted according to the `$room_version` but the
+   # room will be declared as this room version instead.
    my $room_version_used_for_create_event = $args{room_version_used_for_create_event} // $room_version;
 
    my $room = SyTest::Federation::Room->new(
--- a/lib/SyTest/Federation/Datastore.pm
+++ b/lib/SyTest/Federation/Datastore.pm
@@ -382,7 +382,7 @@ sub create_room
    # Allow override the room version we add to the create event, to test the
    # case where the create event's room version is different from the room's
    # actual version.
-   my $room_version_create_event = $args{room_version_create_event} // $room_version;
+   my $room_version_used_for_create_event = $args{room_version_used_for_create_event} // $room_version;
 
    my $room = SyTest::Federation::Room->new(
       datastore => $self,
@@ -391,7 +391,7 @@ sub create_room
 
    $room->create_initial_events(
       creator => $creator,
-      room_version => $room_version_create_event,
+      room_version => $room_version_used_for_create_event,
    );
 
    $self->{rooms_by_id}{ $room->room_id } = $room;
--- a/lib/SyTest/Federation/Room.pm
+++ b/lib/SyTest/Federation/Room.pm
@@ -176,9 +176,9 @@ sub create_initial_events
    my $creator = $args{creator} or
       croak "Require a 'creator'";
 
-   my $room_version = $args{room_version} // (
-      $self->room_version eq "1" ? undef : $self->room_version
-   );
+   # Only declare a room version if its not 1.
+   my $room_version = $args{room_version} // $self->room_version;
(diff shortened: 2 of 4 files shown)


tests/50federation/30room-join.pl
```diff
@@ -941,7 +945,7 @@ sub assert_is_valid_pdu {
                user_id => $user_id,
             );
 
-            $proto->{origin_server_ts} = $inbound_server->time_ms;
+            $proto->{origin_server_ts} = JSON::number($inbound_server->time_ms);
```
MadLittleMods, line 951: What happened before? What type was this?

lib/SyTest/Federation/Room.pm
```diff
@@ -176,9 +176,9 @@ sub create_initial_events
    my $creator = $args{creator} or
       croak "Require a 'creator'";
 
-   my $room_version = $args{room_version} // (
-      $self->room_version eq "1" ? undef : $self->room_version
-   );
+   # Only declare a room version if its not 1.
+   my $room_version = $args{room_version} // $self->room_version;
+   $room_version = $room_version eq "1" ? undef : $room_version;
```
MadLittleMods, line 181: Why change this logic?

I suppose `$args{room_version}` was defined as `1` and we don't want that for some reason?

Perhaps this is just a preference thing as the spec defaults to `1`:

> `room_version`	string	
The version of the room. Defaults to "1" if the key does not exist.
>
> *-- https://spec.matrix.org/v1.18/client-server-api/#mroomcreate*

tests/50federation/30room-join.pl
```diff
@@ -928,7 +929,10 @@ sub assert_is_valid_pdu {
          creator => $creator_id,
          alias   => $room_alias,
 
-         room_version => 'sytest-room-ver',
+         # We want this room to act like a v1 room, but declare itself an
+         # unknown version.
+         room_version              => '1',
+         room_version_create_event => 'sytest-room-ver',
```
MadLittleMods: As far as I understand, we're declaring the `room_version` as `1` but creating all of the events with some unknown room version so they are rejected.

I also assume this means the events are created with room version 1 format but declared as some unknown format in the `m.room.create` event.

Comment could use updating

lib/SyTest/Federation/Datastore.pm
```diff
@@ -379,13 +379,19 @@ sub create_room
    my $creator = $args{creator};
    my $room_version = $args{room_version} // 1;
 
+   # Allow override the room version we add to the create event, to test the
+   # case where the create event's room version is different from the room's
+   # actual version.
+   my $room_version_create_event = $args{room_version_create_event} // $room_version;
```
MadLittleMods: Perhaps `room_version_used_for_create_event`

tests/50federation/30room-join.pl
erikjohnston: Perl doesn't differentiate between the different types of scalar values, so you can't tell if something is a string or a number. The JSON lib therefore serialises all scalars as strings, and requires the above annotation for numbers.

lib/SyTest/Federation/Room.pm
erikjohnston: Yeah, so: a) by default the room version sytest uses is `1`, and b) `$args{room_version}` is `undef` if using the default room version. Previously, this meant that if you explicitly set the room version to `1` you would get different behaviour than if you didn't, i.e. we would add the `room_version` field to the create event if explicitly set. 

This change just makes us always treat v1 rooms the same, regardless of whether we explicitly set the version or not. I don't think this is actually required for this PR in the end, but I think it makes sense.

tests/50federation/30room-join.pl
erikjohnston: I've tried updating the comments.

In other words: we want to create a room that declares itself as an unknown version. However, to actually create the events we need to base it off an actual room version (otherwise we don't know the format to use for the events). We choose v1 here as the base, since we later rely on the fact that the events are formatted as v1 (in the /make_join step below).

An alternative change would have been to have `room_version` and `base_room_version` variables, and update all the locations in the perl room code to use `base_room_version`. I think that would be a larger change than this, and we don't use the functionality anywhere else so.

lib/SyTest/Federation/Datastore.pm
erikjohnston: 7a24e79

lib/SyTest/Federation/Room.pm
MadLittleMods: ehhh, perhaps you want to test the default when no room_version set, etc
The new behavior seems like a personal preference and could be surprising but is fine ⏩

tests/50federation/30room-join.pl
MadLittleMods: Perhaps deserves a comment given it was wrong in the first place

lib/SyTest/Federation/Datastore.pm
```diff
@@ -379,13 +379,19 @@ sub create_room
    my $creator = $args{creator};
    my $room_version = $args{room_version} // 1;
 
+   # Allow override the room version we add to the create event, to test the
+   # case where the create event's room version is different from the room's
+   # actual version.
+   my $room_version_used_for_create_event = $args{room_version_used_for_create_event} // $room_version;
```
MadLittleMods: Previous comment made more sense to me:

```
# We want this room to act like a v1 room, but declare itself an
# unknown version.
```

Perhaps:

```
# Allow override of the room version declared in the create event.
#
# Events will be formatted according to the `$room_version` but the
# room will be declared as this room version instead.
```

Related discussion, https://github.com/matrix-org/sytest/pull/1423#discussion_r3249487229

MadLittleMods (approved):

lib/SyTest/Federation/Datastore.pm
erikjohnston: 2ae2eed

tests/50federation/30room-join.pl
erikjohnston: 6ccaa14

